PR #13, opened 2017-10-03
fixed sparse efficiency warning

babbush:
Changed code in _jw_hf_state.py to use dok_matrix instead of csr_matrix in order to avoid sparse efficiency warning.


babbush:
Good advice. This code was originally contributed by @idk3. I was just addressing the issue that arises due to using the wrong type of sparse matrix.

PR #24, opened 2017-10-10
Updated Dockerfile and README

yudongcao:
Dockerfile builds openfermion-projectq, openfermion-psi4, openfermion-pyscf

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,28 @@
+# Dockerfile for OpenFermion+ProjectQ (including plugin)
+
+FROM ubuntu
+
+USER root
+
+RUN apt-get update
+
+# Install python, pip
+RUN yes | apt-get install wget
+RUN cd /root; wget https://bootstrap.pypa.io/get-pip.py
+## Change the following lines for different Python versions/distributions
+RUN yes | apt-get install python
+RUN python /root/get-pip.py
+
+# Install projectq (login to Github may be requested)
+RUN yes | apt-get install git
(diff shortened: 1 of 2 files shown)


idk3:
@yudongcao are there ways to test that this builds correctly? 

Otherwise, my understanding is that one of the main ideas for Docker is to get things working cross-platform more easily - how much have you tried with this? (E.g. getting the C++ simulator working in ProjectQ can be a little tricky, Psi4 can be annoying on Windows, etc)

yudongcao:
@idk3 Yes there are ways to test this (see below and also Hannah gave some test file to run - everything seems fine). I got the C++ simulator to work in the Docker container (when a MainEngine instance is initiated there is no warning message saying "this is a (slow) Python simulator"). There is no testing of the Psi4 and PySCF yet - the compilation and installation both seem successful though.

---
Testing the Dockerfile
---
Go to the directory "docker" and run

docker built -t image_openfermion .

where "image_openfermion" is just the name of the image to be built (can be anything). After the image is built, which should take maximum a few minutes, run the image

docker run -it image_openfermion

and the docker session should emulate a Ubuntu environment with ProjectQ, OpenFermion and its plugins installed. The C++ simulator on ProjectQ should be also working. Running some preliminary testfile provided Hannah seems to suggest that everything is OK.

babbush:
Thanks for this contribution @yudongcao! The one thing I would request is that you add the Apache 2 header to the top of the docker file. This should technically be in all code files. Just copy it from one of the files in src. I will then merge the pull request.

PR #29, opened 2017-10-12
Meanfield dwave

kevinsung:
It's useful to me to have modules for constructing Hamiltonian models other than the Hubbard. This one contains one for the BCS meanfield dwave model, and my intention is for it to contain other meanfield models. I'm new to the terminology so maybe some names of things need to be changed. I can also add more tests if necessary. A lot of this code is adapted from _hubbard.py and _hubbard_test.py.


babbush:
This module looks great! I am in Korea right now (where it is midnight) and so I'll need to take a closer look tomorrow. Things look good to go for the most part but it seems that test coverage is decreasing slightly. Our standards for test coverage are admittedly obnoxiously high but it looks like you'll need to test about one additional line of code to pass the threshold we set in coveralls (I believe right now the max you can decrease coverage by is 0.01% and this one decreases by 0.02%). Using the pytest-cov module (which needs be pip installed separately from pytest) you can figure out which lines of code are being missed by your tests. Note that if you wanted to (for some reason) you could also bring up test coverage on some other part of the library to correct the discrepancy.

babbush:
Errr. So the CLA bot has gone wild. Kevin's CLAs were approved before I commented on this pull request but now that I have, CLA bot has decided to block the PR. @dabacon can you help us with this? There is definitely something wrong. The CLA bot has also made a similarly illogical decision on Kevin's other pull request.

src/openfermion/utils/_meanfield.py
```diff
@@ -0,0 +1,127 @@
...
+x_dimension by y_dimension. They can have periodic boundary conditions or not.
+"""
+from __future__ import absolute_import
+
+from openfermion.ops import (FermionOperator,
+                             hermitian_conjugated)
+
+
+# Function to return up-orbital index given orbital index.
+def up(index):
```
babbush: Can you perhaps import these functions from the Hubbard model module instead of rewriting them? Definitely better to avoid code duplication. Since up and down are not available for import from openfermion.utils you should import them from ._hubbard.

src/openfermion/utils/_meanfield.py
```diff
@@ -0,0 +1,127 @@
...
+    return 2 * index
+
+
+# Function to return down-orbital index given orbital index.
+def down(index):
```
babbush: Same as above comment.

src/openfermion/utils/_meanfield.py
```diff
@@ -0,0 +1,127 @@
...
+    return 2 * index + 1
+
+
+def meanfield_dwave(x_dimension, y_dimension, tunneling, sc_gap,
+                  periodic=True, verbose=False):
```
babbush: Super nit-picky: there should be a couple spaces added so that "periodic" lines up right under "x_dimension". If you run the PEP 8 linter, it will make the same comment.

src/openfermion/utils/_meanfield.py
```diff
@@ -0,0 +1,127 @@
...
+        verbose: An optional Boolean. If True, print all second quantized
+            terms.
+
+    Returns:
+        meanfield_dwave_model: An instance of the FermionOperator class.
+    """
+    # Initialize fermion operator class.
+    n_sites = x_dimension * y_dimension
+    n_spin_orbitals = 2 * n_sites
+    meanfield_dwave_model = FermionOperator((), 0.0)
```
babbush: This is fine to write (okay to be explicit) but you can just write FermionOperator() to initialize the zero operator.

src/openfermion/utils/_meanfield.py
```diff
@@ -0,0 +1,127 @@
...
+            meanfield_dwave_model += hermitian_conjugated(hopping_term)
+            # Add spin-down hopping term
+            operators = ((down(site), 1), (down(right_neighbor), 0))
+            hopping_term = FermionOperator(operators, -tunneling)
+            meanfield_dwave_model += hopping_term
+            meanfield_dwave_model += hermitian_conjugated(hopping_term)
+
+            # Add pairing term
+            operators = ((up(site), 1), (down(right_neighbor), 1))
+            pairing_term = FermionOperator(operators, sc_gap / 2)
```
babbush: What you write here is totally functional but a nit picky comment is that in python it is best to always write, e.g. "2." instead of "2" when you intend a float. Since sc_gap is a float this is okay. But if by some chance somebody provided an int like "7" instead of "7." for the sc_gap parameter, then 7/2 = 3 instead of 3.5 in some versions of python. This comment applies to a few things in this PR.

kevinsung:
Thank you for the comments. This is my first time contributing to an open source project and I appreciate the nitpicking!

src/openfermion/utils/__init__.py
```diff
@@ -18,6 +18,8 @@
 
 from ._hubbard import fermi_hubbard
 
+from ._meanfield import meanfield_dwave
```
babbush: I missed this last time: imports should be alphabetical (i.e. jellium imports should come before meanfield). But I won't let this hold things up.

PR #34, opened 2017-10-14
Docker update

yudongcao:
- Updated dockerfile to include installation for jupyter-notebook
- Updated the README file for including how to run Docker containers, how to set up OpenFermion in Docker and how to set up Jupyter Notebook in Docker.


docker/README.rst
```diff
@@ -0,0 +1,173 @@
+Docker Setup for FermiLib + ProjectQ
```
babbush: FermiLib!? Be careful about that please. You've written FermiLib instead of OpenFermion in several places.

docker/README.md
```diff
@@ -1,47 +0,0 @@
-# Docker setup for OpenFermion and select plugins
```
babbush: I'm not sure why you deleted these instructions. Why don't you just add the part that is new and keep these old instructions otherwise. I spent some time polishing these instructions last time. Not to mention the issues I mention below, e.g. calling the package "FermiLib".

docker/README.rst
```diff
@@ -0,0 +1,173 @@
...
+====================================
+
+This Docker image will help users to easily install `FermiLib <https://github.com/ProjectQ-Framework/FermiLib.git>`__ and `ProjectQ <https://github.com/ProjectQ-Framework/ProjectQ>`__. Check out Docker's `website <https://www.docker.com/what-container>`__ that describes what a container image is and why it can be so useful.
+
+What is included?
+-----------------
+
+- Python 2.7 (you can also use Python 3 with one minor change in the Dockerfile. See the Dockerfile for instructions.)
```
babbush: Dockerfile -> dockerfile. Also there should be a link to it. Again, I think its best to add to our old instructions instead of replacing them entirely.

docker/README.rst
```diff
@@ -0,0 +1,173 @@
...
+
+
+Running OpenFermion with Docker
+-------------------------------
+
+5. Now that Docker is set up, one could navigate to the folder containing the
+Dockerfile for building the OpenFermion image (docker/dockerfile) and run
+
+.. code-block:: bash
+	docker build -t kickass_openfermion .
```
babbush: While I appreciate the humor here, perhaps do not call it this. OpenFermion is a family friendly package!

docker/dockerfile
```diff
@@ -1,15 +1,3 @@
-#   Licensed under the Apache License, Version 2.0 (the "License");
```
babbush: Why are you removing the Apache 2 license notice?

docker/README.md
yudongcao: I will add those back - sorry about the typos and hasty deletes

babbush: No problem. Did you see other comments? I don't mean to rush you but since you addressed these but not the others I wanted to check.

yudongcao: Yes I saw all the comments. I will also get Psi4 and PySCF to work in the new dockerfile. The thing with capital D in Dockerfile is just that Docker recognizes "Dockerfile" as a default file name. Otherwise the build command will need to specify the name of the file for building the Docker image. It's not a big deal though.

babbush: ok yeah sorry. The "Dockerfile" -> "dockerfile" comment was the one I cared about least ;-)

docker/dockerfile
yudongcao: I must have removed it accidentally... sorry

docker/README.rst
yudongcao: I have replaced FermiLib with OpenFermion everywhere

docker/README.rst
yudongcao: I have added back the old instructions, with my stuffs starting from "Setting up Docker for the first time". From the latest README.md, one issue I see is that the section "Running OpenFermion in Docker" I wrote has a large overlap with the old "Usage" section. My section is a bit more hand-holding and covers slightly more common issues that I came across when I was using Docker (and I imagine other people will run into these issues too e.g. how to open more than one terminals running the same container). But feel free to edit the README however you see fit. I don't have particular insistence on any of the content being there.

docker/README.rst
yudongcao: fixed

PR #44, opened 2017-10-15
Preparation of fermionic Gaussian states

kevinsung:
This lets you compile circuits for preparing fermionic Gaussian states (ground states of general quadratic Hamiltonians). I combined this and my other code for preparing Slater determinants into one file called `_slater_determinants.py`, although maybe it should be called `_fermionic_gaussian_states.py` instead.

The input is an n x (2n) matrix that represents new annihilation operators as sums of old creation and annihilation operators, where n is the number of sites. This matrix determines a unitary operator, which, when applied to the vacuum state, prepares the ground state of a quadratic Hamiltonian. The code decomposes the input matrix, and hence the preparation unitary, into a sequence of elementary steps. The details of this procedure will appear in an upcoming paper of Zhang and myself.

I plan to shortly write code that can get the input matrix from the description of a quadratic Hamiltonian.


babbush:
I will override CLA bot when testing finishes.

PR #54, opened 2017-10-17
Move UCC independent functions back in OpenFermion

jarrodmcc:
Moves the routines related to UCC that were independent of ProjectQ out of the OpenFermion-ProjectQ repo back into the main branch under utilities.  Merge this before merging branch.

### DIFF
--- a/src/openfermion/utils/_unitary_cc.py
+++ b/src/openfermion/utils/_unitary_cc.py
@@ -1,0 +1,198 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Module to create and manipulate unitary coupled cluster operators."""
+
+import itertools
+import numpy
+
+from openfermion.ops import FermionOperator, QubitOperator
+
+def uccsd_operator(single_amplitudes, double_amplitudes, anti_hermitian=True):
+    """Create a fermionic operator that is the generator of uccsd.
+
+    This a the most straight-forward method to generate UCCSD operators,
+    however it is slightly inefficient. In particular, it parameterizes
+    all possible excitations, so it represents a generalized unitary coupled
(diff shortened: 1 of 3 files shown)


src/openfermion/utils/_unitary_cc.py
```diff
@@ -0,0 +1,199 @@
...
+                (2 * j + s, 0),
+            ),
+            coefficient=t1[t1_ind(i, j)])
+
+        uccsd_generator += FermionOperator(
+            (
+                (2 * j + s, 1),
+                (2 * (i + n_occupied) + s, 0),
+            ),
+            coefficient=-t1[t1_ind(i, j)])
```
babbush: I'm pretty sure that this right here is a pep8 violation. Run the linter?

src/openfermion/utils/_unitary_cc.py
```diff
@@ -0,0 +1,199 @@
...
+
+    for i, j, s, i2, j2, s2 in itertools.product(*spaces, repeat=2):
```
babbush: Can you add a couple single-line comments to the function to explain what you are doing in blocks of code such as this one?

src/openfermion/utils/_unitary_cc_test.py
```diff
@@ -0,0 +1,204 @@
...
+"""Module to test unitary coupled cluster operators."""
+
+from __future__ import absolute_import
+
+import numpy
+import os
+import scipy
+import unittest
+
+from numpy.random import randn
```
babbush: one space before the openfermion imports.

src/openfermion/utils/_unitary_cc.py
jarrodmcc: Passes just fine on mine with pep8, I'll add some spaces if that's what you mean?

babbush: Oops, yeah you're good. Its because that equal sign is setting the value of an optional parameter. That is the correct way to do it. I read quick and thought you were assigning an variable without spaces, which would have been a violation.

src/openfermion/utils/_unitary_cc.py
jarrodmcc: Added some description to the iterator space

src/openfermion/utils/_unitary_cc_test.py
jarrodmcc: k

src/openfermion/utils/_unitary_cc.py
jarrodmcc: I removed the coefficient label since it wasn't needed.  The reason it passed was that this is the PEP8 standard for keyword arguments (no space), even though it looks like an independent assignment line.

babbush:
You're good to merge when tests pass

PR #56, opened 2017-10-17
Molecular Data Update

jarrodmcc:
Makes a few changes to the MolecularData class, in particular

1. Adds overlap integrals back in, with corresponding changes to Psi4 and PySCF plugins
2. Removes restrictions for storing data related to abstract molecules (e.g. jellium, hubbard) without a geometry, but with all the other features such as integrals, basis set, RDMs etc
3. Adds a field for collecting arbitrary uncharacterized data called general_calculations.  This lets one store the results from many calculations related to this particular molecule within the file.


src/openfermion/hamiltonians/_molecular_data.py
```diff
@@ -36,7 +35,7 @@
 # Define a compatible basestring for checking between Python 2 and 3
 try:
```
babbush: What is this doing? I'm not criticizing it, just asking a question.

src/openfermion/hamiltonians/_molecular_data.py
```diff
@@ -229,6 +232,8 @@ class MolecularData(object):
         ccsd_energy: Energy from coupled cluster singles + doubles.
         ccsd_single_amps: Numpy array holding single amplitudes
         ccsd_double_amps: Numpy array holding double amplitudes
+        general_calculations: A dictionary storing general calculation results
```
babbush: Good idea.

src/openfermion/hamiltonians/_molecular_data.py
```diff
@@ -359,6 +376,18 @@ def canonical_orbitals(self, value):
         self._canonical_orbitals = value
 
     @property
+    def overlap_integrals(self):
```
babbush: I know what you are doing here but because decorators are an "advanced" aspect of the language it is a good practice to add a one-liner saying what you are doing since this will confuse more novice developers.

src/openfermion/transforms/_conversion.py
```diff
@@ -171,3 +172,59 @@ def get_fermion_operator(interaction_operator):
...
+                       geometry=None, basis=None, multiplicity=None,
+                       n_electrons=None, reduce_spin=True,
+                       data_directory=None):
+    """Output a MolecularData object generated from an InteractionOperator
+
+    Args:
+        interaction_operator(InteractionOperator): two-body interaction
+            operator defining the "molecular interaction" to be simulated.
+        geometry(string or list of atoms):
+        basis(string):
```
babbush: Do write something, even if it is obvious. Don't just leave empty.

babbush:
@jarrodmcc note that CLA bot is acting up again. If Travis and coveralls pass, you may need to use admin override to force merge. Once you've addressed my comments, you have my permission to do this (no need to seek another review).

src/openfermion/hamiltonians/_molecular_data.py
jarrodmcc: It's a compatibility fix for data types in hdf5.  Python2 and 3 have different default string classes and it causes problems between the versions. The no cover statement is because I don't see how to get both clauses to execute in both Python2 and 3, which means in both it lowers the test coverage by 1 line, however both lines are run in the tests separately, so it should be okay.

src/openfermion/hamiltonians/_molecular_data.py
jarrodmcc: I'm not sure we want to add one for all 10 getters and setters, but I'll put a block comment at the top of the section.

src/openfermion/transforms/_conversion.py
jarrodmcc: Oops didn't see that

jarrodmcc:
Did you commit from a different computer than normal? I think all CLA bot checks is the email address signed with the commits, which GitHub tolerates a few of for logins, but only puts the one you used (or in your current config file) on the commits

PR #68, opened 2017-10-19
Added hubbard model with particle-hole symmetry

jromerofontalvo:
I added an alternative version of the routine for creating the Fermi Hubbard hamiltonian. In this version the repulsion term U n_{i,\uparrow} n_{i,\downarrow} is replaced by  U (n_{i,\uparrow}-1/2) (n_{i,\downarrow}-1/2), which is unchanged by a particle-hole transformation. This version has the advantage that at chemical potential and magnetic field equal zero, the ground state is always the one at half-filling and the energy is symmetric with respect to chemical potential = 0. This is more convenient for benchmarks than the previous version. Look at the energies for a 2x2 lattice:
 
image

For now it is a separate routine.

### DIFF
--- a/src/openfermion/hamiltonians/__init__.py
+++ b/src/openfermion/hamiltonians/__init__.py
@@ -14,7 +14,7 @@ from ._chemical_series import (make_atomic_ring,
                                make_atomic_lattice,
                                make_atom)
 
-from ._hubbard import fermi_hubbard
+from ._hubbard import fermi_hubbard, fermi_hubbard_phs
 
 from ._jellium import (dual_basis_kinetic,
                        dual_basis_potential,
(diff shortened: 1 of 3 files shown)


babbush:
This looks like a valuable contribution. However, I think we should aim to add this functionality to the existing Hubbard model code. For instance, have an optional input parameter (False by default) which if True, creates the Hubbard model you are interested in here. The goal would be to duplicate as little code as possible and only add the minimum number of extra lines required to create the functionality you want when this flag is true.

Otherwise, the contribution looks pretty solid. You might need a few more tests and you do need to sign the CLA if you haven't done so already.

jromerofontalvo:
I updated it to have a single function with a particle_hole_symmetry Boolean option set by default as False.

babbush:
Thanks for a nice contribution Jhonathan! And welcome to OpenFermion!

PR #85, opened 2017-10-27
Rename references to InteractionTensor to PolynomialTensor

kevinsung:
Related to #61, which renamed InteractionTensor to PolynomialTensor. I also fixed this in the Overleaf draft which was sent out before the Oct. 23 release; I hope that was the right place.


babbush:
Yep, that was the right draft. There is only the one!

PR #88, opened 2017-10-27
Adds basic single qubit error channels for simulation convenience

jarrodmcc:
Adds the ability to apply simple 1-qubit error channels numerically to given density matrices with tests.  Useful for testing error resilience in VQE and QSE methods.


src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+def _verify_channel_inputs(density_matrix, probability, target_qubit):
+    """Verifies input parameters for channels
+
+    Args:
+        density_matrix(numpy.ndarray): Density matrix of the system
+        probability(float): Probability error is applied p \in [0, 1]
+        target_qubit(int): target for the channel error.
+
+    Returns:
+        new_density_matrix(numpu.ndarray): Density matrix with the channel
```
idk3: numpu

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+            applied.
+    """
+    n_qubits = int(log2(density_matrix.shape[0]))
+
+    if (density_matrix.shape[0] != density_matrix.shape[1] or
+            len(density_matrix.shape) != 2):
```
idk3: These should be checked in the opposite order so `shape[0] != shape[1]` can't cause an error

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        raise ValueError("Error in input of density matrix to channel.")
+    if (probability < 0) or (probability > 1):
```
idk3: What if `probability` is complex? (OK to disregard this)

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        raise ValueError("Channel probability must be between 0 and 1.")
+    if (target_qubit < 0) or (target_qubit >= n_qubits):
+        raise ValueError("Target qubits must be within number of qubits.")
+
+
+def _lift_operator(operator, n_qubits, target_qubit):
+    """Lift a single qubit operator into the n_qubit space by kron product
+
+    Args:
+        operator(ndarray): Single qubit operator to lift into full space
```
idk3: spaces after names

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        n_qubits(int): Number of total qubits in the space
+        target_qubit(int): Qubit to act on
+
+    Return:
+        new_operator(Sparse Operator): Operator representing the embedding in
```
idk3: space after name

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+            the full space.
+    """
+    new_operator = reduce(kron,
```
idk3: Could you maybe respace this? If you put kron on a new line you might be able to fit the `eye(2) for i in range(...)` on a single line - I think that would make this easier to read

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+
+def amplitude_damping_channel(density_matrix, probability, target_qubit,
+                              transpose=False):
+    """Apply an amplitude damping channel
+
+    Applied an amplitude damping channel with a given probability to the target
+    qubit in the density_matrix.
+
+    Args:
+        density_matrix(numpy.ndarray): Density matrix of the system
```
idk3: Spaces needed, but nice!!

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        probability(float): Probability error is applied p \in [0, 1]
+        target_qubit(int): target for the channel error.
+        transpose(bool): Transpose channel operators, useful for acting on
+            Hamiltonians in variational channel state models
+
+    Returns:
+        new_density_matrix(numpy.ndarray): Density matrix with the channel
```
idk3: space

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+                          dot(E1, dot(density_matrix, E1.T)))
+
+    return new_density_matrix
+
+
+def dephasing_channel(density_matrix, probability, target_qubit,
+                      transpose=False):
+    """Apply a dephasing channel
+
+    Applied an amplitude damping channel with a given probability to the target
```
idk3: Applied -> Apply

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+    qubit in the density_matrix.
+
+    Args:
+        density_matrix(numpy.ndarray): Density matrix of the system
```
idk3: Spaces

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        probability(float): Probability error is applied p \in [0, 1]
+        target_qubit(int): target for the channel error.
+        transpose(bool): Transpose channel operators, useful for acting on
+            Hamiltonians in variational channel state models
+
+    Returns:
+        new_density_matrix(numpy.ndarray): Density matrix with the channel
```
idk3: Space

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+                          dot(E1, dot(density_matrix, E1.T)))
+
+    return new_density_matrix
+
+
+def depolarizing_channel(density_matrix, probability, target_qubit,
+                         transpose=False):
+    """Apply a depolarizing channel
+
+    Applied an amplitude damping channel with a given probability to the target
```
idk3: Apply

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+    qubit in the density_matrix.
+
+    Args:
+        density_matrix(numpy.ndarray): Density matrix of the system
```
idk3: Spaces

src/openfermion/utils/_channel_state.py
```diff
@@ -0,0 +1,184 @@
...
+        probability(float): Probability error is applied p \in [0, 1]
+        target_qubit(int): target for the channel error.
+        transpose(bool): Transpose channel operators, useful for acting on
+            Hamiltonians in variational channel state models
+
+    Returns:
+        new_density_matrix(numpy.ndarray): Density matrix with the channel
```
idk3: space

src/openfermion/utils/_channel_state_test.py
```diff
@@ -0,0 +1,148 @@
...
+            depolarizing_channel(test_density_matrix, 1, 1,
+                                 transpose=True))
+
+        self.assertAlmostEquals(norm(correct_density_matrix -
+                                     test_density_matrix), 0.0, places=6)
+
+    def test_verification(self):
+        """Verify basic sanity checking on inputs"""
+        with self.assertRaises(ValueError):
+            _ = amplitude_damping_channel(self.density_matrix, 2, 1)
```
idk3: You likely don't need to do assignment for this, `amplitude_damping_channel(...)` would be fine - but it doesn't really matter

src/openfermion/utils/_channel_state.py
jarrodmcc: When did we start doing this? Most of the code was created without spaces before the type from what I recall.

src/openfermion/utils/_channel_state.py
babbush: I don't care about this

jarrodmcc:
Comments should be taken care of

src/openfermion/utils/_channel_state.py
idk3: It's a tiny bit cleaner and all code has it, we might as well maintain it

src/openfermion/utils/_channel_state.py
idk3: It's a tiny bit cleaner and all code has it, we might as well maintain it

idk3:
@babbush @jarrodmcc do we want to maintain the spacing point in general? My thought is that it's nice to have, and cheap to maintain given that we already have it - but if neither of you care at all we could just not keep doing it

babbush:
In general we do care about spacing. But Jarrod is right that this is one area of the style guide where there are enough inconsistencies that I wouldn't hang up a pull request over it.

babbush:
Put another way, right now I'm happy when I see people putting argument types in parenthesis. Because that is not practiced throughout OpenFermion. I try to mandate it in all PRs now though. However, whether there is a space after the arg name or not doesn't matter to me because it is picked up by sphinx just the same as far as I can tell.

PR #89, opened 2017-10-28
Data structure for quadratic Hamiltonians

kevinsung:
Subclass of PolynomialTensor specifically for quadratic Hamiltonians (e.g. mean field models), both particle-number-conserving and not. I considered the use of the Majorana basis but decided that I really only need to use it internally. So this class uses the usual fermionic ladder operators for representation.

This is pretty basic so far, but I plan to add routines for conversion to and from FermionOperator, as well as methods to obtain circuits for ground state preparation. Just thought I should pause and ask for feedback before continuing.


src/openfermion/ops/_quadratic_hamiltonian.py
```diff
@@ -0,0 +1,104 @@
...
+        constant: A constant term in the operator given as a float.
+    """
+
+    def __init__(self, constant, hermitian_part,
+                 antisymmetric_part=None, chemical_potential=None):
+        """
+        Initialize the QuadraticHamiltonian class.
+
+        Args:
+            constant(float): A constant term in the operator.
```
babbush: So, I am wondering if the constant should be stored (in PolynomialTensor) as the entry in the dictionary that is represented by (). This makes sense to me since that is how the constant is stored for FermionOperator. What do you think?

src/openfermion/ops/_quadratic_hamiltonian.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def hermitian_part(self):
+        """Return the Hermitian part not including the chemical potential."""
+        hermitian_part = self.combined_hermitian_part()
+        if self.chemical_potential:
+            hermitian_part += (self.chemical_potential *
+                               numpy.eye(self.n_qubits))
+        return hermitian_part
+
+    def antisymmetric_part(self):
```
babbush: I feel like since you call the other part the "Hermitian part" this should be called the "anti-Hermitian part". What do you think?

src/openfermion/ops/_quadratic_hamiltonian.py
kevinsung: I thought about that too, and the only reason I didn't implement that is that then the line
`self.n_qubits = list(n_body_tensors.values())[0].shape[0]`
wouldn't work, since `list(n_body_tensors.values())[0]` might return the constant, which is a float. However, we could also just remove the `()` entry of `n_body_tensors` when __init__ is called, and replace it with the constant passed in. That just seemed kind of ugly to me.

src/openfermion/ops/_quadratic_hamiltonian.py
kevinsung: I'm pretty sure the matrix really is antisymmetric (A = -A^T), rather than anti-Hermitian (A = -A^\dagger).

src/openfermion/ops/_quadratic_hamiltonian.py
kevinsung: Actually, I think I do have a better solution; I can do it later or submit in another PR.

babbush:
Okay, I'll merge this one then.

PR #98, opened 2017-11-01
Adds totally depolarizing channel to existing depolarizing channel

jarrodmcc:
Adds option to specify "all" for target_qubit such that depolarizing channel will act as the totally depolarizing channel, mapping to the identity.


jarrodmcc:
Those spaces fixed as well as some others

PR #99, opened 2017-11-02
Rewrote jw_hartree_fock_state

kevinsung:
I noticed that `jw_hartree_fock_state` could be made more efficient. I actually wrote a more general function `jw_slater_determinant` which lets you instantiate any computational basis vector. I think it's an improvement.

The original function returns a CSR matrix but I noticed that we mostly use CSC so I returned a CSC. Not sure what the policy on that is (if any). Along those lines, I fixed what seems to be a typo in the initialization of `identity_csc`.


idk3:
These are good changes! Good catch about CSC vs CSR - pre-release the code was mostly CSR but we changed to improve sparse efficiency. Unfortunately CLA bot has gone rogue again - @babbush could you check it out / override?

idk3:
Actually, CLA only got annoyed when I merged master in - can you check if my license is registered properly?

PR #102, opened 2017-11-05
Handle KeyError when using 'old' hdf5 files.

hsim13372:
I ran into a KeyError when I tried to load some 'old' hdf5 files (when attribute 'general_calculations' wasn't yet implemented in MolecularData).

### DIFF
--- a/src/openfermion/hamiltonians/_molecular_data.py
+++ b/src/openfermion/hamiltonians/_molecular_data.py
@@ -699,16 +699,16 @@ class MolecularData(object):
             data = f["ccsd_energy"][...]
             self.ccsd_energy = data if data.dtype.num != 0 else None
             # Load general calculations
-            try:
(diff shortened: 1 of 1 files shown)


hsim13372:
I signed it!

hsim13372:
Hi Ryan, 

Thanks a lot for your comment! I just changed the try statement to an if instead. Regarding the testing, I purposefully made this PR without adding to the test because I wanted to ask you and the other developers about what to do when there's an edit similar to this in MolecularData. Would it be okay to add an old or "faulty" data file to the data subdirectory or would this be setting a bad precedent?

hsim13372:
but for now, I'll try to look for a different line in the code that still needs a test and test that.

babbush:
I think you may have fixed the coverage problem just by switching to the "if" statement. The coverage check passed so I went ahead with it. Thanks!

PR #103, opened 2017-11-06
Bug in Jordan Wigner ordering

kevinsung:
One can convert a FermionOperator to a Jordan-Wigner encoded sparse operator by either directly calling `jordan_wigner_sparse`, or by first converting it to a QubitOperator using `jordan_wigner` and then calling `qubit_operator_sparse`. I think these should give the same result, but they don't:

```
In [4]: ferm_op = FermionOperator('1')

In [5]: qubit_op = jordan_wigner(ferm_op)

In [6]: ferm_op_sparse = jordan_wigner_sparse(ferm_op)

In [7]: qubit_op_sparse = qubit_operator_sparse(qubit_op)

In [8]: print(ferm_op_sparse)
  (0, 1)	(1+0j)
  (2, 3)	(1+0j)

In [9]: print(qubit_op_sparse)
  (0, 1)	(1+0j)
  (2, 3)	(-1+0j)
```

This bug arises from different conventions being used in `jordan_wigner` and `jordan_wigner_ladder_sparse`. In the first case, the string of Pauli Z's are placed at lower indices (i.e., with six modes, a_3 gets mapped to `Z Z Z Q_lower I I`), and in the second case, the string is placed at higher indices (i.e., a_3 gets mapped to `I I I Q_lower Z Z`). I fixed this bug by changing `jordan_wigner_ladder_sparse` to use the first convention.

### DIFF
--- a/src/openfermion/transforms/_jordan_wigner.py
+++ b/src/openfermion/transforms/_jordan_wigner.py
@@ -23,6 +23,10 @@ def jordan_wigner(operator):
     """ Apply the Jordan-Wigner transform to a FermionOperator or
     InteractionOperator to convert to a QubitOperator.
 
+    Operators are mapped as follows:
+    a_j^\dagger -> Z_0 .. Z_{j-1} (X_j - iY_j) / 2
+    a_j -> Z_0 .. Z_{j-1} (X_j + iY_j) / 2
+
     Returns:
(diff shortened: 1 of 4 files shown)


kevinsung:
Okay I made the changes. I added a more explicit test for the reverse JW.

babbush:
Great catch, thanks Kevin!

PR #104, opened 2017-11-07
Procedure for ground state of quadratic Hamiltonians

kevinsung:
- Added `jw_get_quadratic_hamiltonian_ground state`, which computes the ground energy and state of a quadratic Hamiltonian. It can be used instead of `get_ground_state` when possible; it should be faster and more memory efficient. It also serves as a test for the Slater determinant preparation circuits.
- Some minor optimizations and rewriting of Givens rotations code and tests
- Fixed a bug in `get_ground_state` which caused it to return the complex conjugate of the ground state instead of the true ground state.


kevinsung:
I think `jw_get_quadratic_hamiltonian_ground_state` still fits better in `_slater_determinants.py` for now. However, I took your second suggestion.

PR #105, opened 2017-11-08
PolynomialTensor dictionary should be accessed through methods

kevinsung:
When I was rewriting PolynomialTensor, I forgot to make it so that the attributes `one_body_tensor` and `two_body_tensor` of InteractionOperator are updated after a call to `rotate_basis`. After thinking a little about the best way to fix this, I concluded that `one_body_tensor` should actually be a method, rather than an attribute. More generally, for any subclass of PolynomialTensor, information tied to the dictionary `n_body_tensors` should be  retrieved using methods. That way any functions that manipulate PolynomialTensor objects don't need to worry about updating related attributes in subclasses.

What do people think of this? I probably should have made an issue first, but it didn't take long to just implement the change. The one thing I wasn't sure about was whether the method should return a copy or not; for now I made it return a copy because it seems less likely to cause bugs.

### DIFF
--- a/src/openfermion/ops/_interaction_operator.py
+++ b/src/openfermion/ops/_interaction_operator.py
@@ -64,9 +64,14 @@ class InteractionOperator(PolynomialTensor):
                 {(): constant,
                  (1, 0): one_body_tensor,
                  (1, 1, 0, 0): two_body_tensor})
-        self.constant = self.n_body_tensors[()]
-        self.one_body_tensor = self.n_body_tensors[1, 0]
-        self.two_body_tensor = self.n_body_tensors[1, 1, 0, 0]
+
+    def one_body_tensor(self):
+        """Get the one-body tensor."""
+        return self.n_body_tensors[1, 0]
+
+    def two_body_tensor(self):
+        """Get the two-body tensor."""
+        return self.n_body_tensors[1, 1, 0, 0]
 
     def unique_iter(self, complex_valued=False):
         """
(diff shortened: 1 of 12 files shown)


kevinsung:
On second thought, perhaps this change is not necessary, because the `one_body_tensor` attribute is actually a view into the numpy array stored in the `n_body_tensors` dictionary? It might still be a good change though.

jarrodmcc:
I'm not sure I fully understand the motivation behind the change.  We should avoid redundancy in data storage where possible, so if it's being updated somewhere and we are passing references to it that should be fine?  If we want to update it only when the element is accessed, then we should use getters and setters as in molecular_data, so that it acts like an attribute but has the flexibility of a function.

kevinsung:
My main motivation was that I thought the `one_body_tensor` attribute needed to be updated, but I was mistaken because I hadn't realized it was really a view (reference). So actually, I'm thinking most of this PR might not be needed after all. However, the `constant` attribute is not a view, so do you think that should be a method of PolynomialTensor? I'm imagining a scenario where a function changes the value of `n_body_tensors[()]` without updating the corresponding `constant` attribute. But maybe I'm overthinking things.

jarrodmcc:
I'm still not sure why it being a method makes a difference.  You are always free to have the object modify it's own internal components, including the constant, and components of the dictionary.  Not having a view doesn't really matter, as it just rebinds the variable and will still return the correct value when checked?  Loading the attribute from the class that has been updated should still work.

Maybe I'm missing a use case you have personally, where you load the value of the tensor somewhere else, and want changing the object to change a pointer to it elsewhere?

jarrodmcc:
Oh you are talking about wanting both a "constant" attribute and an n_body_tensors[()] value that are consistent? I think I understand now.  I would use the getter and setters via decorators we use in molecular data for this, that will let it behave like an attribute but access the dictionary to get the updated value every time it is accessed.

kevinsung:
Ah, that is a neat solution I wasn't aware of. I think this PR can be dismissed; I might submit another related one in the future using the decorators. Thanks!

PR #112, opened 2017-11-22
Minor restructuring

kevinsung:
- Wrote a wrapper for Slater determinant preparation. Now there are only two functions from `utils/_slater_determinants.py` which are exposed in `utils/__init__.py`. I think this gives a cleaner interface and will make it easier to write about in the paper.
- Minor changes to documentation and a conversion function.


babbush:
Or, more importantly, tests seem to fail in python2.7 and 3.6!

kevinsung:
I got the checks to pass by removing the Givens rotations example. I plan to eventually add a better example, perhaps directly to the demo notebook, which demonstrates Slater determinant preparation at a higher level of abstraction.

babbush:
That seemed like a useful example. What exactly was failing?

kevinsung:
The Givens rotation example notebook used the function `givens_decomposition` from `utils/_slater_determinants.py`, which I chose to remove from `utils/__init__.py` in this PR. My reasoning was that in practice the higher-level function (wrapper) that I introduced here will probably always be called instead. Anyway, I'm going to take a break now until next week but if you think we should keep the lower level function in the namespace as well as the example it should be a quick fix.

PR #113, opened 2017-11-24
Energy cutoff for plan wave hamiltonian.

Spaceenter:


Spaceenter:
@babbush @idk3 Do you have suggestions for how to do unit tests for the energy cutoff? 
My impression is that, after energy cutoff, dual basis and plane wave hamiltonians are not necessarily the same - am I right? If that's the case, how to do unit tests?

babbush:
Hi Wei, I will need to look into it but actually I am under the impression that there is a dual basis Hamiltonian which is isospectral to any plane wave Hamiltonian. They are not related exactly by the Fourier transform, but by a more complicated transformation. Have you checked to see if the dual basis Hamiltonian (when basis functions are centered on points in a sphere instead of in points in a cube) is equivalent to the energy cutoff plane wave Hamiltonian?

babbush:
A somewhat lame test would be to make sure that the Hamiltonian with a sphere of plane waves had greater than or equal to the Hamiltonian with a cube of frequencies where the cube side length is twice the radius of the sphere. Does that make sense?

Spaceenter:
@babbush 
Are you referring to, for example, plane_wave_external_potential()? I think the alias (shift) is already done when computing the indices, right? 
My understanding (from Garnet's and your email) is that, we compute the shifted/aliased indices by assuming a cubic cutoff, then we apply energy cutoff on the momentum which is computed from the shifted/aliased indices - is that right? 
Or, should I firstly apply energy cutoff, then compute the shifted/aliased indices?

babbush:
Hi Wei,

Can you forward the email you mention here to me? I recall that email in the past but I can't find it in my inbox right now. I'll respond after I've considered this a bit more.

Spaceenter:
forwarded to your @google email
@babbush

babbush:
Okay, I just looked over them. So perhaps we do not need to alias in the way I just suggested (perhaps the current implementation is fine). It is hard for me to know for sure though. I think it would be nice to try the test where we compare the Hamiltonian in the basis of the plane wave sphere to the Hamiltonian in the basis of the dual function sphere. If those two things are equivalent, I am happy.

idk3:
The spherical equivalence test is definitely a good one - can one of you ( @babbush @Spaceenter ) forward the email from that to me? The sphere/cube ground state comparison also sounds good, but it would be nice if there's something we can say about the higher energy states..

babbush:
So I am sorry to have confused you @Spaceenter but apparently there is no equivalence between the plane wave and dual basis models with the cutoff. I don't think it makes sense at all to work with the dual basis in a spherical cutoff (only the plane wave Hamiltonian). So I would ask that you remove the changes you made to the dual basis parts of the code and only keep this cutoff business for the plane wave parts. Sorry!

Spaceenter:
I see, probably that explains why I cannot get those two equal to each other with cutoff, I was trying to see if I made any mistake - thanks, I'll remove soon and update here.

Spaceenter:
Done - PTAL. @babbush

babbush:
Thanks @Spaceenter. I guess there needs to be one more test though. Can you think of anything? Like I say, the energy of the ground state can only go down (or stay the same) as one adds plane waves. So one possible test is just to make sure that imposing the cutoff makes energy go down or stay the same.

Spaceenter:
Added test - but I don't quite understand why coverage drops - The following seems not right to me:
https://coveralls.io/builds/14452849/source?filename=src%2Fopenfermion%2Fhamiltonians%2F_plane_wave_hamiltonian_test.py#L71

-- I don't understand why the whole test is not run, given that part of code is not changed. Do you have insights? @babbush

babbush:
Quite strange. Could it have something to do with the word "integration"?
Maybe it is not running that because it thinks it is an integration test
and it only wants to run unit tests? That is just a theory. I really have
no idea what is happening.

Spaceenter:
In general, I think *_test.py shouldn't be included in coverage test - they should be covered 100% by default.
Do you know who set up the Travis/Coverage test? Could we ask him/her to remove all coverage checks for unit test files?
@babbush

fangzh-umich:
@Spaceenter I think of course they need to be included in coverage test. If some of the tests are not run for some reason, it's a problem that needs to be solved, and without the coverage test no one will ever notice that.

Spaceenter:
@fangzh-umich Do you have insights why the following code is not run?
https://stackoverflow.com/questions/20420429/what-the-difference-between-window-settimeout-and-settimeout

fangzh-umich:
Also, I've found the problem: You defined a test with the same name at the end of the file, so it overwrites the first test.

Spaceenter:
Oh, sorry - yea, that's a mistake from my side. Thanks!

Spaceenter:
@babbush PTAL

Issue #37, opened 2017-10-14
Code for BCH expansion


babbush:
Several times now I have wanted to implement code which performs a BCH expansion on operator data structures in OpenFermion such as QubitOperator, FermionOperator, and sparse matrices. I started to write some code for this:

```python

def bch_expand(x, y, order):
    """Compute log[e^x e^y] using the Baker-Campbell-Hausdorff formula
    Args:
        x: An operator for which multiplication and addition are supported.
            For instance, a QubitOperator, FermionOperator or numpy array.
        y: The same type as x.
        order(int): The order to truncate the BCH expansions. Currently
            function goes up to only third order.

    Returns:
        z: The truncated BCH operator.

    Raises:
        ValueError: operator x is not same type as operator y.
        ValueError: invalid order parameter.
        ValueError: order exceeds maximum order supported.
    """
    # Initialize.
    max_order = 4
    if type(x) != type(y):
        raise ValueError('Operator x is not same type as operator y.')
    elif (not isinstance(order, int)) or order < 0:
        raise ValueError('Invalid order parameter.')

    # Zeroth order.
    z = x + y

    # First order.
    if order > 0:
        z += commutator(x, y) / 2.

    # Second order.
    if order > 1:
        z += commutator(x, commutator(x, y)) / 12.
        z += commutator(y, commutator(y, x)) / 12.

    # Third order.
    if order > 2:
        z -= commutator(y, commutator(x, commutator(x, y))) / 24.

    # Fourth order.
    if order > 3:
        z -= commutator(
            y, commutator(y, commutator(y, commutator(y, x)))) / 720.
        z -= commutator(
            x, commutator(x, commutator(x, commutator(x, y)))) / 720.
        z += commutator(
            x, commutator(y, commutator(y, commutator(y, x)))) / 360.
        z += commutator(
            y, commutator(x, commutator(x, commutator(x, y)))) / 360.
        z += commutator(
            y, commutator(x, commutator(y, commutator(x, y)))) / 120.
        z += commutator(
            x, commutator(y, commutator(x, commutator(y, x)))) / 120.

    # Return.
    if order > max_order:
        raise ValueError('Order exceeds maximum order supported.')
    else:
        return z
```

[...]

PR #114, opened 2017-11-28
BCH Efficient expansion

jdaaph:
Resolve issue #37 
I tested the efficient expansion over hard coded coefficient in the first four orders and will post a detailed latex explanation for the algorithm shortly. Only the bch_expand will be exposed to user. 

Algorithm - Cupjin Huang (@pertoX4726), Implementation - Chengyu Dai


jdaaph:
to Cupjin Huang (@pertoX4726), 
Please sign the google's cla (https://cla.developers.google.com/)

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+    if (not isinstance(order, int)) or order < 0:
+        raise ValueError('Invalid order parameter.')
+    if order > max_order:
+        raise ValueError('Order exceeds maximum order supported.')
+    if type(x) != type(y):
+        raise ValueError('Operator x is not same type as operator y.')
+
+    z = None
+    terms, coeff = generate_nested_commutator(order)
+    for bin_str, c in zip(terms, coeff):
```
babbush: better to call "c" something like "coefficient" so the code is more readable.

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+        t = bin_str_to_commutator(bin_str, x, y)
+        if z is None:
+            z = t * c
+        else:
+            z += t * c
+
+    # Return.
+    return z
+
+def bin_str_to_commutator(bin_str, x, y):
```
babbush: All functions (even helper functions) must have at least a short one line description. Ideally, also provide args and returns.

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+        return [bin_str[i:j] for i, j in zip(split_idx, split_idx[1:]+[None])]
+
+def compute_coeff(split_bin_str):
+    N = len(''.join(split_bin_str))
+    l = len(split_bin_str) - 1
+    cn = lambda n: dfs_root(split_bin_str, n, len(split_bin_str))
+    c = sum([(-1)**(n+1) / n * cn(n) for n in range(l+1, N+1)])
+    return c/N
+
+def dfs_root(split_bin_str, n, l):
```
babbush: A better name for this function would be good. Not sure what "dfs" is. Depth first search?

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+    assert len(split_bin_str) == len(eta_lst)
+    ret = 1
+    for block, eta in zip(split_bin_str, eta_lst):
+        cnt_x = block.count('0')
+        cnt_y = block.count('1')
+        ret *= g(cnt_x, cnt_y, eta)
+    return ret
+
+
+def g(cnt_x, cnt_y, eta):
```
babbush: definitely need a better name here.

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+import numpy as np
+from scipy.misc import comb, factorial
+
+from openfermion.utils import commutator
+
+def bch_expand(x, y, order):
+    """Compute log[e^x e^y] using the Baker-Campbell-Hausdorff formula
+    Args:
+        x: An operator for which multiplication and addition are supported.
+            For instance, a QubitOperator, FermionOperator or numpy array.
```
kevinsung: Really minor: I think you should replace the phrase "numpy array" with "Scipy sparse matrix", because multiplication for Numpy arrays is actually element-wise. Your code might still work for Numpy arrays though because the `commutator` function will support them.

jdaaph:
Thanks for the help! We'd love to be added as the v2 paper's coauthors. I'll update it as soon as Cupjin finishes his latex document and change function / variable name accordingly.

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+    Returns:
+        z: The truncated BCH operator.
+
+    Raises:
+        ValueError: operator x is not same type as operator y.
+        ValueError: invalid order parameter.
+        ValueError: order exceeds maximum order supported.
+    """
+
+    max_order = 11
```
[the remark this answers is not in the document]
jdaaph: I think because the implementation is still an exponential complexity (though much better than naive) it's still necessary to set an upper bound, since we don't want to blow up the user's RAM, we use mostly on the fly computation and this can lead to super long wait time. I think it's reasonable that we simply raise an exception but perhaps with better explanation.

src/openfermion/utils/_bch_expansion_test.py
```diff
@@ -0,0 +1,103 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+from __future__ import absolute_import
+import unittest
+from numpy.random import rand, seed
```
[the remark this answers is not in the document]
bbbbbbbbba: You are not OK with abbreviations but you are OK with "from package import name"?

src/openfermion/utils/_bch_expansion.py
bbbbbbbbba: Correct me if I'm wrong, but I think this is a fairly time-consuming operation (need to compute one commutator per order) that are called many times (one time for each term in the expansion, and then in recursion) with same parameters (many intermediate results are previously computed terms). Optimization potential here?

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,157 @@
...
+
+def generate_nested_commutator(order):
+    """
+    using bin strings to encode nested commutators like [X,[Y,[X, ...]]] as '010...'
+    """
+    terms = []
+    coeff = []
+
+    for i in range(1, order + 1):
+        t = [list(x) for x in itertools.product(['0', '1'], repeat=i)]
```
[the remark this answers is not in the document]
bbbbbbbbba: Just asking: are you OK with "terms" and "term" in the same namespace?

src/openfermion/utils/_bch_expansion_test.py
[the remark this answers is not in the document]
babbush: yes

src/openfermion/utils/_bch_expansion.py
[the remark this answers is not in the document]
babbush: doesn't seem ideal but that should be okay. I am just asking for more descriptive names.

babbush:
Sounds good. For the paper, what are the full names for the two of you and if you have an affiliation that you would like listed, what is the entire affiliation as you'd like it show up on the paper?

src/openfermion/utils/_bch_expansion.py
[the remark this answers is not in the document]
fangzh-umich: I agree that setting an maximum order is not ideal. I will argue that a hard cutoff doesn't achieve anything: either computing the 11th order is already too slow (and the cutoff doesn't achieve its intended purpose), or computing the 12th order still only takes a reasonable time (and you are blocking the user from doing reasonable work).

jdaaph:
Chengyu Dai, Department of Physics, University of Michigan, Ann Arbor, MI 48109
Cupjin Huang, Department of Electrical Engineering and Computer Science, University of Michigan, Ann Arbor, MI 48109

Thanks!

babbush:
By the way, we are hoping to release OpenFermion v0.2 on Monday. So if there is any possibility of wrapping up this pull request by then it would be nice so it could make it in.

src/openfermion/utils/_bch_expansion.py
jdaaph: Yes, I'll optimize it after v0.2. It's fast enough for 10by10 matrices up to order 7-8 so I think I'll leave like this in the current version, because it makes the code way easier to read.

babbush: That's fine with me.

README.rst
```diff
@@ -131,8 +131,10 @@ Authors
 `Daniel Sank <https://github.com/DanielSank>`__ (Google),
 `Nicolas Sawaya <https://github.com/nicolassawaya>`__ (Harvard),
 `Kanav Setia <https://github.com/kanavsetia>`__ (Dartmouth),
-`Hannah Sim <https://github.com/hsim13372>`__ (Harvard) and
-`Wei Sun <https://github.com/Spaceenter>`__ (Google).
+`Hannah Sim <https://github.com/hsim13372>`__ (Harvard),
+`Wei Sun <https://github.com/Spaceenter>`__ (Google),
+`Chengyu Dai <https://github.com/jdaaph>`__ (University of Michigan) and
```
babbush: This part of the author list is alphabetical by last name. So please put your names in the correct order.

src/openfermion/utils/_bch_expansion.py
```diff
@@ -0,0 +1,188 @@
...
+
+    Returns:
+        z: The truncated BCH operator.
+
+    Raises:
+        ValueError: operator x is not same type as operator y.
+        ValueError: invalid order parameter.
+        ValueError: order exceeds maximum order supported.
+    """
+    MAX_ORDER = 15
```
babbush: Sorry but I really don't think this MAX_ORDER should be there at all. If you really want to have it you can have the code raise a warning but do not have it raise an exception, as it is doing below.

README.rst
babbush: You should also update the NOTICE file. The NOTICE file is required by law to stay with any forks of the repo so it will make sure you get credit forever!

PR #115, opened 2017-11-28
Commutator function

kevinsung:
For Numpy arrays, `*` actually means element-wise multiplication, so I think the description is incorrect. I simply changed the description, but an alternative way to fix this issue is to add support for Numpy arrays, which I would be glad to do.


jarrodmcc:
This is a correct change, well spotted.  If you would be willing to add a line to support 2D numpy arrays I think that would be a good update, as well.  Will merge either way once you let me know.

kevinsung:
I added support for Numpy arrays.

jarrodmcc:
Looks great, thanks for the addition!

Issue #86, opened 2017-10-27
General n-body rotations for PolynomialTensor


kevinsung:
Currently, the `rotate_basis` method of the PolynomialTensor class only rotates the (1, 0) and (1, 1, 0, 0) tensors, i.e., the tensors corresponding to terms of the form a^\dagger_p a_q and a^\dagger_p a^\dagger_q a_r a_s. Ideally, it would correctly rotate all tensors that are stored in the PolynomialTensor instance.

The rotation of the (1, 0) tensor is straightforward (just a basis change for a matrix), and the implementation of the rotation for the (1, 1, 0, 0) tensor is described in equations (22) and (23) of the release paper. The task for this issue is to work out how to generalize these expressions to arbitrary tensors and implement it in the `rotate_basis` method.

fangzh-umich:
Before this is resolved, maybe make the code throw an exception if there are other tensors stored? Failing silently feels dangerous.

fangzh-umich:
From what I have seen, the rotation is just applied to all the indices of the tensor, with no regards to the "key", i.e. the (1, 1, 0, 0) and the (0, 0, 1, 1) tensors should be rotated the same way. Can anyone confirm that this is indeed the correct behavior?

fangzh-umich:
It seems that the release paper "specialized" the rotation _U_ to be real and orthogonal. Shouldn't we support the general case anyway?

kevinsung:
Yes, we should support the general case. The rotations of the (1, 1, 0, 0) tensors and (0, 0, 1, 1) tensors should differ by some complex conjugate signs (as we discussed in person).

kevinsung:
I'll note here that for the complex one-body case, the transformation M' = R M R^T generalizes to R* M R^T.

babbush:
Closed by #117

Issue #116, opened 2017-11-30
Inconsistency in basis transformations between code and paper


kevinsung:
I believe there is an inconsistency in how basis transformations of PolynomialTensor are treated between the code and the paper. I think it applies to the implementation of both one- and two-body basis changes, but I'll explain it for one-body. In Eq. (21) of the paper, the molecular orbital functions are transformed by the matrix U. This corresponds to a transformation of the creation operators by the same matrix U. In other words, the column vector of creation operators (a^\dagger_1, ..., a^\dagger_N)^T is multiplied on the left by U. Let's assume that U is real and call it R instead. Then equivalently, the row vector of creation operators (a^\dagger_1, ..., a^\dagger_N) is multiplied on the right by R^T. This implies that for a one-body tensor of type (1, 0), i.e., with terms of the form a^\dagger_p a_q, the one-body tensor matrix M should be transformed as M' = R M R^T. We put R on the left side which will cancel out the R^T which transforms the creation operators; similarly, the R^T on the right side cancels out the R which transforms the annihilation operators. However, in the current implementation of `one_body_basis_change`, the transformation is M' = R^T M R. Does this make sense to anyone? I'm happy to give a more detailed (prettier) explanation with LaTeX. If I'm correct, then I propose that we change the code to be consistent with the paper. Perhaps this can be resolved at the same time as #86 .

babbush:
So you are suggesting that we are implementing R^T M R instead of R M R^T? @jarrodmcc do you have any concerns about this? If not, I am fine signing off on the change.

kevinsung:
Yeah, the code for `one_body_basis_change` actually implements R^T M R, and I'm suggesting that it should be R M R^T instead. I think the code for `two_body_basis_change` would need to be changed too.

babbush:
The thing is that rotation matrices have the property that R^T = R^{-1}, which is just a rotation in the other direction. So you could view what we're actually doing as (R^{-1}) M (R^{-1})^T. So its just rotating in the opposite direction. This doesn't seem like a significant problem to me. Can we not just change what is in the paper? I don't really have strong feelings though.

babbush:
I should also mention that when changing the basis of an operator you should rotate in the opposite direction as when you change the basis of an RDM. I don't think we're being careful about that right now though.

kevinsung:
[...]

PR #117, opened 2017-12-01
N-body rotations

fangzh-umich:
Resolves issues #86 and #116.

I implemented general n-body rotations of PolynomialTensor using np.einsum, with the option optimize = True to ensure that we trade space for time (e.g. in the 2-body term, use the O(N^5) algorithm instead of the O(N^8) one).

Also, as per issue #116, @kevinsung and I changed the convention for the rotation matrix so that it's consistent with the release paper.


fangzh-umich:
OK, I signed it.

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -25,10 +25,57 @@ class PolynomialTensorError(Exception):
     pass
 
 
+def general_basis_change(general_tensor, rotation_matrix, key):
+    """Change the basis of an general interaction tensor.
+
+    M' = R^T.M.R where R is the rotation matrix, M is the general tensor
+    and M' is the transformed general tensor.
```
kevinsung: Should it be R.M.R^T? Also, perhaps there is a better way to express what this does, as transpose seems to me to be a 2-d concept; I might be wrong.

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -25,10 +25,57 @@ class PolynomialTensorError(Exception):
...
+
+    subscripts = subscripts_first + ',' + subscripts_rest
+
+    rotation_matrices = [rotation_matrix.conj() if x == 1 else
+                         rotation_matrix for x in key]
+
+    # "optimize = True" does greedy optimization, which will be enough here
+    transformed_general_tensor = numpy.einsum(subscripts,
+                                              general_tensor,
+                                              *rotation_matrices,
```
kevinsung: What does the `*` in `*rotation_matrices` do?

kevinsung:
For testing, perhaps we can compute the rotated matrix in two different ways, one using your new function, and another using a naive sum, like Eq. (22), and make sure they match up?

Perhaps the functions `one_body_basis_change` and `two_body_basis_change` should be removed completely?

Yes, of course I'm okay with my commits being used in this PR.

fangzh-umich:
I'm not sure how to increase the coverage of my test.

First, there is a maximum order restriction because of a limitation of numpy.einsum: it uses letters to indicate subscripts, and there are only 52 (upper case and lower case) in the English language. I don't expect anyone to actually hit it, and I hope a future version fixes this (probably when numpy itself overcomes this limitation).

Second, my code has no use for `one_body_basis_change` and `two_body_basis_change`, but I left them in because `__init__.py` explicitly imports those functions. I don't know what those imports are for, so I didn't just remove them.

src/openfermion/ops/_polynomial_tensor.py
fangzh-umich: It's a rotation "matrix", of course it's 2-d.

src/openfermion/ops/_polynomial_tensor.py
fangzh-umich: It's a Python syntax to pass a variable number of parameters to a function. For example, if `key` is `(0,0,0)` then this is equivalent to `numpy.einsum(subscripts,  general_tensor, rotation_matrix, rotation_matrix, rotation_matrix)`.

src/openfermion/ops/_polynomial_tensor.py
kevinsung: Haha ok.

src/openfermion/ops/_polynomial_tensor.py
kevinsung: Cool!

kevinsung:
Ok then, this looks good to me after you address the "R.M.R^T" thing.

fangzh-umich:
Actually, my test just showed that numpy complains about "too many subscripts in einsum" before it gets to 26 indices.

fangzh-umich:
Also, I realized that you defined PolynomialTensorError, but you never raised it, and _you never exported it_ (meaning that it's not visible to the user). Since my code raised that exception, I exported it.

But am I supposed to raise some other exception like ValueError? What is the intended use of PolynomialTensorError?

fangzh-umich:
I'm still not sure what I should do with `one_body_basis_change` and `two_body_basis_change`. Should I unexport them or should I test them? Should I export my `general_basis_change` too?

kevinsung:
Don't export PolynomialTensorError, and raise a ValueError instead.

I think we should remove `one_body_basis_change` and `two_body_basis_change`, and also not export `general_basis_change`. I don't see any reason `general_basis_change` needs to be exposed to users, but somebody else (maybe @babbush ) should have the final say on that.

I don't know what to say about `einsum` complaining before there are 26 indices.

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -25,64 +25,58 @@ class PolynomialTensorError(Exception):
...
         rotation_matrix = numpy.kron(rotation_matrix, numpy.eye(2))
 
-    # Effect transformation and return.
-    transformed_one_body_tensor = numpy.einsum('qp, qr, rs',
-                                               rotation_matrix,
-                                               one_body_tensor,
-                                               rotation_matrix)
-    return transformed_one_body_tensor
+    order = len(key)
+    if order > 26:
```
babbush: Where is this order 26 limit coming from? I suggest not having a maximum order restriction.

src/openfermion/ops/__init__.py
```diff
@@ -15,9 +15,7 @@
                                 normal_ordered,
                                 number_operator)
 from ._qubit_operator import QubitOperator
-from ._polynomial_tensor import (PolynomialTensor,
-                                 one_body_basis_change,
-                                 two_body_basis_change)
+from ._polynomial_tensor import PolynomialTensor
```
babbush: we like to have alphabetically organized imports when possible.

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -25,64 +25,58 @@ class PolynomialTensorError(Exception):
...
+        raise ValueError('Order exceeds maximum order upported (26).')
 
+    # 'abcd'
+    subscripts_first = ''.join(chr(ord('a') + i) for i in range(order))
 
-def two_body_basis_change(two_body_tensor, rotation_matrix):
-    """Change the basis of 2-body interaction tensor such as 2-RDM.
+    # 'aA,bB,cC,dD'
```
babbush: Can you add a line or two explaining what is happening here a bit more? I am a little confused.

src/openfermion/ops/_polynomial_tensor.py
fangzh-umich: It comes from using the `numpy.einsum` function, which uses letters for indices, and there are 26 upper case and 26 lower case in total (and I need two letters for each dimension of the tensor). However, actually numpy complains about "too many subscripts in einsum" before the 26 limit is reached.

I don't think anything more than 16 is useful in any case.

src/openfermion/ops/__init__.py
fangzh-umich: Sure, but I didn't introduce this problem. If the person originally writing this code didn't have any particular reason to do it this way, I will change that.

src/openfermion/ops/_polynomial_tensor.py
fangzh-umich: I'm doing the basis change through a single call of `numpy.einsum`, which requires constructing a fitting subscript string. For example, for the (1, 1, 0, 0) tensor, I'm doing `numpy.einsum('abcd,Aa,Bb,Cc,Dd', general_tensor, rotation_matrix.conj(), rotation_matrix.conj(), rotation_matrix, rotation_matrix)`.

Although I just realized my comment is incorrect after the convention change. I will fix that, and at the same time try to make it more explicit.

src/openfermion/ops/__init__.py
fangzh-umich: It turns out that I can't change the order freely, because some files imported later depends on some earlier imports having been processed.

PR #120, opened 2017-12-09
Minor restructuring

kevinsung:
- Made the function for diagonalizing quadratic Hamiltonians a method of QuadraticHamiltonian, renamed it `diagonalizing_bogoliubov_transform`, and updated its description.
- Moved the function `majorana_operator` to `ops/_fermion_operator.py` and exposed it to users.
- Minor changes to QuadraticHamiltonian and Slater determinant preparation code.
- Edited the description of `get_sparse_operator`.


src/openfermion/utils/_slater_determinants_test.py
```diff
@@ -448,26 +441,11 @@ def test_bad_input(self):
...
-
-    Returns:
-        QuadraticHamiltonian
-    """
-    constant = numpy.random.randn()
-    chemical_potential = numpy.random.randn()
-    hermitian_mat = random_hermitian_matrix(n_qubits, real)
-    if conserves_particle_number:
-        antisymmetric_mat = None
+def random_unitary_matrix(n, real=False):
```
babbush: I wonder if this should go in a more general module. Perhaps somewhere else in utils? Because this clearly has nothing to do with Slater determinants, per se, and could be useful in other contexts. Otherwise the PR is good to go.

kevinsung: I had similar thoughts when I wrote the other functions that generate random objects for testing, too. I don't think they fit in any file that already exists. What if we put them in a new module called something like `utils/testing_utils.py`?

babbush: That sounds fine to me. There is also a function in "performance_benchmarks.py" (in examples) called "artificial_molecular_operator" which could probably go in such a module. I think the docs still reference MolecularOperator (so perhaps that should also be updated). Feel free to rename that as you see fit.

babbush: Actually, you might consider putting this in tests/testing_utils.py. I don't feel strongly though.

babbush:
Also, do you have any idea why it now says "docs failing?" I sort of assume this is a mistake as they seem to be fine...

kevinsung:
Weird, it does not say "docs failing" when I view the GitHub page. Anyway, I noticed some Sphinx warnings and fixed them, perhaps that will resolve the issue.

I put the testing functions in `tests/_testing_utils.py`.

babbush:
Try reloading the page with shift, command, R or opening incognito. The reason you likely don't see it is because of caching.

kevinsung:
You're right, it says "docs failing" for me now too. Perhaps it's due to the Sphinx warnings that I fix here? If it's not I'll look into it further.

By the way I moved the test functions to `utils`. To keep it in `tests` I think we would have to add an `__init__.py` there to satisfy Python 2.7 and I'm not sure that's appropriate.

PR #123, opened 2017-12-13
Improved documentation of Hamiltonian models, fix bug in spinless Hubbard model

kevinsung:
- Edited the documentation of the Hubbard and mean-field models and moved them from the module header to the actual function so it shows up in the online documentation.
- Fixed what seems to be a bug in the spinless Hubbard model in which the chemical and magnetic potentials aren't added as described in the model description.


babbush:
@kevinsung Can you push a trivial change to this? I am not sure why it isn't running on Travis but maybe if you make a change it will run.

babbush:
@dabacon do you have insight about what might be going on here? It's been two days and Travis hasn't started on this PR.

Strilanc:
I accidentally fixed the checks not having run when I created a test PR with the same commits and travis did trigger against it. I guess it groups across PR boundaries by commit id.

Anyways, that means the only remaining issue is the CLA check being unable to verify author consent.

Issue #100, opened 2017-11-02
Higher dimensional Jordan-Wigner transform


babbush:
Explained in [this 2005 paper](https://arxiv.org/abs/cond-mat/0508353) by Verstraete and Cirac, it is possible to transform local fermion Hamiltonians on a lattice in D dimensions into local spin Hamiltonians on a lattice in D dimensions. Myself and others I've spoken with recently, including @dabacon, would love to play around with this mapping. So it would be great to have it implemented in OpenFermion. This project would be a great introduction to OpenFermion but requires that one has enough physics background to get through the paper.

kevinsung:
I implemented the transformation on 2-d square grids with an even number of columns. However, I can still think of some things to do:
- Extend to an odd number of columns (this is straightforward and explained in the paper).
- It would be great if we could check that the ground state decomposes as a tensor product, but I don't know how to do this. I don't think it works to check that the qubit state vector is a tensor product, since the tensor product structure of the Hilbert space of qubits is different from that of the Hilbert space of fermionic modes.
- The general idea of the transform is to multiply some operators by stabilizer operators to to cancel out Jordan-Wigner strings. Right now, I only handle quadratic operators corresponding to vertical edges, as described in the paper. I haven't thought about whether it's necessary to handle other operators, and how we would do so.
- My implementation works only for spinless models. I think the way to handle spin varies depending on the particular interactions between orbitals of different spin which are present in the Hamiltonian we want to transform, so this might be a little complicated.
- It would be cool to generalize to different 2-d lattice geometries and also higher dimensions.
- I use one particular graph for the auxiliary Hamiltonian, corresponding to Fig. 2a in the paper, but other auxiliary graphs are possible, and we may want to explore them.

PR #126, opened 2017-12-16
Verstraete-Cirac transform on 2-d grid

kevinsung:

### DIFF
--- a/src/openfermion/transforms/_verstraete_cirac_test.py
+++ b/src/openfermion/transforms/_verstraete_cirac_test.py
@@ -58,10 +58,6 @@ class VerstraeteCirac2dSquareGroundStateTest(unittest.TestCase):
         self.assertAlmostEqual(self.transformed_ground_energy,
                                self.ferm_op_ground_energy)
 
-    def test_ground_state(self):
-        """Test that the ground state is a tensor product state between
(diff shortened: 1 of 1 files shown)


src/openfermion/transforms/_verstraete_cirac.py
```diff
@@ -0,0 +1,278 @@
...
+from openfermion.transforms import jordan_wigner
+
+
+def verstraete_cirac_2d_square(operator, x_dimension, y_dimension,
+                               add_auxiliary_hamiltonian=True,
+                               snake=False):
+    """Apply the Verstraete-Cirac transform on a 2-d square lattice.
+
+    Currently only spinless models are supported. Furthermore, x_dimension
+    should be even.
```
idk3: Consider a NotImplementedError if x_dimension and y_dimension are odd :)

babbush:
It seems that the problem with Travis running was not a fluke. @Strilanc fixed it last time but by some accident essentially. Craig, do you have any ideas what might be going on? Is it possible there is a problem with Kevin's fork and that he copy this PR, delete his fork, refork and resubmit the PR? Or is the problem definitely on our side?

kevinsung:
I deleted my fork and re-forked and made another pull request.

PR #127, opened 2017-12-17
Verstraete-Cirac transform on 2-d grid

kevinsung:


babbush:
Something is broken with Travis. We will get to the bottom of this early next week.
@dabacon @Strilanc there is definitely a problem with Travis. It is no longer running automatically and there is no obvious way to force it to start.

babbush:
Please try one more time. This time, try making a trivial change to the travis.yml file. Like add an extra line or something.

PR #129, opened 2017-12-18
Test to get travis to trigger

dabacon:
Testing travis ci.


babbush:
So why isn't it working from Kevin's fork? He submitted that PR twice, once after deleting his fork and reforking.

PR #130, opened 2017-12-18
Verstraete cirac

kevinsung:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,6 +47,5 @@ install:
 
 script: export OMP_NUM_THREADS=1 && pytest src/openfermion --cov src/openfermion
 
-
 after_success:
   - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,5 +47,6 @@ install:
 
 script: export OMP_NUM_THREADS=1 && pytest src/openfermion --cov src/openfermion
 
+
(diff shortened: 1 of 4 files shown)


babbush:
Nope, it's not that. Kevin, the time you made the duplicate PR, did you completely 100% delete your fork and then refork?

kevinsung:
Yes, I did.

babbush:
So weird. I'm going to try to add to your PR. Let's see if that activates Travis.

kevinsung:
I'll try it one more time.

kevinsung:
OK let's try that first

babbush:
Ok so me also contributing did not work. But now I have another question: when you reforked, how did you save the changes you make in this PR? What I would have suggested is to:
1) Copy all changed files in this PR to your desktop
2) Delete your fork from GitHub
3) Make a clean fork of OpenFermion
4) Checkout a new feature branch
5) Copy changed files in this PR from your desktop to your feature branch
6) Make the PR again.

Did you already try all of this? This is very perplexing to us. Notice also that when Dave made a test PR recently, it incorrectly lists you (not Dave) as the creator of that PR.

kevinsung:
I didn't re-make the local version of my branch. I can try it now.

PR #136, opened 2017-12-19
Added Tarini to author list

babbush:
Please confirm this is okay @TariniHardikar. I will also add you to release paper if you like (but will not push new version to arXiv for a little while).


TariniHardikar:
This is definitely okay, Ryan. And sure, I'd like to be added to the paper.

Thanks!
Tarini

PR #138, opened 2017-12-20
pep8 fixes

idk3:
Main errors involve E126: continuation line over-indented for hanging indent


idk3:
@babbush @kevinsung this PR is ~600 lines of fixes mostly linked to PEP8 E126 (continuation line over-indented for hanging indent) from some of your PRs - please be careful with this one! :P

kevinsung:
My pep8 doesn't catch those, and in the documentation [here](http://pycodestyle.pycqa.org/en/latest/intro.html#error-codes) it says
> In the default configuration, the checks E121, E123, E126, E133, E226, E241, E242, E704 and W503 are ignored because they are not rules unanimously accepted, and PEP 8 does not enforce them.

PR #139, opened 2017-12-21
Improve documentation and hide a helper function

kevinsung:
- Made some additions and changes to documentation.
- Removed the function `get_sparse_polynomial_tensor` from `__init__.py`, since it appears to be only a helper function. I also renamed it to `polynomial_tensor_sparse`. Actually, it seems to fit more in `utils/sparse_tools.py`, but it uses `get_fermion_sparse` which would cause a circular dependency.
- Removed some unnecessary code from `get_sparse_operator`.


babbush:
Looks like Travis is no longer having problems.

src/openfermion/transforms/_verstraete_cirac.py
```diff
@@ -26,8 +26,12 @@ def verstraete_cirac_2d_square(operator, x_dimension, y_dimension,
                                snake=False):
     """Apply the Verstraete-Cirac transform on a 2-d square lattice.
 
-    Currently only spinless models are supported. Furthermore, x_dimension
-    should be even.
+    Note that this transformation adds one auxiliary fermionic mode
+    for each mode already present, and hence it doubles the number of qubits
+    needed to represent the system.
+
+    Currently only supports even values of x_dimension and only works
```
idk3: To be clear, odd y_dimension is also not supported, right?

src/openfermion/utils/_sparse_tools.py
```diff
@@ -69,7 +69,7 @@ def jordan_wigner_ladder_sparse(n_qubits, tensor_factor, ladder_type):
         n_qubits(int): Number qubits in the system Hilbert space.
 
     Returns:
-        The corresponding SparseOperator.
+        The corresponding Scipy sparse matrix.
```
idk3: Very good to catch these, looks like the rest of us have missed it for a long time :P

src/openfermion/transforms/_verstraete_cirac.py
kevinsung: Actually, odd y_dimension is supported! It has to do with the way the auxiliary graph is constructed.

PR #140, opened 2017-12-21
Allow "from openfermion import *"

kevinsung:
I don't know if this would be controversial or if this is the best way to achieve this. But when I'm messing around with Jupyter notebooks or on IPython sometimes I find I want to say `from openfermion import *` or `import openfermion`. Of course I would _never_ do such a thing when actually writing a contribution.


idk3:
+1, this could be very useful. Especially with some of the rearrangements we're still making it could be difficult for a user to keep track of where different things are.

PR #141, opened 2017-12-22
slight refactoring of hubbard model

idk3:


src/openfermion/hamiltonians/_hubbard.py
```diff
@@ -117,36 +118,42 @@ def fermi_hubbard(x_dimension, y_dimension, tunneling, coulomb,
...
         # Add chemical potential and magnetic field terms.
         if chemical_potential and spinless:
-            coefficient = -1. * chemical_potential
+            coefficient = -chemical_potential
             hubbard_model += number_operator(
                 n_spin_orbitals, site, coefficient)
 
         if magnetic_field and spinless:
-            sign = (-1.) ** (site)
+            sign = -1 ** site
```
kevinsung: `**` has higher precedence than `-` so you need parentheses around the `-1`.

src/openfermion/hamiltonians/_hubbard.py
```diff
@@ -218,21 +225,20 @@ def fermi_hubbard(x_dimension, y_dimension, tunneling, coulomb,
...
-                hubbard_model += hopping_term
-                hubbard_model += hermitian_conjugated(hopping_term)
 
-    # Return.
-    return hubbard_model
+            hopping_term = FermionOperator(operators, -tunneling)
+            hubbard_model += hopping_term
+            hubbard_model += hermitian_conjugated(hopping_term)
+
+    return normal_ordered(hubbard_model)
```
kevinsung: Is there a particular reason you want to return a normal-ordered operator? I think normal-ordering should generally be left up to the user to perform. In this case all it does is change the ordering within the coulomb terms, and if someone is examining the returned operator, it may be helpful for it to match the ordering used in the definition of the model.

src/openfermion/hamiltonians/_hubbard.py
```diff
@@ -117,36 +118,42 @@ def fermi_hubbard(x_dimension, y_dimension, tunneling, coulomb,
...
                 n_spin_orbitals, site, coefficient)
 
         if magnetic_field and spinless:
-            sign = (-1.) ** (site)
+            sign = -1 ** site
             coefficient = sign * magnetic_field
             hubbard_model += number_operator(
                 n_spin_orbitals, site, coefficient)
 
         if chemical_potential and not spinless:
```
kevinsung: Since `chemical_potential` and `magnetic_field` now actually have a value (rather than None), you may want to consider replacing these 4 if statements (if chemical_potential and not spinless, etc.) with just 2 (if spinless, if not spinless). It would make it more readable.

kevinsung:
I added some inline comments; overall this is a good change!

src/openfermion/hamiltonians/_hubbard.py
idk3: Good catch! I'll fix this and see if I can add a quick test for this.

src/openfermion/hamiltonians/_hubbard.py
idk3: That could be nice, yeah :) The one thing I worry about is that it might lead to operators with zero coefficients which you'd want to compress - I'll look into it.

src/openfermion/hamiltonians/_hubbard.py
idk3: This is something for discussion - my belief is that it's OK to normal order when there's very little overhead in doing so. The thing I don't like is having to wrap `normal_ordered()` around every call to a Hamiltonian, which feels like a lot of unnecessary duplication.

My feeling is that the tradeoff is between the user's ability to have control and not be forced to do expensive operations unnecessarily, while also keeping the code simple - instead of having to call the same two functions together every time, simplifying it into a single function. Thoughts? @babbush ?

src/openfermion/hamiltonians/_hubbard.py
kevinsung: A possibly good solution to that problem is to modify the init function of FermionOperator to always return a trivial zero operator if the coefficient passed to it is zero.

idk3: Actually, what's the reason for this term more generally? There isn't any splitting with a magnetic field for the spinless case, right? I may just remove the field term there..

kevinsung: I don't know, who defined the spinless Hubbard model in the first place?

idk3: Looks like that code is fairly old, it must've been mistakenly added a while back. I think we can just agree to remove it then!

src/openfermion/hamiltonians/_hubbard.py
babbush: I'd rather you didn't put the normal ordered call here. The user should ask for that explicitly if it's something they want.

src/openfermion/hamiltonians/_hubbard.py
```diff
@@ -90,9 +91,9 @@ def fermi_hubbard(x_dimension, y_dimension, tunneling, coulomb,
         tunneling (float): The tunneling amplitude :math:`t`.
         coulomb (float): The attractive local interaction strength :math:`U`.
         chemical_potential (float, optional): The chemical potential
-            :math:`\mu` at each site. Default value is None.
+            :math:`\mu` at each site. Default value is 0.
         magnetic_field (float, optional): The magnetic field :math:`h`
-            at each site. Default value is None.
+            at each site. Default value is None. Ignored for the spinless case.
```
babbush: Does it still say "Default value is None"?

src/openfermion/hamiltonians/_hubbard.py
```diff
@@ -117,54 +118,48 @@ def fermi_hubbard(x_dimension, y_dimension, tunneling, coulomb,
...
+        n_spin_orbitals *= 2
+
+    hubbard_model = FermionOperator.zero()
+
+    # Select particle-hole symmetry
     if particle_hole_symmetry:
         coulomb_shift = FermionOperator((), 0.5)
     else:
-        coulomb_shift = FermionOperator((), 0.0)
+        coulomb_shift = FermionOperator.zero()
```
babbush: Doesn't FermionOperator() also work? I prefer using that.

idk3:
I've made these final changes, including removing normal ordering.

The only one I disagree with is using `FermionOperator()` instead of `FermionOperator.zero()`. The static method ensures that no matter what `FermionOperator()` does (which can and has changed in the past!) that we get a zero operator. If we're committed to never changing the default behaviour of `FermionOperator()` again, I'll make this change.

PR #143, opened 2017-12-29
Hubbard trotter code

idk3:
Hubbard Trotter error code + minor changes in _hubbard.py and _dual_basis_trotter_error.py


src/openfermion/utils/_hubbard_trotter_error.py
```diff
@@ -0,0 +1,142 @@
...
+
+    Notes: follows Equation 9 of Poulin et al.'s work in "The Trotter Step
+        Size Required for Accurate Quantum Simulation of Quantum Chemistry".
+    """
+    return dual_basis_error_operator(
+        terms, indices=indices, is_hopping_operator=is_hopping_operator,
+        verbose=verbose)
+
+
+def fermi_hubbard_error_bound(terms, indices=None, is_hopping_operator=None,
```
babbush: This function is specific to the second order Trotter formula. Somehow that should be reflected in the name of the function. Also, what is it that makes this code specific to the Fermi-Hubbard model? Is it simply that it anticipates only having a^\dagger_p a_q and n_p n_q terms? Because that form also appears for the dual basis.

src/openfermion/utils/_hubbard_trotter_error.py
```diff
@@ -0,0 +1,142 @@
...
+    Returns:
+        A float upper bound on norm of error in the ground state energy.
+
+    Notes:
+        Follows Equation 9 of Poulin et al.'s work in "The Trotter Step
+        Size Required for Accurate Quantum Simulation of Quantum
+        Chemistry" to calculate the error operator.
+    """
+    # Return the 1-norm of the error operator (upper bound on error).
+    return dual_basis_error_bound(
```
babbush: This whole function seems kinds of pointless. Wouldn't it be better to rename the dual basis function to reflect that it is really specific to diagonal fermionic Hamiltonians as opposed to being specific for either Fermi-Hubbard or dual basis?

src/openfermion/utils/_hubbard_trotter_error.py
```diff
@@ -0,0 +1,142 @@
...
+        terms, indices=indices, is_hopping_operator=is_hopping_operator,
+        verbose=verbose)
+
+
+def simulation_ordered_grouped_hubbard_terms_with_info(
```
babbush: The name needs to reflect the very specific ordering that is being reflected here. Refer to the arXiv paper in the doc strings so it is clear what the order is.

src/openfermion/utils/_hubbard_trotter_error.py
idk3: I don't know the word for what the dual basis function actually allows (linear combinations of number-conserving Hamiltonians acting on only two modes). There is no independent use for this code right now (though there could be in the future?) other than allowing the simulation to be run from a single module.

src/openfermion/utils/_hubbard_trotter_error.py
idk3: This code always uses the same restrictions on the dual_basis_error_operator code (forcing several optional variables to their defaults) so I think it's nice to have for that simplicity.

PR #144, opened 2018-01-04
Fixed orbital rotations

babbush:
Fixes https://github.com/quantumlib/OpenFermion-Psi4/issues/25

As @conta877 mentioned in the above issue, there are some problems with the molecular orbital rotations in the current implementation of the generalized basis rotation. As far as I can tell, the problem basically boils down the change in convention we were discussing in (https://github.com/quantumlib/OpenFermion/issues/116). I think that we do want to stay with the old convention.

However, my changes now break two tests in _polynomial_tensor_test. I need help fixing those tests (either edit the PR directly or give me suggestions) and I would also ask @fangzh-umich and @kevinsung to try to explain to me how the current implementation of the generalized basis rotation is working. It appears that numpy.einsum magically knows to transpose the second half of the rotation matrices. How is it doing that? @jarrodmcc and I are rather confused.

Another issue is that I am not at all convinced that just turning on the "optimize" flag in einsum is indeed allowing us to accomplish the basis rotation at complexity O(N^5) versus O(N^8). @fangzh-umich you claimed this was the case. Can you please elaborate?


fangzh-umich:
About the O(N^5) versus O(N^8) thing:

https://ahaldane.github.io/reference/generated/numpy.einsum_path.html

I was afraid I would need to write some complicated demonstration to convince you, but thankfully the second example on this page is essentially equivalent to our 2-body rotation. The relevant lines from the output:

```
         Naive scaling:  8
     Optimized scaling:  5
```

fangzh-umich:
> It appears that numpy.einsum magically knows to transpose the second half of the rotation matrices.

No, all the rotation matrices are factored into the transformation in the same way.

In `M' = R^T.M.R` one of the two rotation matrices _seems_ to be transposed, but that's an artifact of writing one on the left of the tensor and one on the right (and using the 2-dimensional tensor M as a matrix).

In general, I don't know what formula you could write down that would make _half_ of the transformation matrices look transposed.

babbush:
Thanks @fangzh-umich. Both of your comments are correct and helpful. Indeed, the only problem was moving from the R^T M R to R M R^T convention. The problem, I realize now, is that all modern electronic structure codes use our old convention.

In any event, the last thing the remains is we need to fix the two tests that I've skipped. Any advice?

babbush:
Nevermind, I have fixed the test now. Ready for code review @jarrodmcc

jarrodmcc:
It's worth mentioning for point of record that the R matrices that come out of electronic structure codes are neither unitary nor orthogonal.  Thus there are some non-trivial aspects to when the orders are switched related to contra- vs co-variance in tensor transformations. As such it's probably better to keep the original convection.

fangzh-umich:
@jarrodmcc In that case, are they _real_? Because we believed the original code didn't conjugate the rotation matrix correctly (it assumed that R is real), and we fixed that in our pull request. But if your code was working correctly with the old code then the fix may break it.

babbush:
@fangzh-umich they are real for the case of atomic orbitals to molecular orbitals. They are not necessarily real for other basis sets, including plane waves.

fangzh-umich:
> the R matrices that come out of electronic structure codes are neither unitary nor orthogonal

In that case I really doubt whether you should use our basis rotation code, which is designed with unitary matrices in mind. Are you sure you are not using a piece of code that just happens to do the same thing you want to do mathematically?

babbush:
As I now understand it, for the one and two body case your code implements the exact same operation that our old code implemented. It working for LiH (as it now does) is a strong endorsement.

jarrodmcc:
I believe the current code uses the transformation matrices correctly, despite them not being unitary.

fangzh-umich:
It may use them "correctly" right now, but that can be changed any moment now by the next programmer, based on the (written) assumption that `rotation_matrix` is unitary.

Yes, the chance is small (matrix inverse is more expensive than transpose), but I think discussing about it, and maybe removing the assumption from the in-code comment, is the right thing to do.

babbush:
As you can see, we already merged the PR. But I think it would probably be appropriate to remove the written assumption. Please feel free to do that and I'll approve the PR.

fangzh-umich:
I would like to hear a quick explanation how the code is being used with non-unitary `rotation_matrix` before changing anything...

babbush:
It's being used to transform between atomic orbitals and molecular orbitals. I'm not sure this adds much more information but it's discussed here: https://github.com/CrawfordGroup/ProgrammingProjects/tree/master/Project%2304

This is a standard step in pretty much all electronic structure codes.

fangzh-umich:
Hmm, my current understanding is that it's _part of_ a unitary matrix. Both the AO basis and the MO basis should be orthonormal, but the catch is that those are infinite, so for practical reasons we only consider the valence orbitals, which corresponds to taking some rows and columns from the infinite matrix. Am I right?

babbush:
I think you are saying that if we have an infinite number of AOs and an infinite number of MOs then the transformation between the bases would be unitary. I suppose that is true but it seems somewhat irrelevant from a practical point of view. The MOs we have are not a unitary rotation of the original AOs in this case. This is related to the fact that the AOs are not originally orthogonal and any attempt to orthogonalize them without adding more basis functions is, in general, not unitary. @jarrodmcc care to say more?

jarrodmcc:
The central part of the non-unitarity is the non-orthgonality of the atomic orbital and more general basis sets.  In the language of tensors, this gives the space a non-trivial metric (or overlap matrix).  As a result the transformation from AO's to MO's contains a non-unitary component that yields an orthonormal MO basis.  In the most general case, where the AO's have linear dependencies (singular metric), the transformation matrix can become rectangular, due to the need to project out redundant orbitals that cause numerical problems.  We've never used a basis big enough in QC yet for this to actually manifest, but in principle it can.

fangzh-umich:
Oh, of course the AOs are not all orthogonal, they are from different atoms.

I suppose this makes sense, but I would like to confirm with @kevinsung . I would also tend to change `rotation_matrix` to something like `transformation_matrix` in this case, since "rotation" implies unitarity.

kevinsung:
I'm confident that `general_basis_change` performs the same operation as the old code in the one- and two- body case. However, my understanding of how it should work comes from Section III.D.3 of the OpenFermion release paper. I think the derivation of the expression (23) for the coefficients in the new basis assumes that U is real and orthogonal (if U is complex unitary than we would have some complex conjugates, as we implemented in the new code). Both the old and new code transform coefficients according to (23) regardless of whether U is unitary or not, but I don't think this is the correct expression if U is not unitary; in that case we would need to compute the matrix inverse to get the correct coefficients. This wouldn't be hard to implement, but is not supported by the current code. Another thing is that we wouldn't want to compute the inverse if U actually is unitary, since matrix inversion is relatively expensive. I'm not sure what should be done in the case U is not invertible.

In short, I think the general (non-unitary) version of Eq. (23) in the paper involves entries of the inverse matrix of U; both the old and new code take advantage of the fact that U is unitary to perform this inversion in place (since the inverse is just U^\dagger).

PR #147, opened 2018-01-05
fix for nul description loading

conta877:
(fix for my fix)

this is the fix for [issue#14](https://github.com/quantumlib/OpenFermion-Psi4/issues/14) in Openfermion-psi4.

When the hdf5 file is generated from a run_psi4 call with no description, which defaults to '', upon loading at the end of run_psi4, its read as NUL(\x00).

In the following run, the *.inp file is generated with description='.' that raises a TypeError.


babbush:
Thanks for this! I'll take a closer look after tests finish. Unfortunately, Travis CI has been having a lot of problems today (not related to our project at all) and so tests are happening very slowly and sometimes randomly getting cancelled by the Travis server.

PR #149, opened 2018-01-08
Fermionoperator list init

idk3:


kevinsung:
I'm not sure I agree with this change; it seems to add only a very minor convenience. Perhaps you have a particular use case in mind, but I don't think it's a big deal to just convert the list to a tuple first. I think it makes sense that a term should be represented by an immutable object like a tuple or string, and not a list.

Here is a suggestion: a list input should contain multiple terms (as tuples of tuples), and they are all added to the FermionOperator with the same coefficient. For instance, in the Hubbard model, we could have a list of all the hopping terms and then pass them in with the tunneling coefficient. What do you think?

idk3:
It's true, I'm working on this for a PR for commutators ;)

I actually think we should be able to init from any array-like object, but that adding tuple and list is a good compromise / first step. This still keeps the internal immutable representation, but gives much greater flexibility in how we work with terms before we pass them to init. In general, I think we should avoid boilerplate code as much as possible - e.g. my continued annoyance with `normal_ordered(commutator(x))` showing up everywhere - but for `tuple(x)` in this case there's no cost to the conversion so the usual argument doesn't apply.

babbush:
I don't have strong feelings about this. Like Ian says, the internal representation remains immutable so I am inclined to approve the PR if it would make things more convenient for Ian. However, I am also a big proponent of maintaining as much similarity between QubitOperator and FermionOperator classes as possible (as I've mentioned in an issue, I think they should inherit from the same parent). So if this change is being made for FermionOperator I also think a similar change should be made to QubitOperator (prior to merging).

idk3:
@babbush addressed

src/openfermion/ops/_qubit_operator_test.py
```diff
@@ -55,6 +55,15 @@ def test_init_tuple(coefficient):
     assert qubit_op.terms[loc_op] == coefficient
 
 
+@pytest.mark.parametrize("coefficient", [0.5, 0.6j, numpy.float64(2.303),
```
babbush: Since when are we using pytest? It is a reasonable idea to move everything over to pytest but I was under the impression everything was using unittest now. Is there any place in this code where pytest is used? If not I am somewhat inclined to suggest removing the pytest import and also this test. I think it is a reasonable goal to move over to pytest at some point (it has many advantages) but I think that is sort of its own "issue" and until that point I'd prefer that we remain consistent. But I am open to other thoughts about this. If there are other places you can point to in OpenFermion where we use pytest, I'll accept this.

idk3: Most of this file uses the old pytest tests ;)

PR #173, opened 2018-01-18
Add QuadraticHamitonian and Slater determinant examples to demo

kevinsung:
I added a section to the demo to show off the QuadraticHamiltonian data structure and functionality to compile circuits to prepare their eigenstates.

I also re-ran the entire demo on my computer to get the order of the cells right, but it was also necessary because the output of the Hubbard model section changed due to bug fixes. However, this caused some numbers to be printed in a more ugly way; for instance, `-4.913` got changed to `-4.912999999999999`. For some reason this is due to my using Python 3 instead of Python 2, but running the demo with Python 2 had the even uglier consequence of giving me this warning:
```
/usr/lib/python2.7/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
```
So I decided to leave it as is.


gaming-hacker:
did this really get commited?  as i had to manually merge this pr onto my fork.

kevinsung:
I had no problem pulling it into my fork.

PR #178, opened 2018-01-19
add flag to allow overwrite in save_operator

idk3:


src/openfermion/utils/_operator_utils_test.py
```diff
@@ -147,6 +147,14 @@ def test_save_on_top_of_existing_operator_utils_error(self):
...
             save_operator(self.fermion_operator, self.file_name)
 
+    def test_overwrite_flag_save_on_top_of_existing_operator(self):
+        save_operator(self.fermion_operator, self.file_name)
+        save_operator(self.fermion_operator, self.file_name,
+                      allow_overwrite=True)
+        fermion_operator = load_operator(self.file_name)
+
+        self.assertTrue(fermion_operator.isclose(self.fermion_operator))
+
```
Spaceenter: Have another similar test for the case when allow_overwrite=False? (An error should be raised for that case.)

idk3: Adding one to verify! It looks to have dismissed your review though :(

PR #179, opened 2018-01-19
Implementation of Sz, S-, S+, S^{2} fermion ops

ncrubin:
Implented these useful observables at the level of fermion operators.
That way they can be transformed into the qubit representation by an
arbitrary fermion-to-qubit transform.

Unlike the previously implemented number operator these spin operators
require knowledge of the fermionic modes corresponding to spin-up or
spin-down.  As a first pass I have assumed a number occupation vector
corresponding to spin-less fermions where the even-odd pairs correspond
to a single spatial orbital (or site). This assumption is listed in the
warning of the docstring.  Generalizing to arbitrarying ordering of
alpha, beta spin-orbital sets would require the user to input these sets
when calling these functions.

Test implementations:

Implemented according to the style of the other tests.  But, further
accuracy tests can be included by first performing JW transform,
tensoring up, and then looking at the diagonal elements.  These elements
will correspond to the expected value of the operator for the bitstring.
For example, sz op diagonal elements should give [0, 0.5, -0.5, 0] with
the given orbital order convention.

I am also open to moving these somewhere else.  I'm not sure they belong in the FermionOps module.


kevinsung:
Perhaps you should import and use the functions `up_index` and `down_index` from `hamiltonians/_hubbard.py`? Also, maybe those functions should be moved somewhere else, like `_fermion_operator.py`.

ncrubin:
@kevinsung Good call.  Changed so that these functions take an up_map and a down_map which default to canonical (though not sure the described ordering of alpha/beta spins is truly common) mapping of site index to spin-orbtal index.

src/openfermion/ops/_fermion_operator.py
```diff
@@ -99,6 +117,139 @@ def number_operator(n_orbitals, orbital=None, coefficient=1.):
     return operator
 
 
+def sz_operator(n_spatial_orbitals, up_map=up_index, down_map=down_index):
+    """Return the sz operator.
+
+    SZ = 0.5 * \sum_{i = 1}^{n} n_{i \alpha} - n_{i, \beta}
```
babbush: forgot a comma after i in n_{i \alpha}.

src/openfermion/ops/_fermion_operator.py
```diff
@@ -99,6 +117,139 @@ def number_operator(n_orbitals, orbital=None, coefficient=1.):
...
+        operator += number_operator(n_spinless_orbitals, up_map(ni), 0.5) + \
+                    number_operator(n_spinless_orbitals, down_map(ni), -0.5)
+
+    return operator
+
+
+def s_plus_operator(n_spatial_orbitals, up_map=up_index, down_map=down_index):
+    """Return the s+ operator.
+
+    S+ = \sum_{i=1}^{n} a_{n, \alpha}^{\dagger}a_{n, \beta}
```
babbush: I am pretty sure you mean a_{i ...} not a_{n ...}.

src/openfermion/ops/_fermion_operator.py
```diff
@@ -99,6 +117,139 @@ def number_operator(n_orbitals, orbital=None, coefficient=1.):
...
+
+     Args:
+        n_spatial_orbitals: number of spatial orbitals (n_qubits // 2).
```
babbush: Actually, I think this would be (n_qubits + 1) // 2. Right?

src/openfermion/ops/_fermion_operator.py
```diff
@@ -99,6 +117,139 @@ def number_operator(n_orbitals, orbital=None, coefficient=1.):
...
+    return operator
+
+
+def s_squared_operator(n_spatial_orbitals):
+    """Return the s^{2} operator.
+
+    S^{2} = S- S+ + Sz( SZ + 1)
+
+     Args:
+        n_spatial_orbitals: number of spatial orbitals (n_qubits // 2).
```
babbush: same comment as above.

kevinsung:
I agree that these functions belong in a separate module. I think `utils/_fermion_utils.py` is good, though maybe we should be more descriptive and name it `_fermion_operator_utils.py`.

jarrodmcc:
I agree we should try to keep the _fermion_ops definitions as minimal as possible and probably build a fermion_utils file of some type. Like Ryan mentioned, number operator should go with it for consistency, though this may require repairing some errors that will occur from using it and probably break a bit of user code as well.

ncrubin:
Sounds good.  I will move the methods over to `_fermion_operator_utils.py` along with the associated test file.  FYI, I am planning on just moving over the new methods to prevent scope creep on this PR.  I searched for `number_operator`, `majorana_operator`, etc and they appear in other places.  I agree with @jarrodmcc and think that moving fermion ops to new methods might break stuff.  I'd rather have a PR dedicated to doing that properly.

src/openfermion/ops/_fermion_operator.py
ncrubin: Where does the +1 come from?  n_qubits is supposed to stand for the number of qubits which corresponds to the number of spin-orbitals.  Thus number of qubits / 2 is the number of spatial functions.

babbush: you wrote floor division

ncrubin: :sweat_smile: right you are.  :heavy_division_sign:

ncrubin:
@babbush so fun fact, moving the up/_down index functions into utils introduces a circular import.  So I decided to move the module under `hamiltonians` (which I think makes sense) and call it `_angular_momentum_generators.py`  Let me know if this is not satisfactory.

babbush:
How about just calling it _spin_operators.py?

babbush:
Okay, you're really going to hate me (I hate me for this) but I realize _spin_operators.py is a terrible name because it might mean operators that act on qubits. Last change I swear: can you call it _fermion_spin_operators.py?

ncrubin:
237siq

babbush:
@kevinsung @ncrubin any reason why this PR seems to have caused the docs to "fail" (see badge on front of repo). I have a test that automatically tries to make the docs and it didn't run into any problems. Also, the docs seem to render fine. So I'm not sure what the problem is...

ncrubin:
Seems to be an indentation error in `s_minus` and `s_plus`.  I'll fix now.

babbush:
Thanks! How could you tell?

ncrubin:
you provide the `Makefile` generated from sphinx which can be used to generate the docs.  When I ran `make html` which runs the default command `sphinx-build -b html -d _build/doctrees   . _build/html` the output indicated there was an indentation error in `s_minus...` and `s_plus_...`.  Also, when I opened the files in PyCharm the editor showed a slightly different indentation line.

PR #191, opened 2018-01-24
Added SymbolicOperator Class

bgimby:
This PR adds the file _symbolic_operator.py, but does not modify QubitOperator or FermionOperator in any way.

### DIFF
--- a/src/openfermion/ops/_symbolic_operator.py
+++ b/src/openfermion/ops/_symbolic_operator.py
@@ -16,9 +16,7 @@ import itertools
 
 import numpy
 from abc import abstractmethod
-
-
-EQ_TOLERANCE = 1e-12
+from openfermion.config import EQ_TOLERANCE
 
 
 class SymbolicOperatorError(Exception):
@@ -28,25 +26,28 @@ class SymbolicOperatorError(Exception):
 class SymbolicOperator(object):
     """
     The base class for QubitOperator and FermionOperator. All methods defined
-    here can be accessed from FermionOperator or QubitOperator objects.
-    This is an abstract class and objects of this type cannot be created,
-    only those of subclasses.
-
-    Subclasses are sums of terms of operators for a particular category of
-    particle. Subclasses support addition and multiplication with objects of
-    the same type.
+    here can be accessed from FermionOperator or QubitOperator objects. This
+    class and subclasses are sums of terms of operators for a particular 
+    category of particle. Subclasses support addition and multiplication with
+    objects of the same type.
 
     Attributes:
         terms (dict):
-            **key** (tuple of tuples): Each tuple represents a fermion term,
-            i.e. a tensor product of fermion ladder operators with a
-            coefficient. The first element is an integer indicating the
-            mode on which a ladder operator acts and the second element is
-            a bool, either '0' indicating annihilation, or '1' indicating
-            creation in that mode; for example, '2^ 5' is ((2, 1), (5, 0)).
+            **key** (tuple of tuples): Each tuple represents a term,
+            i.e. an (action, index) pair. The action can be any from a 
+            specified tuple of base operators that can be obtained in the
+            from the class method `actions()`.
             **value** (complex float): The coefficient of term represented by
             key.
     """
+
+    # Class attribute, I set it at random for testing right now
+    _actions = ("I", "A", "B", "C")
+
+    @classmethod
+    def actions(cls):
+        # _actions  should not be changed so only provide a getter
+        return _actions
 
     @abstractmethod
     def __init__(self):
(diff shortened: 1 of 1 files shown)


kevinsung:
Thanks Brendan, this is looking great.

I actually think that we can do most of the work of `__init__` in SymbolicOperator rather than its subclasses if we agree on the following abstraction: A SymbolicOperator stores a weighted sum of *terms* of the form (*action*, *index*), where *action* is drawn from a set defined by the subclass, and *index* is a nonnegative integer. For instance, in FermionOperator, *action* can be from {0, 1}, and for QubitOperator, *action* can be from {X, Y, Z}. Then we can require each subclass to have a class variable (attribute) called `actions` or something like that defining the set of valid actions, and the subclass should also have a helper function that does the analogue of `_parse_ladder_operator` for FermionOperator (it should now be a class method and have the same name across subclasses, maybe something like `parse_operator_string`). @babbush does this align with your thoughts on this class?

Brendan, you've already done a bunch of work though so I'm happy to do this part myself. Having SymbolicOperator have a functioning `__init__` would also make it easier to write a test file for it; I do think that it deserves to have its own test file.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,301 @@
...
+
+"""SymbolicOperator is a base class for QubitOperator and FermionOperator"""
+import copy
+import itertools
+
+import numpy
+from abc import abstractmethod
+
+
+EQ_TOLERANCE = 1e-12
```
babbush: you should import this from openfermion.config. Do not set it here.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,301 @@
...
+
+class SymbolicOperatorError(Exception):
+    pass
+
+
+class SymbolicOperator(object):
+    """
+    The base class for QubitOperator and FermionOperator. All methods defined
+    here can be accessed from FermionOperator or QubitOperator objects.
+    This is an abstract class and objects of this type cannot be created,
```
babbush: Why can't they be created? It seems like being able to create them would be necessary for testing.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,301 @@
...
+    only those of subclasses.
+
+    Subclasses are sums of terms of operators for a particular category of
+    particle. Subclasses support addition and multiplication with objects of
+    the same type.
+
+    Attributes:
+        terms (dict):
+            **key** (tuple of tuples): Each tuple represents a fermion term,
```
babbush: This description is a bit too narrow since it seems to describe FermionOperators. I think we want something more general which would also encompass QubitOperators. Kevin had a good suggestion to make the abstract class (index, action), etc.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,301 @@
...
+
+    @classmethod
+    def zero(cls):
+        """
+        Returns:
+            additive_identity (SymbolicOperator):
+                A symbolic operator o with the property that o+x = x+o = x for
+                all fermion operators x.
+        """
+        # Maybe throw a TypeError if called using SymbolicOperator? Or maybe
```
babbush: I think it is okay to let users initialize SymbolicOperator. I don't know why they would, but it seems like a strange trap for them to fall into unless they know what they are doing.

babbush:
@kevinsung yes, I agree with your comments completely!

bgimby:
Okay, I made most of the requested changes. I imported `EQ_TOLERANCE` from config, changed some of the documentation, and added a class attribute `_actions` and a class method getter `actions()` for it. 

I probably won't have too much time to work on this for the next few weeks, but in the meantime @kevinsung should be able to work on implementing the required methods to make `SymbolicOperator` objects initializable (for now, I left all the necessary methods abstract since they are not implemented. You can just remove the decorator once they are implemented).

babbush:
Great, thanks! The main thing now is that we'll need tests and decent coverage to merge the PR.

bgimby:
Tests are definitely needed to simplify testing `QubitOperator` and `FermionOperator`, but the version of my fork that combines both of my PRs has good coverage with existing tests. My results are here:

https://hastebin.com/wagemipote.erl

If you want, I can combine the PRs at this point so the automated testing works.

babbush:
The main thing is that we have a rule against merging pull requests that lower test coverage. Merging this would  lower test coverage by about 1%, which is far too much. Unfortunately, I think we are going to need full tests on just this class. Fortunately, it sounds like @kevinsung is willing to help with that.

kevinsung:
Ok will write the `__init__` function and tests and finish up this PR.

PR #150, opened 2018-01-08
iadd isub __contains__ improvement

idk3:

Issue #43, opened 2017-10-14
QubitOperator and FermionOperator should inherit from same parent


babbush:
In a very early version of this library (known then as OpenPQRS), we had a parent class from which both QubitOperator and FermionOperator inherited. This was an obvious design choice because the classes are extremely similar and differ in only a few places. However, when making FermiLib (the predecessor to OpenFermion) we decided that the QubitOperator class should be embedded in ProjectQ and so we removed this inheritance structure. But in OpenFermion, we have again moved both QubitOperator and FermionOperator into the same library.

There are many advantages to having both of these classes inherit from the same parent. For one thing, we'd need many fewer tests to cover the same functionality. More importantly, there are a variety of methods I would like to add to these classes but it would be annoying to need to add essentially the same function to both. For instance, I would like to add the __abs__ method to both classes which returns the L1 norm of the operators by summing the absolute value of the coefficients. This should be added to the parent class. I think the parent should be called SymbolicOperator.

This is not an easy change it will likely take significant effort. However, fortunately, one will likely only need to change the files which define the QubitOperator and FermionOperator and test them since all other parts of the library will still import QubitOperator and FermionOperator the same.

idk3:
As a smaller part of this - FermionOperator has the static methods identity and zero, but QubitOperator doesn't have these methods.

bgimby:
Hey, I was thinking of working on this issue. 

Reading through the definitions of `FermionOperator` and `QubitOperator`, I think once this is completed, behavior of `compress` should be changed to remove small real parts of complex coefficients, as it does for small imaginary coefficients. At present, it would reduce `1+i*1e-15` to `1`, but leave `i+1e-15` unchanged, which I don't see a reason for. 

Also, compress may be able to be done in place to save on unnecessary construction of new terms, although it may not be a bottleneck in any common applications, so I'm not sure if it'd be worth it.

Anyway, I'll get started on implementing a `SymbolicOperator` base class and making the others inherit from it.

babbush:
Awesome! I agree completely with you that compress should also remove the small real components.

If you'd like, email me at my github username @google.com and I can send you some very old code that might be helpful. As I said in comment, in the early days of the code (before this GitRepo), we did have things setup this way. So I do have some old code implementing the SymbolicOperator that also has a bunch of tests that come with it. Of course, a number of things have changed since then but it still might be useful.

kevinsung:
[...]

PR #192, opened 2018-01-24
Changed QubitOperator and FermionOperator to inherit from SymbolicOperator

bgimby:
This PR includes the changes to QubitOperator and FermionOperator, but does not include the SymbolicOperator class.


src/openfermion/ops/_fermion_operator.py
```diff
@@ -15,9 +15,10 @@
 from openfermion.config import *
 from future.utils import iteritems
 import numpy
+from _symbolic_operator import *
```
babbush: You will need to write from ._symbolic_operator import *. We tend to discourage the "import *" thing though - better if you can write explicitly what you are importing.

src/openfermion/ops/_qubit_operator.py
```diff
@@ -15,6 +15,7 @@
 import itertools
 
 import numpy
+from _symbolic_operator import *
```
babbush: Same comment as above

src/openfermion/ops/_qubit_operator.py
```diff
@@ -15,6 +15,7 @@
...
 
 
 EQ_TOLERANCE = 1e-12
```
babbush: Please import from config

src/openfermion/ops/_fermion_operator.py
bgimby: `from ._symbolic_operator import SymbolicOperator, SymbolicOperatorError` should work I think?

babbush: I think so, yes

bgimby:
I made the requested changes. As I mentioned in my other PR, I'll probably be a bit busy the next couple of weeks, but I'll check in often to answer questions or see if you need anything else from me.

babbush:
Sounds good. Clearly this PR will need to wait until https://github.com/quantumlib/OpenFermion/pull/191 is merged.

kevinsung:
I'm willing to finish this PR. @bgimby would you mind giving me write access to your branch so I can push commits to it?

bgimby:
@kevinsung You should get an invite email.

@idk3 I ended up switching `__iadd__` to using `EQ_TOLERANCE ** 2` for now because I'm not sure how else to continue to pass the jellium tests and continue to drop negligible terms.

kevinsung:
@bgimby It seems that you are the only one having problems getting the Jellium test to pass. The coveralls check would not pass if any tests failed on their server. Can you please restore the functionality of `__iadd__` to the original version? Just to be sure, you didn't change the functionality of any other methods in SymbolicOperator, did you?

bgimby:
See Issue #43 for discussion.

kevinsung:
@bgimby I reverted the behavior of `__iadd__` because I think we should fix the problem before merging this PR.

kevinsung:
@babbush @idk3 @bgimby I think this PR is ready to go except for the problem we have been discussing. Namely, the `__iadd__` method in the old QubitOperator class did not use `EQ_TOLERANCE`, and updating it to do so (by inheriting from SymbolicOperator) causes a test in `hamiltonians/_jellium_test.py` to fail.

kevinsung:
I also ran `performance_benchmarks.py` a bunch of times and the only noticeable difference was that the benchmark `InteractionOperater.jordan_wigner_transform()` runs faster with the new code (about 7.5 seconds compared to 9.2 seconds on my laptop).

kevinsung:
@bgimby I added your name to the README and NOTICE. Are you okay with that? Please check that it's correct.

src/openfermion/ops/__init__.py
```diff
@@ -10,6 +10,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from ._symbolic_operator import SymbolicOperator
```
babbush: I suppose it is necessary to import this before FermionOperator? If so then you get a pass on the alphabetical imports...

src/openfermion/ops/_fermion_operator.py
```diff
@@ -226,131 +202,11 @@ class FermionOperator(object):
     3.17 2^ 1 - 66.2 * 8^ 7 6^ 2
     The Fermion Operator class overloads operations for manipulation of
     these objects by the user.
-
-    Attributes:
```
babbush: Are you completely getting rid of the attributes docs? Or is that part of the docs going to be inherited or something?

src/openfermion/ops/_qubit_operator.py
```diff
@@ -65,167 +60,11 @@ class QubitOperator(object):
...
-                ham2 = QubitOperator('X0 Y3', 0.5)
-                ham2 += 0.6 * QubitOperator('X0 Y3')
-
-        Note:
-            Adding terms to QubitOperator is faster using += (as this is done
-            by in-place addition). Specifying the coefficient in the __init__
-            is faster than by multiplying a QubitOperator with a scalar as
-            calls an out-of-place multiplication.
-
-        Args:
```
babbush: Shouldn't we keep these docs?

babbush:
Evidently this is passing tests. Do we still have a problem with jellium / iadd? I haven't been able to follow the conversation around this iadd / tolerance / jellium concern. Is this related to Ian's PR? What precisely is failing?

src/openfermion/ops/__init__.py
kevinsung: Yup, exactly.

src/openfermion/ops/_fermion_operator.py
kevinsung: All the attributes are inherited. I'm not sure what our convention about documenting subclasses is (if we have one) but now that I think of it, it would be good to document exactly what `actions` and `action_strings` are for subclasses of SymbolicOperator.

src/openfermion/ops/_qubit_operator.py
kevinsung: Yes, in hindsight I went overboard with trimming the docs. I'll add some back in.

src/openfermion/ops/_fermion_operator.py
babbush: These are the most important classes in the whole library. I think it is very likely that many people actually take a look at these docs. So I think we should error on the side of having excessive documentation here.

kevinsung:
@babbush Basically the old `__iadd__` in QubitOperator did not use `EQ_TOLERANCE` and was instead checking for exactly equality with zero. Updating it caused the bug I fixed in #198 to take effect. Apparently it is not related to the thing @idk3 is having trouble with because they concern different Jellium tests.

I'll update the documentation of these classes and then I think we should be good to go with this PR.

kevinsung:
Ok I also had to port @idk3's changes to `__iadd__` and `__isub__` from #150 to SymbolicOperator. I added the Examples and Notes from the original `__init__` functions to the class docstrings for the subclasses. I think we're ready.

babbush:
Tests seem to be failing still.

kevinsung:
Hmm, now it's failing on `JelliumTest.test_kinetic_integration`. @idk3 you mentioned that you made some change before which caused this test to fail. What was it? Did you figure anything out about it?

babbush:
I believe it is failing in a different way than what Ian had seen in the past. Whereas what Ian saw in the past was a numerical issue, there now seems to be some rather profound implementation problem since two arrays are now different sizes which were previously the same size!

idk3:
No, the different length issue is the one I faced. The problem is that the numbers of qubits aren't judged to be the same so `jordan_wigner` and other functions give different lengths for the spectra vector. A simple but dirty way to fix this is by directly setting `x.terms[((n, 1), (n, 0))] = 0.0` - then it thinks the operator does have n+1 qubits.

kevinsung:
I found that using the versions of `__iadd__` and `__isub__` before #150 make the tests pass. I can't find a bug in the new versions of these methods though!

kevinsung:
Ah @idk3 the fact that it's the same issue explains this.

kevinsung:
I proposed a fix at #199; if we merge that this should become mergeable.

babbush:
Travis is failing. It seems that the PR that was merged with this trace norm business is causing problems. I think that perhaps the automatic merging is to blame. See if you can fix this so tests pass. Obviously, in the next PR that I just opened, I move part of this over into SymbolicOperator.

babbush:
I think it just unhappy that certain import statements were removed.

babbush:
I tried to fix it myself. Let's see how it goes...

PR #193, opened 2018-01-26
Fix bug in FermionOperator multiplication

kevinsung:
Duplicated terms were not being accounted for; see the test I added.

### DIFF
--- a/src/openfermion/ops/_fermion_operator.py
+++ b/src/openfermion/ops/_fermion_operator.py
@@ -466,7 +466,11 @@ class FermionOperator(object):
                     product_operators = left_term + right_term
 
                     # Add to result dict.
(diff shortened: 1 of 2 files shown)


src/openfermion/ops/_fermion_operator.py
```diff
@@ -466,7 +466,11 @@ def __imul__(self, multiplier):
                     product_operators = left_term + right_term
 
                     # Add to result dict.
-                    result_terms[tuple(product_operators)] = new_coefficient
+                    product_operators = tuple(product_operators)
+                    if product_operators in result_terms:
+                        result_terms[product_operators] += new_coefficient
+                    else:
+                        result_terms[product_operators] = new_coefficient
```
idk3: `result_terms.get(product_operators)` would be cleaner here

Issue #177, opened 2018-01-19
Expand string initializations


babbush:
As discussed in https://github.com/quantumlib/OpenFermion/issues/122 it would be helpful if sums of FermionOperators and QubitOperators could be initialized from their string representation so that
my_fermion_operator - FermionOperator(str(my_fermion_operator)) = 0 and
my_qubit_operator - QubitOperator(str(qubit_operator)) = 0.
This should be easy for a beginner to implement.

max-radin:
I'll work on implementing this. I'm new to OpenFermion and would like to get familiar with the code.

kevinsung:
I had this issue in mind while completing #194 so I did it in a way that made this task easy to incorporate. Basically you just fill in the body of `_long_string_init` in `ops/_symbolic_operator.py`. @max-radin, are you okay with doing it this way?

max-radin:
Yes that makes sense to me. I have written some code for long string initialization in FermionOperator that I can move into `_long_string_init`.

kevinsung:
Great. Keep in mind that the task now is basically to reverse-engineer the `__str__` method of SymbolicOperator, which now uses the attributes `action` and `action_string`, so your implementation will also need to use these attributes of SymbolicOperator. Let me know if you have any questions.

babbush:
Closed by #201

babbush:
`my_operator = FermionOperator('1.7 [3^ 2] - 8 [4^]')`
seems to throw and error whereas
`my_operator = FermionOperator('1.7 [3^ 2] + 8 [4^]')`
and
`my_operator = FermionOperator('1.7 [3^ 2] -8 [4^]')`
and
`my_operator = FermionOperator('1.7 [3^ 2] + -8 [4^]')`
do not. I guess subtraction is not supported as expected? @max-radin can you assist with this?

babbush:
Also, it seems to always assume numbers are complex, for instance
`my_operator = FermionOperator('-1.7 [3^ 2] + 8.0 [4^]')
print(my_operator)`
returns
`(-1.7+0j) [3^ 2] +
(8+0j) [4^]`

babbush:
Also, I am not sure how one is supposed to provide complex coefficients.
`my_operator = FermionOperator('(2.3 + 1.7j) [3^ 2]')`
throws an error.

max-radin:
[...]

Issue #175, opened 2018-01-18
h5py warning


gaming-hacker:
hello,
OSX 10.12.6, clang-6.0, python2.7,numpy 1.14.0, scipy 1.0.0

installed openfermion from github repo as well as the plugins.

running the basic example
```
from openfermion.ops import FermionOperator, hermitian_conjugated
from openfermion.transforms import jordan_wigner, bravyi_kitaev
from openfermion.utils import eigenspectrum

# Initialize an operator.
fermion_operator = FermionOperator('2^ 0', 3.17)
fermion_operator += hermitian_conjugated(fermion_operator)
print(fermion_operator)
```

the following warning happens

.../h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.

can someone confirm this solution is correct?
3.17 [2^ 0] +
3.17 [0^ 2]

kevinsung:
The output you got is correct.

This is the same warning that I mentioned getting with Python2.7 in #173 . My details are
> Arch Linux with kernel version 4.14.11
Python 2.7.14
Jupyter notebook server version 5.0.0
numpy 1.14.0
scipy 1.0.0
h5py 2.7.1

I also get the warning when simply running `import openfermion`.

babbush:
This seems to have come up very recently. It is just a warning, not a bug, and the code output still looks correct. This is likely due to something that changed recently in one of our dependancies. I'll take a look briefly tonight but may need to return to this tomorrow or Saturday - it doesn't seem critical (albeit ugly). Kudos to anyone else who tracks it down.

kevinsung:
I just found out that I get the same warning from simply executing "import h5py". So I think this has nothing to do with OpenFermion.

kevinsung:
From looking at https://github.com/h5py/h5py/issues/961, it appears that the h5py developers are aware of this issue and are currently working on a fix at https://github.com/h5py/h5py/pull/926.

babbush:
Hmm, thanks Kevin. Unfortunately they haven't updated that PR in ten days. The funny thing though is that I cannot reproduce this error on my mac. @kevinsung @gaming-hacker What version of h5py are you running?

kevinsung:
I'm using h5py 2.7.1

babbush:
[...]

PR #194, opened 2018-01-30
Add SymbolicOperator and tests

kevinsung:
I couldn't add commits to #191 because I don't have write access to the main repo, so I'm replacing it with this PR. Some notes on my implementation:
- I added four attributes that all subclasses need to define; see their descriptions in the docstring for SymbolicOperator. So far, the only thing I used the attribute `different_indices_commute` for is to see if we can reorder the factors in terms so that the indices are in ascending order. To get a functioning subclass it's enough to set these four attributes; I removed the uses of abstract base class methods (from the module `abc`) which were introduced in #191.
- The implementation of `__str__` preserves the current string representations for FermionOperators, but would slightly alter those of QubitOperators. Namely, operators would be surrounded in square brackets, and the identity term would be represented as `[]` instead of `I`. For instance, we would have
````
2.0 [] +
0.6 [Y1 Y3 Z8] +
0.5 [X1 Y3 Z8]
````
instead of
````
2.0 I +
0.6 Y1 Y3 Z8 +
0.5 X1 Y3 Z8
````
Of course, we can override this behavior but using the unified implementation will make it possible for #177 to be implemented only in SymbolicOperator.
- On that note, a unified implementation of #177 now amounts to writing the method `_long_string_init` which I left unimplemented.
- All the tests are pretty much straight from `_fermion_operator_test.py` and `_qubit_operator_test.py`. In the test file, the tests for `DummyOperator1` were from `_fermion_operator_test.py` and the tests for `DummyOperator2` were from `_qubit_operator_test.py` (modified to remove the dependence on `pytest`).
- I replaced all uses of `SymbolicOperatorError` with `ValueError` because I couldn't tell when one should be used over the other.
- It appears that most of the work that will go into defining a subclass is in overriding `__imul__`.
- I started calling the things that terms are made of "factors", but then realized later on that these were sometimes called "local operators" which is perhaps a better term. I still document them as "factors" but we can change this.

### DIFF
--- a/src/openfermion/ops/_symbolic_operator.py
+++ b/src/openfermion/ops/_symbolic_operator.py
@@ -16,9 +16,7 @@ import itertools
 
 import numpy
 from abc import abstractmethod
-
-
-EQ_TOLERANCE = 1e-12
+from openfermion.config import EQ_TOLERANCE
 
 
 class SymbolicOperatorError(Exception):
@@ -28,25 +26,28 @@ class SymbolicOperatorError(Exception):
 class SymbolicOperator(object):
     """
     The base class for QubitOperator and FermionOperator. All methods defined
-    here can be accessed from FermionOperator or QubitOperator objects.
-    This is an abstract class and objects of this type cannot be created,
-    only those of subclasses.
-
-    Subclasses are sums of terms of operators for a particular category of
-    particle. Subclasses support addition and multiplication with objects of
-    the same type.
+    here can be accessed from FermionOperator or QubitOperator objects. This
+    class and subclasses are sums of terms of operators for a particular 
+    category of particle. Subclasses support addition and multiplication with
+    objects of the same type.
 
     Attributes:
         terms (dict):
-            **key** (tuple of tuples): Each tuple represents a fermion term,
-            i.e. a tensor product of fermion ladder operators with a
-            coefficient. The first element is an integer indicating the
-            mode on which a ladder operator acts and the second element is
-            a bool, either '0' indicating annihilation, or '1' indicating
-            creation in that mode; for example, '2^ 5' is ((2, 1), (5, 0)).
+            **key** (tuple of tuples): Each tuple represents a term,
+            i.e. an (action, index) pair. The action can be any from a 
+            specified tuple of base operators that can be obtained in the
+            from the class method `actions()`.
             **value** (complex float): The coefficient of term represented by
             key.
     """
+
+    # Class attribute, I set it at random for testing right now
+    _actions = ("I", "A", "B", "C")
+
+    @classmethod
+    def actions(cls):
+        # _actions  should not be changed so only provide a getter
+        return _actions
 
     @abstractmethod
     def __init__(self):
@@ -72,8 +73,6 @@ class SymbolicOperator(object):
                 A symbolic operator o with the property that o+x = x+o = x for
                 all fermion operators x.
         """
-        # Maybe throw a TypeError if called using SymbolicOperator? Or maybe
-        # just don't expose SymbolicOperator to users
 
         return cls(term=None)
 
@@ -81,7 +80,7 @@ class SymbolicOperator(object):
     def identity(cls):
         """
         Returns:
-            multiplicative_identity (FermionOperator):
+            multiplicative_identity (SymbolicOperator):
                 A symbolic operator u with the property that u*x = x*u = x for
                 all fermion operators x.
         """
--- a/src/openfermion/ops/_symbolic_operator.py
+++ b/src/openfermion/ops/_symbolic_operator.py
@@ -15,7 +15,7 @@ import copy
 import itertools
 
 import numpy
-from abc import abstractmethod
+from abc import ABCMeta, abstractmethod
 from openfermion.config import EQ_TOLERANCE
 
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/openfermion/ops/_symbolic_operator.py [hunk 1]
```diff
@@ -0,0 +1,520 @@
...
+    and the possible values for `action` are determined by the subclass.
+    For instance, for the subclass FermionOperator, `action` can be 1 or 0,
+    indicating raising or lowering, and for QubitOperator, `action` is from
+    the set {'X', 'Y', 'Z'}.
+    The coefficients of the terms are stored in a dictionary whose
+    keys are the terms.
+    SymbolicOperators of the same type can be added or multiplied together.
+
+    Attributes:
+        actions (tuple): A tuple of objects representing the possible actions.
```
babbush: I wonder if it would be better if everything other than the terms attribute were a private attribute that is accessed with getters/setters. I am not necessarily of that belief, but it is worth thinking about. @jarrodmcc do you have any thoughts?

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,520 @@
...
+        elif isinstance(term, str):
+            term = self._parse_string(term)
+        # Invalid input type
+        else:
+            raise ValueError('term specified incorrectly.')
+
+        # Add the term to the dictionary
+        self.terms[term] = coefficient
+
+    def _long_string_init(self, term, coefficient):
```
babbush: Is there a reason why this needs to be implemented differently for FermionOperator and QubitOperator? Otherwise, I feel we should implement the method now.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,520 @@
...
+        Initialization from a long string representation, i.e., term
+        is a string such as '1.5 [2^ 3] + 2.4 [3^ 0]'.
+        """
+        raise NotImplementedError('Initializing a SymbolicOperator from a '
+                                  'long string is not yet supported.')
+
+    def _parse_sequence(self, term):
+        """Parse a term given as a sequence type (i.e., list, tuple, etc.).
+
+        i.e. For QubitOperator:
```
babbush: I think you mean "e.g." not "i.e."

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,520 @@
...
+            if self.different_indices_commute:
+                term = sorted(term, key=lambda factor: factor[0])
+
+            # Return a tuple
+            return tuple(term)
+
+    def _parse_string(self, term):
+        """Parse a term given as a string.
+
+        i.e. For FermionOperator:
```
babbush: You mean "e.g." not "i.e."

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -0,0 +1,520 @@
...
+                    tmp_string += '{}{} '.format(action_string, index)
+                else:
+                    tmp_string += '{}{} '.format(index, action_string)
+            string_rep += '{}] +\n'.format(tmp_string.strip())
+        return string_rep[:-3]
+
+    def __repr__(self):
+        return str(self)
+
+    def __imul__(self, multiplier):
```
babbush: So is the idea that for QubitOperator we will overwrite this class? Because in that case there is a more optimized way to multiply which is based on knowledge of the fact that the QubitOperator terms are sorted by tensor factor. I think that is fine, I am just trying to understand.

kevinsung:
Actually, I did work off of @bgimby's fork. I just couldn't add commits to the PR he created because that required me to push directly to his branch, which I don't have write access to. Creating a PR gives write access to maintainers of the main repository. The first two commits of this PR belong to him so I think he will get credit on the Contributor's page.

@bgimby was the one who ported most of the magic methods. It looks to me that they are unmodified; do you have anything else to say about that @bgimby? I made sure that `__imul__` and `__iadd__` work exactly the same. I promise to check the performance benchmarks after I complete the inheritance structure.

src/openfermion/ops/_symbolic_operator.py
kevinsung: No, there is no reason. It's just that @max-radin said he would do it at #177 so I figured I'd leave it to him. Perhaps we should post there asking if he's still willing to do it?

src/openfermion/ops/_symbolic_operator.py
kevinsung: Ok I will change these.

src/openfermion/ops/_symbolic_operator.py
kevinsung: Yes, the idea is that QubitOperator will overwrite this method using its existing implementation.

bgimby:
I did not modify any of the magic methods. It's possible that there is extra overhead in calling a base class's method from a subclass, though I'm not sure. 

If there are any performance losses, or indeed any performance issues I'll be happy to work on them when I have the time- I like working on optimization, though I haven't had a chance to do much in python.

babbush:
In that case, I think we're pretty much good to go. Kevin, let me know when we're ready. Also, @kevinsung did you have a chance to figure out why the docs were failing from your last PR? If you have any idea it would be good to also fix that in this PR.

kevinsung:
@babbush I think this PR is ready. No, I couldn't figure out why the docs are failing; I tried building the docs on my computer and did not get any errors or warnings besides the h5py one mentioned in #175 . Currently I have no idea how to diagnose the problem but I'll keep it in mind.

src/openfermion/ops/_symbolic_operator.py
babbush: Yeah let's ask him about that.

babbush:
Somehow the docs are unbroken. I think it was some glitch in read-the-docs.

PR #195, opened 2018-01-30
Trace norm function for QubitOperator

obriente:
Added functions to calculate the trace norm of a QubitOperator and renormalize so that the trace norm is 1.


src/openfermion/ops/_qubit_operator.py
```diff
@@ -464,3 +464,20 @@ def __str__(self):
 
     def __repr__(self):
         return str(self)
+
+    def tracenorm(self):
+        """Calculate the trace norm (sqrt(Tr[O^O])) of the operator"""
```
jarrodmcc: I think it's clear from context but maybe still write O^\dagger O instead of O^O to be clear what's happening

src/openfermion/ops/_qubit_operator.py
```diff
@@ -464,3 +464,20 @@ def __str__(self):
...
+            return 0
+        norm2 = 0
+        for coeff in self.terms.values():
+            norm2 += numpy.abs(coeff)**2
+        return numpy.sqrt(norm2)
+
+    def renormalize(self):
+        """Fix the trace norm of an operator to 1"""
+        if not self.terms:
+            raise ZeroDivisionError
```
jarrodmcc: I think it would be good to include a descriptive error message with this raise.  E.g. ZeroDivisionError('Cannot renormalize empty or 0 operator') and perhaps check for the same below at the same time after calculating norm, as I think a similar error could arise if the operator happened to have 0 coefficients.  I think we have compress routines that eliminate 0 operators, but they aren't always called automatically if I recall correctly.

src/openfermion/ops/_qubit_operator.py
obriente: Definitely a good point. Updated.

src/openfermion/ops/_qubit_operator.py
obriente: Sounds good. As there shouldn't be an error calculating the norm of a zero operator, I've shifted the error to after the norm is found, and added the message you suggested.

PR #196, opened 2018-01-30
Added possibility for QPE to have rotations controlled by an ancilla

obriente:
Minor update to the trotter_exp_to_qgates file, allowing the Rz rotations to be controlled by an ancilla for QPE purposes. Also fixed a bug where passing the identity operator would cause an error.

Now passing the identity passes for time evolution (i.e. it doesn't return a phase gate). Not sure what is preferable here.

### DIFF
--- a/src/openfermion/utils/_trotter_exp_to_qgates.py
+++ b/src/openfermion/utils/_trotter_exp_to_qgates.py
@@ -156,7 +156,8 @@ def trotter_operator_grouping(hamiltonian,
 
 
 def pauli_exp_to_qasm(qubit_operator_list,
-                      evolution_time=1.0):
+                      evolution_time=1.0,
+                      ancilla=None):
     """Exponentiate a list of QubitOperators to a QASM string generator.
 
     Exponentiates a list of QubitOperators, and yields string generators in
@@ -167,6 +168,8 @@ def pauli_exp_to_qasm(qubit_operator_list,
             QubitOperators to be exponentiated
         evolution_time (float): evolution time of the operators in
             the list
+        ancilla (string or None): if any, an ancilla qubit to perform
+            the rotation conditional on (for quantum phase estimation)
 
     Yields:
         string
@@ -220,8 +223,12 @@ def pauli_exp_to_qasm(qubit_operator_list,
             ret_list = ret_list + cnots1
 
             # 3. Rotation (Note kexp & Ntrot)
-            ret_list = ret_list + ["Rz {} {}".format(
-                term_coeff * evolution_time, qids[-1])]
+            if ancilla is not None:
+                ret_list = ret_list + ["C-Phase {} {} {}".format(
+                    term_coeff * evolution_time, qids[-1], ancilla)]
+            else:
+                ret_list = ret_list + ["Rz {} {}".format(
+                    term_coeff * evolution_time, qids[-1])]
 
             # 4. Second set of CNOTs
             ret_list = ret_list + cnots2
@@ -237,7 +244,8 @@ def trotterize_exp_qubop_to_qasm(hamiltonian,
                                  trotter_number=1,
                                  trotter_order=1,
                                  term_ordering=None,
-                                 k_exp=1.0):
+                                 k_exp=1.0,
+                                 ancilla=None):
     """Trotterize a Qubit hamiltonian and write it to QASM format.
 
     Assumes input hamiltonian is still hermitian and -1.0j has not yet been
@@ -266,5 +274,6 @@ def trotterize_exp_qubop_to_qasm(hamiltonian,
                                                     trotter_order,
                                                     term_ordering,
                                                     k_exp):
-        for exponentiated_qasm_string in pauli_exp_to_qasm([trotterized_op]):
+        for exponentiated_qasm_string in pauli_exp_to_qasm([trotterized_op],
+                                                           ancilla=ancilla):
             yield exponentiated_qasm_string
--- a/src/openfermion/utils/_trotter_exp_to_qgates.py
+++ b/src/openfermion/utils/_trotter_exp_to_qgates.py
@@ -12,6 +12,7 @@
 """Module to perform Trotter-Suzuki decompositions to output as circuits."""
 
 from openfermion.ops import QubitOperator
+from openfermion.utils import count_qubits
 import numpy
 import copy
 import collections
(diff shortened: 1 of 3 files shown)


src/openfermion/utils/_trotter_exp_to_qgates.py
```diff
@@ -167,6 +168,10 @@ def pauli_exp_to_qasm(qubit_operator_list,
             QubitOperators to be exponentiated
         evolution_time (float): evolution time of the operators in
             the list
+        qubit_list: a list of names for qubits (otherwise they will be labeled
```
jarrodmcc: Can we include the type a bit more explicitly?  Is it a list of strings with length of the number of qubits for example?

src/openfermion/utils/_trotter_exp_to_qgates.py
```diff
@@ -253,6 +272,8 @@ def trotterize_exp_qubop_to_qasm(hamiltonian,
         term_ordering (list of (tuples of tuples)): list of tuples
             (QubitOperator terms dictionary keys) that specifies
             order of terms when trotterizing
+        qubit_list: a list of names for qubits (otherwise they will be labeled
```
jarrodmcc: Same comment as above about more explicit type descrtiption

src/openfermion/utils/_trotter_exp_to_qgates.py
jarrodmcc: Perhaps even a little elaboration as to its use and sanity checking on its size / conditions for proper operation.

src/openfermion/utils/_trotter_exp_to_qgates.py
```diff
@@ -220,8 +228,17 @@ def pauli_exp_to_qasm(qubit_operator_list,
             ret_list = ret_list + cnots1
 
             # 3. Rotation (Note kexp & Ntrot)
-            ret_list = ret_list + ["Rz {} {}".format(
-                term_coeff * evolution_time, qids[-1])]
+            if ancilla is not None:
+                if len(qids) > 0:
```
jarrodmcc: Instead of checking qubit_list in a few places, it might be more compact to introduce one concept of it at the top of the function.  For example, if it's none, qubit_list is initialized to the ordered list of qubits that one would normally expect.  This should make the logic follow a bit more easily I think.

src/openfermion/utils/_trotter_exp_to_qgates_test.py
```diff
@@ -229,3 +230,124 @@ def test_qasm_string_XYZ(self):
...
+
+    def test_qasm_string_controlled_identity(self):
+        qasmstr = str(count_qubits(self.op_id)+1) + "\n"
+
+        # Write each QASM operation
+        qasmstr += "\n".join(
+            trotterize_exp_qubop_to_qasm(self.op_id, ancilla='ancilla'))
+
+        strcorrect = '''1
+Rz 1.0 ancilla'''
```
jarrodmcc: This test is a little confusing.  Typically if we are controlling on an operation, we hope to see something like
C-op angle control target
but we get an unusual result because it's a controlled identity where the control is the same as the target.  Do we want to allow such a behavior or raise an exception if the control is identical to the target?

src/openfermion/utils/_trotter_exp_to_qgates_test.py
```diff
@@ -229,3 +230,124 @@ def test_qasm_string_XYZ(self):
...
+            trotterize_exp_qubop_to_qasm(self.opA, ancilla='ancilla',
+                                         qubit_list=qubit_list))
+
+        # Correct string
+        strcorrect = '''5
+H q0
+Rx 1.5707963267948966 q3
+CNOT q0 q1
+CNOT q1 q3
+C-Phase 0.5 q3 ancilla
```
jarrodmcc: I think more typically QASM commands use
c-op angle control target
rather than
c-op angle target control

src/openfermion/utils/_trotter_exp_to_qgates.py
```diff
@@ -220,8 +228,17 @@ def pauli_exp_to_qasm(qubit_operator_list,
...
+                    ret_list = ret_list + ["C-Phase {} {} {}".format(
+                        term_coeff * evolution_time, qids[-1], ancilla)]
+                else:
```
jarrodmcc: I think these swap the more typical convention of control, target vs target, control

src/openfermion/utils/_trotter_exp_to_qgates.py
obriente: Completely fair point. I'm not entirely happy with the way this is implemented to be honest. The issue is that if I define a qubit in a piece of code I often want it to have a label that is a valid variable name, but when I define a qubit or fermion operator in a system I almost always just label with a number. This was my way of trying to bridge between the two, but it feels a bit hacky.

As it is, I'll insert a sanity check on the size with a detailed error message and add a more detailed description. But there may be a better way to deal with this.

src/openfermion/utils/_trotter_exp_to_qgates.py
obriente: This is a much better way of dealing with this, thanks!

src/openfermion/utils/_trotter_exp_to_qgates.py
obriente: Fair point, my mistake. It's not an issue for a C-Phase gate as it's symmetric, but we should definitely try to keep a convention.

src/openfermion/utils/_trotter_exp_to_qgates.py
obriente: Copied the description from the above string

src/openfermion/utils/_trotter_exp_to_qgates_test.py
obriente: Fixed.

src/openfermion/utils/_trotter_exp_to_qgates_test.py
obriente: I think we want to allow this behavior. If we raise an error then no-one can pass in a Hamiltonian with any coefficient in front of the identity (before this commit that raised an error). And if we just ignore the identity term then every time you did phase estimation with the circuit you'd wind up with a set of phases offset from their true variables (which I can't imagine someone would want).

Definitely the gate is the right choice, although I agree that the nomenclature is a bit odd - I guess technically this is a controlled global phase on the system, but that is just a Rz rotation, and I'm not sure if people would understand what the other gate meant when they read the output.

But, this again raises a good question - does OpenFermion have a set of standards for the gates that it outputs? If not, it might be a good idea to write this for future reference (and so that people that want to use the program don't have to dig into the code to find out what it might give them).

babbush: The code that you are editing was never designed as a particularly serious way of compiling circuits in OpenFermion. In fact, the core OpenFermion library is currently not really designed to compile things all the way to circuits - it is more of a set of tools for obtaining and manipulating representations of Hamiltonians and states of relevance to quantum simulations of electronic structure. The idea was always to compile circuits using plugins that link to specific quantum programming frameworks. For instance, there is Forest-OpenFermion to use the Rigetti stack and there is OpenFermion-ProjectQ to use the ProjectQ stack. Google is also working a similar framework for our chips and we'll also release an OpenFermion plugin designed to work with that. We also welcome the creation of new plugins like this.

So the answer to your question "does OpenFermion have a set of standards for the gates that it outputs?" is "no" because we never really intended OpenFermion core to a serious tool for compiling all the way to the hardware level. That said, I think it is fine to add the code you are adding.

jarrodmcc: Fair point about the action of a controlled global phase, whatever that might mean.

PR #198, opened 2018-01-31
Bug in Jellium test?

kevinsung:
I did not think about this in detail and just guessed that this is a bug by looking at the code. @idk3 or @babbush please verify. This would fix the problem discussed at #192 and make that PR mergeable.

### DIFF
--- a/src/openfermion/hamiltonians/_jellium_test.py
+++ b/src/openfermion/hamiltonians/_jellium_test.py
@@ -255,6 +255,7 @@ class JelliumTest(unittest.TestCase):
         for indices_a in grid.all_points_indices():
             for indices_b in grid.all_points_indices():
 
+                potential_coefficient=0.
                 paper_kinetic_coefficient = 0.
                 paper_potential_coefficient = 0.
 
--- a/src/openfermion/hamiltonians/_jellium_test.py
+++ b/src/openfermion/hamiltonians/_jellium_test.py
@@ -255,7 +255,7 @@ class JelliumTest(unittest.TestCase):
         for indices_a in grid.all_points_indices():
             for indices_b in grid.all_points_indices():
 
-                potential_coefficient=0.
+                potential_coefficient = 0.
                 paper_kinetic_coefficient = 0.
                 paper_potential_coefficient = 0.
 


idk3:
What does this actually change? My thought from my bug earlier (haven't had time to look in detail yet) was that it was caused by the test not expecting the higher number terms to be compressed out, and as a result computing the wrong number of qubits -> the spectra not having the same number of elements. But I'm not sure this would lead to the same change (though I'm not sure it should to begin with ;))

kevinsung:
It seems that in each iteration of the loop starting at line 256, we're computing the `potential_coefficient` and `paper_potential_coefficient` and checking that they're equal. However, as it is, the value of `potential_coefficient` is not being reset at the start of the loop, and if the if statement at 276 doesn't evaluate to true, then the value of `potential_coefficient` would not be changed from the previous iteration, and the assertion on line 290 would be using the value of `potential_coefficient` from the previous iteration.

idk3:
Mine actually is failing on `test_kinetic_integration`, so I don't think it's likely the same issue :(

Looking more closely at the PR though, shouldn't this addition be in an `else` under the `if` on line 276? Otherwise it looks like we continue to use the same potential coefficient as before (for the new p, q) - so I'm not sure this resolves the issue you're pointing out.

kevinsung:
I think your suggestion to put it in the `else` statement results in equivalent code. I like my way better because it makes the initialization of `potential_coefficient` more explicit.

The only test that gives #192 a problem is `test_coefficients`, and this change makes it pass.

idk3:
I think they might be functionally equivalent given the current code, but that isn't something we should rely on.

If we go through the code, line 258 assigns `potential_coefficient = 0.`. But `p` and `q` will change if the test is changed to the case with spin (this is only set by a flag in line 191), which will cause you to use an old value of `potential_coefficient` if you fail the `if` on 276, but had previously set it for an older `p, q` in that block.

So I think the two solutions are 1. put the zero assignment in an `else` under 276, or 2. eliminate the code (currently not being used) that would do things for the case with spin entirely. I don't think the current solution is totally safe.

babbush:
@idk3 I just saw your last comment now (after merging). I would have waited had I seen that. But I am not sure I completely understand what you are saying still...

PR #200, opened 2018-02-01
moved induced norm functionality to parent class

babbush:
@obriente please confirm these changes are acceptable. Note that we've add you as an author.

This PR will fail until we merge the SymbolicOperator PR in https://github.com/quantumlib/OpenFermion/pull/192


jarrodmcc:
Did you also want to change these to functions that act on the class instead of member functions?

babbush:
@jarrodmcc I decided I don't really care.

babbush:
@kevinsung can you approve PR again?

NOTICE
```diff
@@ -18,6 +18,7 @@ Tarini Hardikar (Dartmouth)
 Vojtech Havlicek (Oxford)
 Cupjin Huang (University of Michigan)
 Zhang Jiang (NASA)
+Thomas O'Brien (QuTech)
```
obriente: Thanks for the addition - my official affiliation is Leiden University though.

babbush:
@obriente I have updated your affiliation.

obriente:
LGTM - my only comment would be that the induced norm is not preserved across transformations.

This isn't a huge issue - there are always plenty of norms to choose, and the one that we might state as 'the' norm is the operator norm, and this is difficult to calculate. But we might want to note it somewhere because otherwise it might trip someone up at some point.

PR #201, opened 2018-02-01
Implemented long string initialization and tests for SymbolicOperator

max-radin:
This allows for operators to be initialized using long strings, e.g.

```FermionOperator('2.0 [0^ 1] + -1.0 [2^ 3]')```

The format of the string is fairly flexible. For example, the above operator could also be initialized with

```FermionOperator('2.0[0^ 1]-[2^ 3]')```

### CODE REVIEW

File: src/openfermion/ops/_symbolic_operator.py [hunk 1]
```diff
@@ -96,13 +96,64 @@ def __init__(self, term=None, coefficient=1.):
...
+
+        e.g. For FermionOperator:
+            '1.5 [2^ 3] + 1.4 [3^ 0]'
         """
-        raise NotImplementedError('Initializing a SymbolicOperator from a '
-                                  'long string is not yet supported.')
+
+        index = 0  # Index of current character in long_string
+        subtraction = False  # Whether this term is being subtracted
+        while index < len(long_string):
```
babbush: Is there a good reason to use while loops that increment an index as opposed to using for loops and "enumerate"? Unless there is a good reason, I think for loops and enumerate tend to lead to code that is bit easier to parse.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -96,13 +96,64 @@ def __init__(self, term=None, coefficient=1.):
...
+            else:
+                try:
+                    coef = float(
+                            long_string[coefficient_start:coefficient_end])
+                except ValueError:
+                    raise ValueError('Invalid coefficient {}.'.format(
+                            long_string[coefficient_start:coefficient_end]))
+            coef *= coefficient
+            if subtraction:
+                coef = -coef
```
babbush: I think it is slightly nicer to write "coef *= -1". I don't feel too strongly though.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -96,13 +96,64 @@ def __init__(self, term=None, coefficient=1.):
...
+            while long_string[index] is not ']':
+                index += 1
+            factor_end = index
+
+            # Determine the coefficient for this term
+            if coefficient_start == coefficient_end:
+                coef = 1.0
+            else:
+                try:
+                    coef = float(
```
babbush: In general, it is necessary that coefficients can be complex valued. It seems to me that this code will only work when coefficients are real-valued floats. I could be wrong about that though. At the very least, we'll need a test showing that this works for complex values.

src/openfermion/ops/_symbolic_operator_test.py
```diff
@@ -151,6 +151,17 @@ def test_init_str(self):
         self.assertIn(correct, fermion_op.terms)
         self.assertEqual(fermion_op.terms[correct], -1.0)
 
+    def test_init_long_str(self):
+        fermion_op = DummyOperator1(
+                '-2.0 [0^ 1] +\n\n -1.0[ 2^ 3 ] - []', -1.)
```
babbush: what is going on with the "\n\n"? Are you just making sure this will work if there is a new line?

src/openfermion/ops/_symbolic_operator_test.py
```diff
@@ -151,6 +151,17 @@ def test_init_str(self):
...
+        correct = \
```
babbush: I could be wrong, but I believe the way you are escaping to a new line here is a pep8 violation. Can you please run the pep8 linter to check?

src/openfermion/ops/_symbolic_operator_test.py
```diff
@@ -151,6 +151,17 @@ def test_init_str(self):
         self.assertIn(correct, fermion_op.terms)
         self.assertEqual(fermion_op.terms[correct], -1.0)
```
babbush: I would really like to see tests like 
my_fermion_operator - FermionOperator(str(my_fermion_operator)) = 0
my_qubit_operator - QubitOperator(str(qubit_operator)) = 0
where you initialize somewhat random things that look like fermion / qubit operators. Since the SymbolicOperator class is more general than just QubitOperator and FermionOperator, it might be a good idea to also have a test that does this for a dummy operator that is initialized with different "actions". But that is less important.

babbush:
Also, it might be good to document this functionality in the demo and/or in the release paper. But I won't require that to merge this PR.

kevinsung:
My original thought was to enforce that the input terms had to be separated by ` + `, and then we can use the string method `split` and a for loop to do this in a few lines of code. The flexibility of @max-radin 's implementation is nice though; it still might be possible to simplify it using the string method `split`.

ncrubin:
Building off @kevinsung suggestion, maybe it's possible to use the regular expression library `split` method.  It should allow for a variety of matching.  The regex code would probably be fairly complex given the variety of inputs `+, \s+\s, -, \s-\s, \s-[,`etc...

max-radin:
@babbush I think you're right that complex numbers might be a problem – I'll look into that. With regards to init being the inverse of str, there are tests for both [fermions](https://github.com/max-radin/OpenFermion/blob/b83e133a9a052ec21b9b2415acf2f84f241fca3b/src/openfermion/ops/_symbolic_operator_test.py#L162-L163) and [qubits](https://github.com/max-radin/OpenFermion/blob/b83e133a9a052ec21b9b2415acf2f84f241fca3b/src/openfermion/ops/_symbolic_operator_test.py#L677-L678).

In terms of simplifying the code, I don't think we can use `+` by itself to split terms because complex coefficients can also contain `+`. But a regex split might work well.

babbush:
@max-radin yes you did have those tests there. Don't know how I missed them! I think the main thing then is to make sure we can handle complex coefficients.

src/openfermion/ops/_symbolic_operator_test.py
max-radin: The pep8 linter says it's ok.

src/openfermion/ops/_symbolic_operator.py
max-radin: I've changed it now to loop over regex matches.

max-radin:
Ok, it now works with complex coefficients and should be good to go.

babbush:
This is great, thanks! Finally, you are welcome to add your name / affiliation to the alphabetical part of the authors section both in NOTICE and README, if you would like. If you don't want to do that, let me know and I will merge as is. Otherwise, I'll wait until you've done those things to merge.

Issue #154, opened 2018-01-09
Using symmetries to reduce qubit requirements


babbush:
This is a very broad topic, and one that is also long overdue for including in OpenFermion.

There are several interesting papers about reducing the number of qubits in second quantized Hamiltonians. The usual idea is that second quantization does not leverage good quantum numbers such as particle number and having definite spin, and thus wastes qubits. Here are a few papers that introduce valuable ideas along these lines. These papers outline ideas that would be straightforward to implement in OpenFermion:
https://arxiv.org/abs/1510.04048
https://arxiv.org/abs/1701.08213
https://arxiv.org/abs/1712.07067

conta877:
I'm already working on this. If anyone is interested in collaborating/working together let me know.

babbush:
The papers I mentioned all introduce different approaches. Which are you working on?

conta877:
I wasn't working on any one of the papers mentioned above but after reading I'm thinking of switching to 1712.07067

babbush:
Cool. I am at a conference right now and I believe one of the authors on that paper are also here. I will try to track them down and ask if they'd be willing to share any of their code.

I think it is important to note that while 1510.04048 seems easier to implement than the latter two papers, the technique has the unfortunate consequence of increasing the number of terms in the Hamiltonian exponentially. Thus, it is really only of academic interest at this point. The techniques introduced in 1701.08213 and 1712.07067 are efficient.

conta877:
did you manage to track them down? are they willing to share?

babbush:
Yes! I spoke to Stephanie and Mark a few hours ago actually. Mark said that at this point the code is entirely in mathematica and that they actually distributed it on arXiv with the source, which you should be able to find here: https://arxiv.org/e-print/1712.07067 (you will need to change the file to another name with the extension .tar.gz). Mark also indicated that he may be interested in helping work on this.

conta877:
Awesome! Will translate to python. (probably won't happen before monday - in case someone else wants to work on it ASAP)

conta877:
sorry - this is taking longer than I anticipated - if anyone wants to collaborate I keep my fork updated. for now introduced openfermion/simplifications not to break anything.

msteudtner:
Hi conta877, I was not aware that someone is already working on implementing my stuff to OpenFermion, but that's great. Can I help you?

conta877:
Hi Mark, yes that would be awesome! I'll send you an email

[...]

PR #202, opened 2018-02-08
the appetizer (1/4). binary_operator

conta877:
first pull request for #154

implementing arxiv 1712.07067
Decoders have binary variables w1, w2 that need to obey binary rules. 
SymbolicBinary class takes care of that. they make up one component of a decoder.


src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
+# symbolic binary class for decoder definitions (arxiv 1712.07067)
+
+import copy
+import numpy as np
```
babbush: okay, here it is, perhaps the most despised rule of OpenFermion: we have a strict "no importing as" rule. I know it's kind of silly in this case since np is almost universally recognized as numpy. But at this point we should stay consistent with other parts of the code.

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
+# symbolic binary class for decoder definitions (arxiv 1712.07067)
```
babbush: Please copy the Apache 2 license header that is in the rest of the OpenFermion files. This is a fine one-line description but it should go under the header and should be commented
`"""Like this."""`
rather than
`# like this.`

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+    pass
+
+
+class SymbolicBinary(object):
+    """
+        SymbolicBinaries are non-linear binary functions that are expressed analytically, they are
+        components of decoders that are used to rotate terms in qubit basis to fermionic basis. 
+        
+        A SymbolicBinary represents a sum of terms. Each term can consist of two types of binary variables:
+        (n,'W') where n stands for the qubit index, or (1,'1') for constant 1.
```
babbush: Can you add a little more context to the description about when this class would be used and for what. Also, I think it is a good idea to put the arXiv paper reference in the doc strings for the class so that people looking through the online documentation can find the reference.

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+         
+        The SymbolicBinary class overloads operations for manipulation of binary functions obeying binary rules.
```
babbush: Please run the pep8 linter. This line is over 100 characters. We're old school - keep to 80 char per line please.

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+            self.terms.append(tuple(self._parse_string(term)))
+
+        # Invalid input type
+        else:
+            raise ValueError('term specified incorrectly.')
+
+        self._check_terms()
+
+    def _check_terms(self):
+        """
```
babbush: General comment: you don't need white space at the top and bottom of the doc strings. As an example just write:
`"""Blah Blah Blah."""`

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+    actions = ('1', 'W')
+    action_strings = ('', 'W')
+    action_before_index = True
+    different_indices_commute = True
+
+    def __init__(self, term=None):
+        """
+        Initialize the SymbolicBinary based on term
+
+        Args:
```
babbush: Since your function raises exceptions you should have "Raises:" in the doc strings in a similar way to how you have "Args:"

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+
+    def _check_terms(self):
+        """
+        Ensures all terms obey binary rules
+        Updates terms in place.
+
+        """
+        sorted_input = []
+        for item in self.terms:
+            if len(item) == 0:
```
babbush: I think it would be nicer to write:
`if len(item):
    binary_sum_rule(sorted_input, tuple(sorted(set(item))))`
There is no need for the negative conditional followed by continue.

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+        for summand in term.split(' + '):
+            # for each sum term
+            parsed_summand = self._parse_string(summand)
+            self.terms.append(parsed_summand)
+
+        self._check_terms()
+
+    def _check_factor(self, term, factor):
+        """
+        makes sure all factor in a term makes sense, cleans it up if there is multiplication with 1.
```
babbush: empty line before Args

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+    def _parse_sequence(self, term):
+        """
+        Parse a term given as a sequence type (i.e., list, tuple, etc.).
+        e.g. [((1,'W'),(2,'W')),...]
+
+        Args:
+            term: (list) of tuples of tuples.
+
+        """
+        if not term:
```
babbush: Code is generally easier to read if try to avoid negatives in conditionals. For instance if you write "if term" and then what comes instead of "if not term"...

src/openfermion/ops/_binary_operator.py
```diff
@@ -0,0 +1,494 @@
...
+                if self.different_indices_commute:
+                    summand = sorted(summand, key=lambda real_factor: real_factor[0])
+                self.terms.append(summand)
+
+    @staticmethod
+    def _parse_string(term):
+        """
+        parse a string term 'w1 w2 w0'
+        Args:
+            term: (str) string representation of symbolicBinary term.
```
babbush: It is best if you write
`term (str): string representation of SymbolicBinary term `
I believe that is picked up in the optimal way by our sphinx documentation.

src/openfermion/ops/_qubit_operator.py
```diff
@@ -93,7 +93,7 @@ class QubitOperator(SymbolicOperator):
 
     def __imul__(self, multiplier):
         """
-        Override in-place mutiply of SymbolicOperator
```
babbush: Good work on these!

src/openfermion/ops/_binary_operator_test.py
```diff
@@ -0,0 +1,108 @@
+import unittest
+from _binary_operator import SymbolicBinary
```
babbush: please update `__init__.py` in this PR so that you can just write `from openfermion.ops import SymbolicBinary`.

src/openfermion/ops/_binary_operator_test.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+class SymbolicBinaryTest(unittest.TestCase):
```
babbush: These are nice tests but unfortunately you will need a few more. This is the downside to introducing a large class like this with all the magic methods. Coveralls will show you which lines are covered and which aren't. Apparently you are covering 156 relevant lines but missed 43.

babbush:
It is now outdated but I also had a comment that you should update `__init__.py` so you can import SymbolicBinary directly from openfermion.ops

conta877:
Oh wow thanks a lot for all the quick comments @babbush !  I just saw them, will work on them asap.

msteudtner:
You have my consent to use my contributions.

babbush:
@conta877 let me know when I should review again. Don't worry about the CLA bot at this point - it can be rather overzealous and I can override. I can see you both signed.

conta877:
@babbush will do! gonna be pushing/merging a bunch until then. sorry for the spam

conta877:
no more x's! @babbush I think (hope) we addressed all your comments.

src/openfermion/ops/__init__.py
```diff
@@ -21,3 +21,4 @@
 from ._interaction_operator import InteractionOperator
 from ._interaction_rdm import InteractionRDM
 from ._quadratic_hamiltonian import QuadraticHamiltonian
+from ._binary_operator import SymbolicBinary
```
babbush: we prefer to import things in alphabetical order, meaning this should be the first import. However, sometimes there is a problem with circular imports. If that is the case here, then it is fine.

conta877:
oh sorry I just forgot about that! will do

PR #204, opened 2018-02-12
code_operator class (2/4)

conta877:
_code_operator class. combines (encoder,decoder). the step before transforms!!
lets see those x's coveralls and travis!

second pull request for #154


src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
+import copy
+
+import numpy
+import scipy
+import scipy.sparse
+
+from openfermion.ops import SymbolicBinary
+
+
+def _shift_decoder(decode, shift_constant):
```
babbush: Python does not really have a concept of private functions. I suppose this underscore is meant to indicate that it is an internal helper function. We've allowed that so far for methods on classes but not for external functions. I feel that this should either be a method on the class or you should just make it a public function (get rid of the underscore).

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+    Args:
+        decode (list of SymbolicBinary): a decoder
+        shift_constant (int): the qubit index that corresponds to the offset.
+
+    Returns (list of SymbolicBinary):  shifted decoder
+    """
+    decode_shifted = []
+    for entry in decode:
+        tmp_entry = copy.deepcopy(entry)
+        tmp_entry._shift(shift_constant)
```
babbush: The fact that you are calling ._shift() externally on the class SymbolicBinary (as opposed to calling some public method which is using ._shift() as part of the internal implementation) makes me think that this ._shift() method should also be a public method, i.e. .shift().

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+        decode_shifted.append(tmp_entry)
+    return numpy.array(decode_shifted)
```
babbush: This doesn't match the doc strings, which say a list is returned. I am surprised this works - I did not know you could have a numpy array populated with a custom class as entries.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+
+
+def _double_decoding(dec1, dec2):
```
babbush: Again, let's talk about whether this should be a "private" function. Also, I would prefer if you had slightly more complete names like "decoder_1, decoder_2" for the arguments of functions.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
+import copy
+
+import numpy
+import scipy
+import scipy.sparse
+
+from openfermion.ops import SymbolicBinary
+
```
babbush: 1) Please add the Apache 2 header.
2) Perhaps add a one line doc string `"""in this style"""` which references the arXiv paper so know what this module does and is about.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+            tmp_term = SymbolicBinary('1')
+            for factor in summand:
+                if factor[1] == 'W':
+                    tmp_term *= dec2[factor[0]]
+            tmp_sum = tmp_term + tmp_sum
+        doubled_decoder += [tmp_sum]
+    return numpy.array(doubled_decoder)
+
+
+def linearize_decoder(mtx):
```
babbush: Please use more descriptive name than "mtx".

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+            size, qubits indexed mismatch
+        """
+
+        if not isinstance(encoding, (numpy.ndarray, list, tuple)):
+            raise TypeError('encoding must be a list, array or tuple .')
+
+        if not isinstance(decoding, (numpy.ndarray, list, tuple)):
+            raise TypeError('decoding must be a list, array or tuple .')
+
+        self.enc = scipy.sparse.csc_matrix(encoding)
```
babbush: I really think these attributes should have a more complete name like "encoder/encoding" instead of "enc". There is something slightly arbitrary about such abbreviations which makes it a little less intuitive to anticipate how things are named without digging through the docs.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if (self.orbitals != len(decoding)):
+            raise BinaryCodeError(
+                'size mismatch, decoder and encoder should have the same'
+                ' first dimension')
+
+        decoder_qubits = set()
+        self.dec = []
+
+        for symbolic_binary in numpy.array(decoding):
```
babbush: Why do you case the decoding into an array in order to loop through it? Is that necessary?

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+    def __init__(self, encoding, decoding):
+        """ Initialization of a binary code.
+
+        Args: 
+            encoding (array or list): nested lists or binary 2D-array
+            decoding (array or list): list of SymbolicBinary, list-like or str
+
+        Raises: 
+            TypeError: non-list, array like encoding or decoding, unsuitable
+            symbolicBinary generators,
```
babbush: Please indent the second line of these argument/raises descriptions. See other code in library for example (e.g. https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/ops/_symbolic_operator.py)

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+        dec_str = dec_str.rstrip(' + ')
+        res.append(SymbolicBinary(dec_str))
+    return numpy.array(res)
+
+
+class BinaryCodeError(Exception):
+    pass
+
+
+class BinaryCode(object):
```
babbush: This is fairly comprehensive docs. Even though I think I forgot to request this on your SymbolicBinary class, we usually have "attributes" listed with descriptions in the doc strings under the class definition. See (https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/ops/_symbolic_operator.py) for an example.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+
+        if len(decoder_qubits) != self.qubits:
+            raise BinaryCodeError(
+                'decoder and encoder provided has different number of qubits')
+
+        if max(decoder_qubits) + 1 > self.qubits:
+            raise BinaryCodeError('decoder is not indexing some qubits. Qubits'
+                                  'indexed are: {}'.format(decoder_qubits))
+
+        self.dec = numpy.array(self.dec)
```
babbush: Again, I think a longer name than ".dec" would be better.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def __iadd__(self, appendix):
+        """ In-place appending a binary code with +=.
+        
+        Args:
+            appendix (BinaryCode): The code to append to the present one. 
+
+        Returns (BinaryCode): a global binary code with size n1+n2, N1+N2
```
babbush: I think if you are going to refer to size of encoded and unencoded spaces as "n" and "N" as opposed to something like "encoded_dimension" and "unencoded_dimension" then I think you need to keep reminding people what "n" and "N" are.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+        Raises:
+            TypeError: appendix must be a BinaryCode
+        """
+        if not isinstance(appendix, BinaryCode):
+            raise TypeError('argument must be a BinaryCode.')
+
+        self.dec = numpy.append(self.dec,
+                                _shift_decoder(appendix.dec, self.qubits))
+        self.enc = scipy.sparse.bmat([[self.enc, None], [None, appendix.enc]])
+        self.qubits, self.orbitals = numpy.shape(self.enc)
```
babbush: In other parts of the code we call these things "n_orbitals" and "n_qubits". You might also consider saying "modes" instead of "orbitals" because if this is used with other parts of OpenFermion there might be an ambiguity versus spatial orbitals and spin-orbitals (usually there are twice as many spin-orbitals than orbitals) but we want this code to be used for models both with and without spin degrees of freedom.

src/openfermion/ops/_code_operator.py
```diff
@@ -0,0 +1,298 @@
...
+            self.qubits *= factor
+            return self
+
+    def __mul__(self, factor):
+        """ Concatenation of two codes or appendage the same code factor times
+        in case of integer factor.
+        
+        Args:
+            factor (int or BinaryCode): the BinaryCode to concatenate. In case
+             of int, it will append the code to itself factor times.
```
babbush: Indent. Is this not caught by pep8 linter?

src/openfermion/ops/_code_operator_test.py
```diff
@@ -0,0 +1,64 @@
+import unittest
+
+from openfermion.ops import SymbolicBinary
+from openfermion.ops._code_operator import BinaryCodeError, \
```
babbush: I think pep8 would complain about this, right? I think it is preferred to use parentheses in imports like how it is done here for ._fermion_operator: https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/ops/__init__.py

src/openfermion/ops/_code_operator_test.py
```diff
@@ -0,0 +1,64 @@
...
+    BinaryCode, linearize_decoder
+
+
+class BinaryCodeTest(unittest.TestCase):
```
babbush: I am virtually certain pep8 linter will complain about the lack of spaces here. Please remember to run pep8 linter: https://pypi.python.org/pypi/pep8

src/openfermion/ops/_code_operator_test.py
```diff
@@ -0,0 +1,64 @@
+import unittest
```
babbush: Apache 2 header.

msteudtner:
All authors are ok with their commits being contributed to this project.

msteudtner:
@babbush 
I officially don't get the thing with the bot - what does it want me to do? 
I feel quite silly writing a comment.

babbush:
@msteudtner the bot is really quite annoying. Just ignore it. It always gets confused when two people made commits on the pull request but only one person pushed those commits. It would be okay if just Isil pushed her own commits or she pushed her own and then you pushed your own but whenever a third party pushes for you it gets confused. However, I know you have both signed the CLA at this point and that is all that matters. You don't need to dignify the bot with a response in the future (on this or other PRs).

PR #208, opened 2018-02-14
transform function 3/4

conta877:
3rd pull request for #154


src/openfermion/transforms/_binary_code_transform.py
```diff
@@ -0,0 +1,187 @@
...
+    and outputs the Pauli-string sum that corresponds to the
+    decomposed multi-qubit operator.
+
+    Args:
+        term (tuple): product of binary variables, i.e.: 'w0 w2 w3'
+
+    Returns (QubitOperator): superposition of Pauli-strings
+
+    Raises:
+        ValueError: if the action is not 'W'
```
babbush: Why "W" where does this convention come from exactly? I am concerned that this design is rather opaque.

babbush:
Note that I am hoping that PR 4/4 can also come with a small example ipython notebook, or perhaps a new section in our main demo, demonstrating how this code works in a pedagogical fashion. That would be really great and I am sure such examples would increase adoption.

src/openfermion/transforms/_binary_code_transform.py
msteudtner: I'm not sure that is a very satisfying answer for you, but 'w' is just labeling the binary variables. It stems from the SymbolicBinary class.  At some point writing the paper I had to decide for symbols for the binary vectors of the Fermion modes and qubit configuration. I just called them \nu and \omega. We don't need  \nu here, but \omega  survived as 'w'.

babbush: So I am starting to question whether the data structure for SymbolicBinary is implemented in a reasonable fashion. What other value can the action take, just the empty string. In this case, it seems to me that the whole data structure is overcomplicated. Instead of storing  (1, 'W'), (2, 'W') shouldn't you just store (1,2)? To represent constants you can use "none", or something like that.

conta877: I actually experimented with different ways of defining the SymbolicBinary class - even tried the (1,2,..) version which would indeed simplify things a lot for the operations - couple of problems I had with that was the constant definition and user input. 
for constant indeed a string is a good idea, did not think of that. As for user definitions would we not allow for str inputs? would we stick to 'W' in str inputs - if not what if the user gives 1+q1+w2+z3 is that ok? or do we assume that anyone thats using the symbolicBinary class knows what they are doing?

msteudtner: It's definitely more streamlined this way, you are right. We basically made it like it is now just because appears not to be that different from the data structures of FermionOperator and QubitOperator. But it's your call:  would you like us to change (1, 'W'), (2, 'W') into (1,2) and so on?

conta877: I guess also when i started writing this I was wondering if it could adopt from SymbolicOperator class as well.

babbush: Yes, I think this is a good change to make. I would actually suggest "None", for a constant, but you could put a string. It's very easy to write "if action is None:" and it is probably less arbitrary than any string you could choose. I can tell you were thinking of inheriting from SybolicOperator, but since you didn't, I think the change makes sense.

Regarding string initialization, I think that is still fine. You can probably allow them to use any char in fact. So if they wrote "a1 a2 + a3 a4" or "W1 W3 + W3 W4" it would do the same thing. Another option is to represent the term a1 a2 with (1 2) in the string input also. So you could write (1 2) + (3 4). I personally find the second option cleaner but that is up to you.

Another thing to think about is whether you want to enforce numerical order. For instance, if you want to disallow people from having a term (3,1) as opposed to (1,3). If you do this, you can certainly perform certain numerical operations faster since you know the list is sorted. But unless numerical performance is a bottleneck it is not necessary to specialize the magic methods for this now (but it might still he a good idea to put systems in place that prevent having both (1 3) and (3 1).

Sorry I did not catch this before!

msteudtner: Ok, we'll make the change. We already take care of several mishaps during the input, like a messed up order or additions with constants other than one, and so on.

babbush: Awesome, thanks!

conta877: I'm ok with it all but None. its a dangerous name to use for a variable. certain data structures are built to ignore None's by default - and defining 1 as None is confusing to me.

babbush: That's fine. If you feel a string would be better that seems reasonable.

babbush:
Let me know when you are ready for review, thanks for making these changes!

conta877:
we are ready for review! (I'll improve the coverage tonight)

src/openfermion/transforms/_binary_code_transform.py
```diff
@@ -0,0 +1,178 @@
...
+    and outputs the Pauli-string sum that corresponds to the
+    decomposed multi-qubit operator.
+
+    Args:
+        term (tuple): product of binary variables, i.e.: 'w0 w2 w3'
+
+    Returns (QubitOperator): superposition of Pauli-strings
+
+    Raises:
+        ValueError: if the _action is not 'W'
```
babbush: I update docs. I think this was left over from prior implementation.

babbush:
Actually, test coverage looks okay currently! Definitely add more tests if you like though.

src/openfermion/transforms/_binary_code_transform.py
conta877: hmm I remember fixing that - meaning I may have messed something up with my merging. I should check to see whats happening before you approve.

conta877: oh nvm - I totally missed that one and fixed the same error within the raise.

conta877:
for the notebook, do you prefer a section in main demo or separate?

babbush:
Either would be great! Maybe easiest to add to main demo.

babbush:
Should I go ahead and merge?

conta877:
yup! there will be a 4/4 with transform functions and addition to the main demo

PR #210, opened 2018-02-15
Add and edit functions related to numerical preparation of Slater determinants

kevinsung:
- Rename the function originally called `jw_slater_determinant` to `computational_basis_state`, which more accurately describes what it does.
- Make it so that `jw_slater_determinant` now takes as input a numpy array storing the coefficients of orbitals defining a Slater determinant and returns the Slater determinant numerically as a sparse array.
- Fix the documentation of the particle-hole transformation in the Slater determinant preparation circuits; the original definition was wrong (related to a minor mistake in arXiv:1711.05395 which will be fixed).


src/openfermion/utils/_sparse_tools.py
```diff
@@ -207,28 +208,28 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
     return sparse_operator
 
 
-def jw_slater_determinant(occupied_orbitals, n_orbitals):
-    """Function to produce a Slater determinant in JW representation.
+def computational_basis_state(one_qubits, n_qubits):
```
babbush: I don't know how I ever let this name slip in the first place. Obviously it is not an arbitrary Slater determinant but only one that is expressed as a configuration in the chosen basis. So we must change the name. However, I am not so sure I like the idea of changing the argument "occupied_orbitals" to "one_qubits". First of all, it sounds too much like the company by that name is multiplying ;-)

Second of all, very often one just wants to provide a list of which orbitals are occupied. For simplicity, most of this sparse code assumes the occupation number basis (associated with the Jordan-Wigner encoding). What if we kept the argument as "occupied_orbitals" and made the function name "jw_computational_basis" or something?

src/openfermion/utils/_sparse_tools.py
```diff
@@ -431,6 +432,46 @@ def jw_get_gaussian_state(quadratic_hamiltonian, occupied_orbitals=None):
     return energy, state
 
 
+def jw_slater_determinant(slater_determinant_matrix):
```
babbush: Nice addition.

src/openfermion/utils/_sparse_tools.py
kevinsung: Okay, I mostly agree with you. What do you think of renaming the function `jw_fock_state` instead? I think `jw_computational_basis_state` is fine too though. I'm wondering which one is clearer, considering that we may later add the analogous functions for the Bravyi-Kitaev transform.

babbush: What about "jw_configuration_state"

kevinsung: Sure, let's do that.

PR #212, opened 2018-02-18
Active space

### LINKED ISSUE #207: freezing orbitals in FermionOperator
I have a FermionOperator, and I'd like to make the assumption that some of the orbitals are always empty and others are always full. (As I understand it, this is similar to the active space functionality in `MolecularData`.) My question is, what's the best way to do this?

What I'm thinking of doing is implementing two methods for `FermionOperator`, `assume_occupied(index)` and `assume_unoccupied(index)`, that will remove the ladder operators that act on a given orbital and multiply the coefficients for the corresponding terms by 0, +1, or -1 as appropriate. I'd also need to implement a `prune()` method for `SymbolicOperator` that determines which indices don't appear in any terms and then renumbers the indices accordingly.

The idea is that 

```FermionOperator('1.0 [4^ 5 6]').assume_occupied(5).prune()``` 

would yield `-1.0 [0^ 1]`.

Does this seem like a reasonable way to go?

### PR DESCRIPTION
**Pull request by max-radin**
Allows for the active space of a FermionOperator to be reduced by freezing specified orbitals.

Closes #207.

### CODE REVIEW

**Comment by babbush** [comment]
This is a neat idea. Are there specific use cases you have in mind for this? Also note that there seems to a problem caught by Travis. However, I think Travis might have errored here and that your PR is fine. You should restart travis.

**Comment by max-radin** [comment]
At the moment I'm just using this to reduce my Hamiltonian down to something that's small enough to do test calculations on. In the general case I'd imagine one might want to freeze low-energy states that aren't important for bonding.

Also I should say that `MolecularData` already has a method for setting the active space when generating a molecular Hamiltonian. The purpose of the new code is to allow you to freeze orbitals in the general case, such as when building a plane-wave Hamiltonian.

File: src/openfermion/ops/_fermion_operator.py [hunk 1]
```diff
@@ -205,3 +205,75 @@ def is_molecular_term(self):
...
+
+        Removes all operators acting on the specified orbitals, and flips the
+        sign of each term as appropriate. Terms whose expectation value
+        between states with the specified occupancy is zero are removed.
+        Additionally, the indices are renumbered such that the frozen orbitals
+        are eliminated.
+
+        Note that the sign change accounts for both the reordering of operators
+        needed to group the operators acting on the frozen orbitals, as well
+        as the impact of pruning frozen orbitals on the Jordan-Wigner strings
```
jarrodmcc: As a matter of phrasing, I would say these things result from fermion anti-commutation rather than the JW representation of that.

src/openfermion/ops/_fermion_operator.py
```diff
@@ -205,3 +205,75 @@ def is_molecular_term(self):
...
+                if n_swaps % 2 is 1:
+                    new_coef *= -1
+                if new_coef is not 0 and current_occupancy is item[1]:
+                    if tuple(new_term) in new_terms:
+                        new_terms[tuple(new_term)] += new_coef
+                    else:
+                        new_terms[tuple(new_term)] = new_coef
+            self.terms = new_terms
+
+        # Determine sign flips arising from Jordan-Wigner strings
```
jarrodmcc: I think you mean fermion anti-commutation

src/openfermion/ops/_fermion_operator.py
```diff
@@ -205,3 +205,75 @@ def is_molecular_term(self):
...
+        for item in frozen:
+            new_terms = {}
+            for term in self.terms:
+                new_term = []
+                new_coef = self.terms[term]
+                current_occupancy = item[1]
+                n_ops = 0  # Number of operations on index that have been moved
+                n_swaps = 0  # Number of swaps that have been done
+
+                for op in enumerate(reversed(term)):
```
jarrodmcc: This seems like it borrows a lot from typical normal ordering, which we already have routines for in this file.  Do you think it's possible to use that routine or modify that routine in a way that you can use it here as a core element in your function to minimize code duplication?

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -563,3 +563,34 @@ def induced_norm(self, order=1):
         for coefficient in self.terms.values():
             norm += abs(coefficient) ** order
         return norm ** (1. / order)
+
+    def prune(self):
```
jarrodmcc: I think the name is a little generic for different types of pruning one could do to a fermionic operator, e.g. removing terms with small coefficients or otherwise.  Maybe something like remove_unused_indices or prune_unused_indices could work or something else?

src/openfermion/ops/_fermion_operator.py
```diff
@@ -205,3 +205,75 @@ def is_molecular_term(self):
...
+                    else:
+                        new_term.insert(0, op[1])
+                if n_swaps % 2 is 1:
+                    new_coef *= -1
+                if new_coef is not 0 and current_occupancy is item[1]:
+                    if tuple(new_term) in new_terms:
+                        new_terms[tuple(new_term)] += new_coef
+                    else:
+                        new_terms[tuple(new_term)] = new_coef
+            self.terms = new_terms
```
jarrodmcc: If you are going to create a new copy internally anyway, I think the the default behavior of the function should be to return that new copy and leave the acted upon fermion operator unchanged.  This gives the user the option of overwriting their previous operator, but they don;t have to manually make a deep copy of the whole object if they want to keep the old one around.

src/openfermion/ops/_fermion_operator.py
```diff
@@ -205,3 +205,75 @@ def is_molecular_term(self):
             if not (particles == spin == 0):
                 return False
         return True
+
+    def freeze_orbitals(self, occupied, unoccupied=[], prune=True):
```
jarrodmcc: Avoid empty list initializers.  They lead to all kinds of problems due to a peculiarity in how they are handled in python.  The preferred method is to initialize to None, then in the body of the function test if unoccupied is None and if so set it to [].  This prevents a lot of accidental bugs from happening.

max-radin:
@jarrodmcc sure thing, I'll work on implementing those changes!

src/openfermion/ops/_fermion_operator.py
max-radin: The default value is now None.

src/openfermion/ops/_fermion_operator.py
max-radin: freeze_orbitals and prune_unused_indices are now functions that return a new copy.

src/openfermion/ops/_symbolic_operator.py
max-radin: I agree. I renamed it prune_unused_indices.

src/openfermion/ops/_fermion_operator.py
max-radin: It's not obvious to me how to use normal_ordered to simplify this part of the code. I want to bring all of the ladder operators corresponding to the given orbital to the right side of the term, whereas normal_ordered will put them somewhere in the middle.

src/openfermion/ops/_fermion_operator.py
max-radin: I've reworded the comments to (hopefully) be more clear and avoid mentioning JW strings.

max-radin:
Ok, I think the issues with this PR have been fixed (assuming the Travis CI build passes). I feel like I should write down somewhere the details of how this feature works. Would the release paper be an appropriate place to do that?

babbush:
Yeah, the release paper would be a great place for this. Also feel free to add a small section the demo if you like.

jarrodmcc:
Thanks for the great contribution Max, I think it's looks good to go!

jarrodmcc:
Will merge when tests pass after merging with master branch

PR #214, opened 2018-02-20
fixing hermiticity error in dual basis external potential

babbush:


babbush:
This is wrong, ignore.

Issue #216, opened 2018-02-23
orbital reordering in FermionOperator class


conta877:
We would like to add a function to FermionOperator class that reorders the orbitals as
0,1,2,3,4,5,... -> 0,2,4,...,1,3,5... 
(from updownupdown... to upupupupdowndowndown)
two ways we can do this:

1) just writing a function that takes in a FermionOperator and spits out a reordered FermionOperator. which is what I did to begin with but @msteudtner raised the following concern:
the FermionOperator does not know its order inherently and a user may end up adding/multiplying two differently ordered FermionOperators :scream_cat: 

2) we can make it a class method and let the FermionOperator know its order. This way it can check if its being added/multiplied with a another FermionOperator with the same order and if not - raises an error. 
This seems to be the _right_ way of doing it to me but it would require overloading basic class methods. (unless there is some magic trick we can pull off :fireworks:)

jarrodmcc:
To clarify, when you say re-order, you are just talking about relabeling
rather than actually permuting the operators to something other than normal
ordering right?

Can you talk a bit more about the application you had in mind? As the
orbital labels usually only matter relative to coefficients of tensors
defining some action on the space, like the Hamiltonian coefficients.
Those objects are perhaps easier to permute in a lot of cases.  Just trying
to help brainstorm potential paths towards what you want.  The even-odd
choices for expanding orbitals into spin orbitals can be re-defined in a
few places as well.

msteudtner:
Hi @jarrodmcc 
We want to relabel the indices.

Maybe to highlight a bit the relevance of this problem:
Let's say we have  a molecule with spatial orbitals labeled  ( A, B, C, ... ). The multiplicity of the spin will give you a spin-up version of A, let's call it A0, and a spin-down version of A, called A1. Similar for all other orbitals.
Now in OpenFermion, we consider (when we talk about a Fermion basis state) that it is generated by creation operators like in the following example:

a^{+}_0  a^{+}_1  a^{+}_2   a^{+}_3   ...   |vac> .

Here the indices  (0, 1, 2, 3, ... ) are short for the labels (A0, A1, B0, B1 , ... ).
That's a good choice: in Jordan-Wigner transform, Hubbard-like terms will probably be local i. e. ~ Z2 Z3 .
However:  hopping terms from e.g.  A0 to C0  will run over some orbitals with opposite spin, so here yield operators  ~ X0 Z1 Z2 Z3 X4.
[...]

PR #217, opened 2018-02-23
Add spin indexing arguments to Hamiltonian model generation functions

kevinsung:
This is the same scheme used in `hamiltonians/_special_operators.py`.


kevinsung:
No, `up_index` is a function (associated with the even-odd indexing scheme) and `up_map` is the name of the keyword argument. We might want to change the names (I simply followed what @ncrubin  did with the fermion spin operators). Prior to this change, you could only generate these Hamiltonians using the even-odd scheme, but there are reasons you'd want to do, e.g., up-first, down-second. My particular motivation is that eventually I want to modify the Givens rotation code to not mix orbitals of different spin if that's not necessary; I was going to wait until I had a concrete use for this but I figured this might help address #216 .

babbush:
got it, thanks

PR #218, opened 2018-02-25
two methods of half-up

### LINKED ISSUE #216: orbital reordering in FermionOperator class

conta877:
this pull request is more of a question 

there are two implementations of half-up (#216),
1) a stand alone function in fermion_operator.py;  **half_up_order**

2) modified get_molecular_hamiltonian function of MolecularData, which kinda spread from there -> InteractionRDM expectation needs to know the ordering of the InteractionOperator, that required the InteractionOperator to know its order. For now I only made expectation function raise NotImplementedError if InteractionOperator has half_up order, just because I am not sure which method we are gonna choose.

I am aware that 1) will cause the same problem with InteractionRDM but half_up_order function reordering feels less internal thus maybe its ok to not handle that. 

Question is, which one do we want, do we want to implement this in another way? 

Ryan edit: Closes #216


kevinsung:
With method (2), do InteractionOperator and InteractionRDM really need to "know its order"? I think we can just assume that the user will only evaluate expectation values for operators that use the same ordering; otherwise it's her own fault. Therefore I think method (2) should only require modifications to MolecularData.

conta877:
@kevinsung no they don't need to know 2) can have the same attitude towards user errors. With that assumption, would you rather 1 or 2?

babbush:
Agree with @kevinsung .

kevinsung:
I think 2 is more user-friendly so I say we go with that.

kevinsung:
I'll add that I don't think `half_up` is a descriptive enough name. I suggest that we actually call the argument `spin_indexing` and set it to 0 or 1, and clearly document that, for instance, 0 indicates the even-odd scheme and that 1 indicates the up-then-down scheme.

kevinsung:
Actually, instead of `spin-indexing` be 0 or 1, perhaps let's make it be a string, either 'even-odd' or 'up-then-down'.

conta877:
what do you think of `indexing` = "orbital" or "spin"

babbush:
If we're going to go with a string I like Kevin's suggestion because it leaves no ambiguity as to what it does.

conta877:
ignore my last push - I failed at properly merging with my master.

babbush:
Let me know when you're ready for review.

conta877:
ready!

PR #221, opened 2018-02-26
started features list

babbush:
To be merged once completed. Help welcome.


kevinsung:
I pushed a commit fixing the reference for the Verstraete-Cirac transform.

The implementation of Bravyi-Kitaev is from arXiv:1701.07072, rather than arXiv:1208.5986, right?

babbush:
Yeah I think you're right about the BK implementation.

jarrodmcc:
For the citations I think it's fair to cite both for BK, as I've never met anyone that actually coded the transformation from the original paper.  I certainly used the work in the Seeley paper when I coded mine.

kevinsung:
I made a few additions to the utils and ops sections. Also, I changed the reference for the BK transform. @jarrodmcc as discussed at #232, the Seeley paper gives a different definition for non-powers of two from our implementation, so I think it would be inaccurate to cite it here.

jarrodmcc:
Fair enough, I hadn't realized we completely scrubbed the Seeley
implementation for the fenwick tree version, but I see that now.

babbush:
I'm tired of this being open and don't want to continue it right now. So I am going to close.

PR #223, opened 2018-02-27
Updated _bksf.py

TariniHardikar:
Removed lines pertaining to the two-unique-indices case. That case is incorrect -- terms with two unique indices should not contribute to the edge operator. Discussed and done with @kanavsetia.


babbush:
Sounds like an important change but it seems that tests are failing.

kanavsetia:
Hey Ryan, 
We recently figured that you can run hydrogen using BKSF transformation requires just 4 qubits (you can do it using 6 qubits too, as was the case earlier). And for one of the test cases, I explicitly initialized the size of one of the vectors as (2^6,1), which needs to be changed to (2^(n_qubits),1). With this change all the test pass on our end, and I think that is the only problem pointed out by Travis too. 
Tarini is making the change and rechecking the code one more time before committing.

PR #227, opened 2018-03-01
Generalize is_hermitian and move it to operator_utils

kevinsung:
- Allow `is_hermitian` to be applied to FermionOperators, QubitOperators, and Numpy arrays, and move it from `utils/_sparse_tools.py` to `utils/_operator_utils.py`
- Allow `hermitian_conjugated` to be applied to sparse matrices and Numpy arrays

### DIFF
--- a/src/openfermion/utils/_operator_utils.py
+++ b/src/openfermion/utils/_operator_utils.py
@@ -12,19 +12,20 @@
 
 """This module provides generic tools for classes in ops/"""
 from __future__ import absolute_import
+from builtins import map, max, zip
 
 import marshal
 import numpy
 import os
 import time
 
-from openfermion.config import *
+from openfermion.config import DATA_DIRECTORY, EQ_TOLERANCE
 from openfermion.hamiltonians._jellium import (grid_indices,
                                                momentum_vector,
                                                orbital_id,
                                                position_vector)
-from openfermion.ops import *
-from future.builtins.iterators import map, zip
+from openfermion.ops import (FermionOperator, InteractionOperator,
+                             InteractionRDM, PolynomialTensor, QubitOperator)
 from scipy.sparse import spmatrix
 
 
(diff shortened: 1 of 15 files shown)


babbush:
Whoops I guess that a test isn't passing.

src/openfermion/utils/_operator_utils_test.py
```diff
@@ -193,6 +194,78 @@ def test_hermitian_conjugated_semihermitian(self):
...
         self.assertTrue(op_hc.isclose(hermitian_conjugated(op)))
 
+    def test_exceptions(self):
+        with self.assertRaises(TypeError):
+            _ = is_hermitian('a')
+
+
+class IsHermitianTest(unittest.TestCase):
+    
+    def test_fermion_operator(self):
```
idk3: These should be separate tests!

src/openfermion/utils/_operator_utils.py
```diff
@@ -12,19 +12,21 @@
...
 
-from openfermion.config import *
+from openfermion.config import DATA_DIRECTORY, EQ_TOLERANCE
 from openfermion.hamiltonians._jellium import (grid_indices,
                                                momentum_vector,
                                                orbital_id,
                                                position_vector)
-from openfermion.ops import *
-from future.builtins.iterators import map, zip
+from openfermion.ops import (FermionOperator, InteractionOperator,
```
idk3: Thanks for removing these `import *`s!

src/openfermion/utils/_operator_utils_test.py
```diff
@@ -193,6 +194,78 @@ def test_hermitian_conjugated_semihermitian(self):
...
+        op = numpy.arange(16, dtype=complex).reshape((4, 4))
+        op += 1.j * op
+        op += op.T.conj()
+        self.assertTrue(is_hermitian(op))
+        op = csc_matrix(op)
+        self.assertTrue(is_hermitian(op))
+
+    def test_exceptions(self):
+        with self.assertRaises(TypeError):
+            _ = is_hermitian('a')
```
idk3: You don't actually need to assign for this to raise an error, but it's OK as is

PR #228, opened 2018-03-01
Fix sphinx warnings

kevinsung:
Mostly escaping some special characters such as `|` or `*` which are used by Sphinx for formatting or other things. Ideally these will be made pretty using TeX formatting eventually.


kevinsung:
Sorry there are some really old commits here, I probably should have created a new branch for this.

babbush:
I don't think it matters. I'm just glad somebody is trying to fix the docs!

PR #231, opened 2018-03-02
Move special operators module to utils

kevinsung:
- Moved `_special_operators.py` from `hamiltonians` to `utils`.

@ncrubin I think this was your original plan; I fixed the circular dependency issue by shuffling around import statements in `__init__.py`.


kevinsung:
No idea, is there a way we can restart the build?

babbush:
Can you make another commit of some sort and push?

babbush:
That seems to have fixed Travis.

PR #232, opened 2018-03-02
4/4! - transform functions.

conta877:
final pull request for #154 
transforms demo,
predefined transforms & their tests

I really wanted to name the file code_transformers... :robot:


src/openfermion/hamiltonians/_molecular_data.py
```diff
@@ -820,16 +820,16 @@ def get_active_space_integrals(self,
                                              active_indices,
                                              active_indices)])
 
-    def get_molecular_hamiltonian(self,occupied_indices=None,
-                                  active_indices=None):
+    def get_molecular_hamiltonian(self, occupied_indices=None,
+                                      active_indices=None):
```
babbush: you had it right the first time actually! It should align under self.

conta877: i'm gonna get it all right next time!!

conta877:
Any suggestions as to how to satisfy travis about the demo test error?

kevinsung:
I think you need to add the code block that is at the top of the general demo (the one that is there for testing purposes). This is admittedly ugly but we haven't figured out a better way yet.

conta877:
ready for review :tada:

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,238 @@
...
+    reps = int(numpy.ceil(numpy.log2(modes)))
+    mtx = numpy.array([[1, 0], [1, 1]])
+    for repetition in numpy.arange(1, reps + 1):
+        mtx = numpy.kron(numpy.eye(2, dtype=int), mtx)
+        mtx[2 ** (repetition + 1) - 1, 2 ** repetition - 1] = 1
+    return mtx[0:modes, 0:modes]
+
+
+def _encoder_checksum(modes):
+    """ Outputs the encoder matrix for checksum codes.
```
babbush: Can you add a couple lines describing the checksum code?

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,238 @@
...
+    for index in numpy.arange(digits):
+        binary_expression *= SymbolicBinary(
+            'w' + str(index) + ' + 1 + ' + address[index])
+
+    return list(map(int, list(address))), binary_expression
+
+
+def weight_one_binary_addressing_code(exponent):
+    """ Weight-1 binary addressing code.
+
```
babbush: Again, it would be good to have some more explanation of what these codes are and why people might be interested in them.

src/openfermion/transforms/_code_transform_functions_test.py
```diff
@@ -0,0 +1,93 @@
...
+        hamiltonian, gs_energy = lih_hamiltonian()
+        code = checksum_code(4, 0)
+        qubit_hamiltonian = binary_code_transform(hamiltonian, code)
+        self.assertAlmostEqual(gs_energy,
+                               eigenspectrum(qubit_hamiltonian)[0])
+
+    def test_jordan_wigner(self):
+        hamiltonian, gs_energy = lih_hamiltonian()
+        code = jordan_wigner_code(4)
+        qubit_hamiltonian = binary_code_transform(hamiltonian, code)
```
kevinsung: Is this supposed to do the same thing as the already-implemented `jordan_wigner`? If so, can we add a test that it does? Same for Bravyi-Kitaev.

idk3: @conta877 @kevinsung @babbush I wonder if it would make sense to have `jordan_wigner` alias to this (eliminating duplicate code) - does being more general mean that this code is slower / is there any other reason not to?

babbush: I agree with @kevinsung 's suggestion but am very opposed to @idk3 's suggestion. The implementation of Jordan-Wigner using the CodeOperator is significantly more complex. The Jordan-Wigner code might be useful to forks of OpenFermion that don't involve CodeOperator. The performance is also likely better in the more simple implementation.

conta877: I agree that we should keep both. I also think its a good example of code transforms for people. I didn't check jordan-wigner implementation but the Bravyi-Kitaevs produce different qubitOperators.

msteudtner: Really ? Let me have a look at that - I tend to accidentally  switch the two matrices. In any case it seems to be a valid transform, since it passed the test.

babbush: A good test would be to see if the transformations are all isospectral. How different is the output of the transformation? If there is a broad family of Bravyi-Kitaev like transformations with truly different structures it would be quite interesting.

msteudtner: I just tested it the code transform vs. bravyi_kitaev, and frankly I am a bit confused.  $a^\dagger_i a_i$ should return a projector .5*(1-Z_F(i)), right ? so F(6) is empty, and I find the following:
>>> bravyi_kitaev(FermionOperator('6^ 6'))
(0.5+0j) [] +
(-0.5+0j) [Z3 Z5 Z6]
>>> binary_code_transform(FermionOperator('6^ 6'),bravyi_kitaev_code(19))
0.5 [] +
-0.5 [Z6]

babbush: Our BK transform is implemented as in arXiv:1701.07072.

msteudtner: Eq. 5 is what I mean (sorry it is  {F(i) U i}, not just F(i)). So it seems the flip sets are different.

kevinsung: If you run
```
bravyi_kitaev(FermionOperator('6^ 6'), n_qubits=8)
```
the output is
```
(0.5+0j) [] +
(-0.5+0j) [Z6]
```

msteudtner: indeed. if you set n_qubits=14, it says:
(0.5+0j) [] +
(-0.5+0j) [Z3 Z5 Z6]

Why is it qubit dependent ?

kevinsung: Because the mode with the highest index is always the root of the Fenwick tree (see arXiv:1701.07072).

kevinsung:
Is it true that arXiv:1208.5986 implicitly assumes that the number of qubits is a power of 2? Our implementation does not.

babbush:
@kevinsung no, that is not assumed in any way by arXiv:1208.5986.

src/openfermion/transforms/_code_transform_functions_test.py
msteudtner: Implementing our version, I followed the instructions in 1208.5986 and S. Tranter's paper. There  they segment the matrix if its in between two dimension that are consecutive integer powers of two. It seems like the Fenwick-tree mapping is different.

babbush:
I had been under the impression that the Fenwick tree transform was just a different way of implementing essentially the same mapping. But you seem to be correct that there are real difference. This is interesting and suggests that perhaps we should have both types of BK implemented. I am curious to understand more about these differences. If I recall correctly, @VojtaHavlicek implemented the Fenwick tree BK. But @kanavsetia or @jarrodmcc may also have some insights about this difference...

In any event, this is really an issue that is unrelated to the quality of the pull request under review here. I think if you show your BK is isospectral and your JW gives the same JW transform as elsewhere in OpenFermion, that should be enough.

src/openfermion/transforms/_code_transform_functions_test.py
kevinsung: @msteudtner I think you're right. Perhaps this PR should use our implementation of Fenwick trees (contained in `transforms`) to define the Bravyi-Kitaev code? @babbush do you have an opinion? It seems that the code definitions are fundamentally different between 1208.5986 and 1701.07072.

kevinsung: Ok I just saw your comment below.

kevinsung:
I have not scrutinized 1208.5986, but my understanding is that it defines the BK transform for powers of 2 and then reduces the case of a general integer by using its binary expansion and applying multiple power-of-2 BK transforms.

kanavsetia:
I think @kevinsung got it right. BK transform of 1208.5986 explicitly defines the transform for powers of 2 (as is evident from figure2 in the paper) but that is not the case for 1701.07072, which is more general. 
And I think this is what is causing the discrepancy. For 1208.5986 no matter how many qubits you use, even qubits will always hold just their occupancy, and odd qubits contain their occupancy as well as occupancy of other adjacent qubits. So, if you are looking for the occupancy of 6th qubit, no matter the total number of qubits (which in this case will always be powers of 2), you will always just operate of 6th qubit.
But this is little different for 1701.07072, where, which qubit stores what is also dependent on the number of qubits. If you have total number of qubits as some power of 2, I think you will get the same answer as in 1208.5986. But for general case this will be different.

babbush:
I understand what you are talking about now. I thought Kevin was saying it would only work for a binary power number of qubits.

kevinsung:
Yes, the Fenwick tree approach is more general, and I think it should be relatively straightforward to use the existing Fenwick tree code to define the BinaryCode used in `bravyi_kitaev_code` (I admit I don't know the details though).

babbush:
This would be a nice but I won't require it. So long as their Jordan-Wigner encoder matches our Jordan-Wigner and their Bravyi-Kitaev encoder gives the transformation from 1208.5986 while also being isospectral, to our JW/BK, I think that is good enough for now.

kevinsung:
Sure, I agree that this should not be a requirement for this PR. However, if we plan to keep two different implementations of BK in OpenFermion then we should remember to make it clear in the documentation.

src/openfermion/transforms/_code_transform_functions_test.py
kevinsung: Based on the discussion below, I'm content with just checking that the JW (not BK) gives the same QubitOperators as the existing implementation, and that the new BK gives isospectral operators (does not need to give identical QubitOperators).

conta877:
Ok - I will add the tests and in the description for BK I'll mention which paper the transformation is based on.

conta877:
ready for review!

src/openfermion/transforms/__init__.py
```diff
@@ -16,6 +16,7 @@
 from ._bravyi_kitaev import bravyi_kitaev
 from ._binary_code_transform import (binary_code_transform,
                                      dissolve)
+from ._code_transform_functions import *
```
babbush: please write explicitly what you are importing here so that it is clear to everyone what exists in the namespace of the transforms module.

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,254 @@
...
+    reps = int(numpy.ceil(numpy.log2(dimension)))
+    mtx = numpy.array([[1, 0], [1, 1]])
+    for repetition in numpy.arange(1, reps + 1):
+        mtx = numpy.kron(numpy.eye(2, dtype=int), mtx)
+        for column in numpy.arange(0, 2 ** repetition):
+            mtx[2 ** (repetition + 1) - 1, column] = 1
+    return mtx[0:dimension, 0:dimension]
+
+
+def _decoder_bk(modes):
```
babbush: I would recommend calling this n_modes to be consistent with the sort of notation we use elsewhere.

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,254 @@
...
+
+""" Pre-existing codes for Fermion-qubit mappings
+    based on (arXiv:1712.07067) """
+
+import numpy
+
+from openfermion.ops import BinaryCode, SymbolicBinary, linearize_decoder
+
+
+def _encoder_bk(dimension):
```
babbush: There is probably some obvious reason why you use "dimension" here and "modes" below?

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,254 @@
...
+    """
+    encoder = numpy.zeros((exponent, 2 ** exponent), dtype=int)
+    decoder = [0] * (2 ** exponent)
+    for counter in numpy.arange(2 ** exponent):
+        encoder[:, counter], decoder[counter] = \
+            _binary_address(exponent, counter)
+    return BinaryCode(encoder, decoder)
+
+
+def weight_one_segment_code():
```
babbush: I'm a little confused why this function that only works for 2-qubits is necessary.

src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -0,0 +1,254 @@
...
+    Note:
+        This code is highly non-linear and might produce a lot of terms.
+
+    Returns (BinaryCode): weight one segment code
+    """
+    return BinaryCode([[1, 0, 1], [0, 1, 1]],
+                      ['w0 w1 + w0', 'w0 w1 + w1', ' w0 w1'])
+
+
+def weight_two_segment_code():
```
babbush: Same question as above.

conta877:
ready!

PR #233, opened 2018-03-02
Add EqualsTester. Add eq/ne/hash/repr/str methods to Grid.

Strilanc:
`EqualsTester` is an extremely useful utility to have. Not only does it make writing equality tests shorter (because you just add a bunch of disjoint groups), it tests easily-overlooked cases such as:

- `__hash__` being inconsistent with `__eq__`
- `__ne__` being inconsistent with `__eq__`
- Comparing as equal to unrecognized types
- Equality not being commutative
- Equality not being reflexive


src/openfermion/utils/_grid_test.py
```diff
@@ -57,3 +57,29 @@ def test_properties(self):
...
+        self.assertTrue(g == g2)
+        self.assertFalse(g != g2)
+        self.assertEqual(g, g2)
+        self.assertEqual(hash(g), hash(g2))
+
+        self.assertFalse(g == Grid(dimensions=4, length=3, scale=0.5))
+        self.assertFalse(g == Grid(dimensions=5, length=2, scale=0.5))
+        self.assertFalse(g == Grid(dimensions=5, length=3, scale=0.75))
+
+        self.assertTrue(g != '')
```
idk3: `self.assertNotEqual()` will work here :)

src/openfermion/utils/_grid_test.py
```diff
@@ -57,3 +57,29 @@ def test_properties(self):
...
+        self.assertFalse(g == '')
```
idk3: self.assertEqual()

src/openfermion/utils/_grid_test.py
```diff
@@ -57,3 +57,29 @@ def test_properties(self):
...
+            str(Grid(dimensions=5, length=4, scale=0.5)),
+            '[-0.25, -0.125, 0.0, 0.125]**5')
+
+    def test_equality(self):
+        g = Grid(dimensions=5, length=3, scale=0.5)
+        g2 = Grid(dimensions=5, length=3, scale=0.5)
+
+        self.assertTrue(g == g2)
+        self.assertFalse(g != g2)
+        self.assertEqual(g, g2)
```
idk3: Why do you use both assertEqual and assertTrue(==)?

Strilanc:
What's up with the continuous integration? It looks like it's trying to pull from this repo instead of the fork, and (no surprise) failing to find the branch.

src/openfermion/utils/_grid_test.py
```diff
@@ -57,3 +57,29 @@ def test_properties(self):
...
+        self.assertEqual(hash(g), hash(g2))
```
idk3: Why does this work? I'm not used to hashes necessarily being equal for user defined classes.

idk3:
@Strilanc maybe if you re-push it'll be happy? Travis has done some weird stuff, but this is the first time I've seen these particular errors

src/openfermion/utils/_grid_test.py
Strilanc: `assertNotEqual` and `assertEqual` assume the class has properly-functioning equality. You have to be careful to avoid using them (or only use them sparingly) when testing equality in the first place.

For example, do you know whether `assertNotEqual` calls `==` or `!=` internally? Or maybe it calls the `__eq__` method on both sides and check for consistency? And regardless of what the code says, what does the library *guarantee*?

In this case, I'm making sure that the correct behavior occurs when an incompatible type `==`'d against Grid. Replacing it with `assertEqual` is not guaranteed to test the same thing.

src/openfermion/utils/_grid_test.py
Strilanc: Not guaranteed to test the same thing.

src/openfermion/utils/_grid_test.py
Strilanc: The `assertEqual` is just a sanity check on top of the other two.

src/openfermion/utils/_grid_test.py
Strilanc: The two instances were defined with the same values, and the class supports value equality, therefore the hashes must be equal.

src/openfermion/utils/_grid_test.py
idk3: Can you provide a case / instance where this isn't true? The docs for unittest explicitly say "This method should also be avoided when more specific methods are available (e.g. assertEqual(a, b) instead of assertTrue(a == b)), because they provide a better error message in case of failure."

src/openfermion/utils/_grid_test.py
idk3: Oh, I missed that you defined __hash__ for Grid. AFAIK Python's hash is basically blind to value equality, e.g. the following code will raise an error in general:

`class Integer(object):
	def __init__(self, i):
		self.i = i
	def __eq__(self, other):
		return self.i == other.i
a = Integer(3); b = Integer(3)
assert hash(a) == hash(b)`

idk3: (Sorry for the lack of line breaks, I did my best.. :( )

src/openfermion/utils/_grid_test.py
Strilanc: Sure:

```python
    def test_tricky_equality_typo(self):
        class C():
            def __init__(self, x):
                self.x = x

            def __eq__(self, other):
                return self.x == other.x

            def __ne__(self, other):
                # Oops: I accidentally wrote self instead of other
                return self.x == self.x

        self.assertEqual(C(1), C(1))
        self.assertNotEqual(C(1), C(2))
```

It is impossible to write a test that will detect the typo using `assertEqual` and `assertNotEqual`.

... you know what, it makes sense to add an EqualsTester class along with this. Hold on...

Strilanc:
I added an `EqualsTester` class to this PR, and used it to implement the Grid equality tests.

Strilanc:
I figured out what went wrong. I accidentally pushed the branch to this repo at one point, then deleted it. Travis must have triggered against _that_ branch. Then, because status is set by commit id and that branch was at the same commit, travis was setting statuses against this PR. And complained that the branch that I deleted was not found.

babbush:
@Strilanc can you explain why this is important for OpenFermion? What are we hashing that we're worried about? How do you see this being useful?

Strilanc:
@babbush `EqualsTester` is a utility class that's generally useful for any library.

It is very common to implement classes that should have value equality instead of reference equality.
 For example, think of something you'd like to check against an expectation in tests, such as `QubitOperator`. I'm guessing you sometimes write tests that basically say "This QubitOperator should be X1 X2+Z3 Z4", right? That means `QubitOperator` should implement an `__eq__` method that makes such checks easy to do. `EqualsTester` then makes it easy to check that the `__eq__` method and friends were implemented correctly.

src/openfermion/utils/_grid_test.py
Strilanc: Use triple-mark to make code blocks. Triplet-mark then python to get highlighting.

image

```python
def oh():
    pass
```

kevinsung:
I just want to throw out a suggestion: we keep the EqualsTester class but put it in `utils/_testing_utils.py`. I can at least see that it could be used to simplify a lot of our existing tests, which we might want to do in the future.

idk3:
I'm inclined to agree with @kevinsung. I think there are places this could come in handy. My one complaint is that I don't see the need for the equals-implies-same-hash requirement - I worry this will make it difficult to use the code more widely within the code base.

Strilanc:
@babbush Oh, were you thinking this is for the *users* of the library? That wasn't the intention. I agree that that would be weird for a chemistry library. `EqualsTester` is an internal tool for the developers.

@idk3 

> I don't see the need for the equals-implies-same-hash requirement

If you implement a hash function, and it doesn't satisfy the "if equal then same hash" requirement, bad things happen. For example, you may add an object into a set and then ask the set if the object is present and have the set respond "nope!".

Of course, it does make sense for some classes to have `__eq__` but not have `__hash__` at all. For example, `list` works this way. `EqualsTester` takes this into account, and will happily allow unhashable types through (as long as the hashability is consistent within the given equivalence groups).

Strilanc:
I've moved `EqualsTester` into `_testing_utils.py`

babbush:
I recognize that this is for developers. But the way people tend to use OpenFermion, I make much less of a distinction than usual between developers and users. I'm still rather keen on seeing how this code will help us avoid problems that might occur in OpenFermion. For instance, can you point out how an error in, e.g. FermionOperator, might creep up without us using this class. Perhaps @idk3 or @kevinsung can provide an example use case for us since they seem to want the class.

idk3:
@babbush @kevinsung IMO `EqualsTester` gives a clean and clear framework for a lot of testing we do for equality and inequality. I can write a test that specifies a set of groups of things (which I expect to be equal within groups, unequal across) and the Tester will do the rest of it for me. This rough pattern appears throughout our equality tests and will make it easier to test more thoroughly.

@Strilanc 

> If you implement a hash function, and it doesn't satisfy the "if equal then same hash" requirement, bad things happen. For example, you may add an object into a set and then ask the set if the object is present and have the set respond "nope!".
> 
> Of course, it does make sense for some classes to have __eq__ but not have __hash__ at all. For example, list works this way. EqualsTester takes this into account, and will happily allow unhashable types through (as long as the hashability is consistent within the given equivalence groups).

The problem I'm concerned about is that this will require many classes to implement `hash`, or lead to unexpected problems down the road. The current if check in https://github.com/Strilanc/OpenFermion/blob/928a33e346d875e0450b1fd5246669977c47a878/src/openfermion/utils/_testing_utils.py#L195 with `collections.Hashable` is risky since all new-style classes are subclasses of `collections.Hashable` by default. Because of this, I think it would be simpler and more importantly safer to separate the hashing requirement into e.g. something like `EqualsAndHashTester`. Hashing is something we haven't really used in the library otherwise, and including it in `EqualsTester` appears to reduce rather than increase its utility (we could use it pretty much everywhere immediately, except for the hash requirement). Then we could use `EqualsAndHashTester` when we're sure we want hashing and `EqualsTester` everywhere else.

Thoughts?

Strilanc:
@idk3 Classes opt out of being hashable by specifying `__hash__ = None`. Any class that implements `__eq__` but doesn't override or set `__hash__` to None is a buggy class (it will not work correctly w.r.t. sets and dicts). Therefore I think it's appropriate for `EqualsTester` to check for this.

idk3:
My concern is that this will lead to problems for users who don't know this and as a result hit errors they won't understand in EqualsTester (presumably most users who define a new-style class). We can clean it up ourselves, but I'm fine with sacrificing purity for simplicity and sheer usability here. Especially since this seems to be the first PR defining `hash`, continuing to not respect equals implies hash equals doesn't seem to be a big issue. People more familiar with this or who more explicitly need it can use `EqualsAndHashTester` when it comes up, and people who only need equality testing can use `EqualsTester` without confusion.

idk3:
(To be clear, I'm OK with it if others don't think these fears of potential pitfalls are realistic. I think there are degrees to which we should respect principles over practicality, but that the principle in this case doesn't really come up in OpenFermion - WYT @babbush @kevinsung ?)

kevinsung:
@babbush I realized this isn't directly applicable to SymbolicOperator because we defined `isclose` rather than `__eq__`. However, if we do define `__eq__` then many of the SymbolicOperator tests (for instance, those that currently test `isclose`) can be simplified and made more readable. Our tests also don't extensively cover "not isclose" being consistent with "isclose", and this would automatically test for that. This leads me to the question, why do we have `isclose` for SymbolicOperator, while we have `__eq__` for PolynomialTensor? Do you see a reason not to define `__eq__` for SymbolicOperator?

@idk3 @Strilanc I think we should enforce equals implies hash equals. We can help the user out by adding a very explicit error message in case EqualsTester detects inconsistent hashes.

idk3:
Fair point @kevinsung - good error messages would be an easier way to fix my complaints ;)

@babbush thoughts on the `isclose` vs `__eq__` issue? In the past we switched over to `isclose` (a purity issue) but I'm ok with changing everything over to `__eq__`. The one catch is we'd have to decide on a scheme for rounding before hashing (e.g. convert to complex, round depending on `EQ_TOLERANCE`, then hash using that value).

kevinsung:
We can also define `__eq__` and set `__hash__` to `None`.

Strilanc:
What do you want the error message to be? Currently the user who forget to clear `__hash__` would get this:

```
Items in the same group produced different hashes. Example: hash(foo) is 4343282 but hash(bar) is 328264.
```

kevinsung:
Maybe add "Did you define __eq__ for a class but forget to set __hash__ to None?" That seems to be the pitfall @idk3 is concerned about.

idk3:
I think that's a good solution :) Do you know if it's possible to nicely check if a class is using `object`'s hash?

babbush:
Sorry I've been quiet on this - its been an extremely busy few days. I am still very concerned that we don't currently have a need for this. In particular, I don't think that any of our classes have __eq__ or __hash__ defined! The problem is that there are floats in our data structures. If I multiply QubitOperator by 3 and then divide by 3, is it equal to what I had at the start? Who knows? This is why we have .isclose().

Unless somebody can show me a compelling example of how we would use this class right now, I am inclined to close the PR. We can always reopen and merge down the line if a need for it arises.

babbush:
I spoke with @Strilanc about this. We will close for now.

kevinsung:
I'm working on a transforms demo, and I found that I want to write the lines
```
a_2 = FermionOperator('2')
a_2_dag = FermionOperator('2^')
a_5 = FermionOperator('2')
a_5_dag = FermionOperator('2^')

zero = FermionOperator()
identity = FermionOperator(())

assert anticommutator(a_2, a_2_dag) == identity
assert anticommutator(a_2, a_5) == zero
assert anticommutator(a_2, a_5_dag) == zero
```

I think the assertions at the end look nice that way. So I want to define `__eq__` for SymbolicOperator. I don't see any harm in doing that and making it do the same thing as `isclose` (and setting `__hash__` to `None`) for now. @babbush, what do you think?

I went ahead and did it on one of my branches and then realized that EqualsTester would significantly reduce the burden of writing tests for it.

babbush:
Very well!

But let's talk about implementing equality. My concern is that since coefficients are floats then if we multiply by 3.17 and divide by 3.17 we might not have the "same" operator. If we're going to do this then I think that we should get rid of ".isclose" in factor of __eq__ and we'll have __eq__ use the tolerance in the ._config.py. We should then also define __neq__ and __hash__ for all operators. This should also be implemented for PolynomialTensor.

kevinsung:
@babbush I'm proposing that we just make `__eq__` call `isclose`. With this, multiplying and then dividing by the same number would result in the "same" operator. Currently `isclose` uses the `EQ_TOLERANCE` defined in `config.py`. We would end up with two ways of doing the same thing; do you think that's bad?

PolynomialTensor already implements `__eq__` the way you described (and does not have `isclose`).

kevinsung:
One way to implement `__hash__` is to round all coefficients before storing them. @Strilanc do you have a better idea?

idk3:
@kevinsung I originally thought rounding could work for hash but realized it isn't possible. The problem is that if we e.g. round to nearest 1, then 2 and 3 should be equal, and 3 and 4 should be equal, i.e. hash(2) = hash(3) and hash(3) = hash(4). But then in fact hash must report the same value for any number (contradiction / hash failure).

babbush:
@kevinsung we should not have __eq__ and isclose() methods which do the same thing. We should stick with one or the other.

idk3:
@babbush @kevinsung my earlier comment suggests that the current __eq__ for these operators is fundamentally inconsistent with __hash__ - isclose is probably the safest way to do this.

kevinsung:
@idk I admit that in your example having 2 and 3 round to different things is unfortunate and would nullify the guarantee that multiplying then dividing by the same number will yield the "same" object, but perhaps it doesn't matter! In any case, is there a problem with just having SymbolicOperator be unhashable? As it stands, hashing a SymbolicOperator is useless and SymbolicOperators should not be used as keys in dictionaries, so perhaps we should already be setting `__hash__` to `None`.

kevinsung:
If we round `a = 2.4` to 2 and `b = 2.6` to 3 then we will be "wrong" about `a == b`, but we will be "wrong" by at most our `EQ_TOLERANCE`, so I think that's fine. But again, I also don't see a big problem with defining `__eq__` and setting `__hash__` to `None` until we have a compelling reason to support hashing SymbolicOperators. Hashing floating-point objects is an intrinsically tricky issue.

idk3:
@kevinsung I don't understand what you're saying here - in that case `a` and `b` would still be within 1 after rounding so we'd be OK. The problem is that this definition of `__eq__` is inconsistent with *any* hashing function as @Strilanc points out. Hashing with floats is possible if we just don't round in `__eq__`.

kevinsung:
@idk3 The meaning of `__eq__` and `__hash__` being consistent is that `a == b` evaluating to True implies that `hash(a) == hash(b)` evaluates to True as well. My proposal is that
- The `terms` dictionary of SymbolicOperator only ever stores rounded values, i.e., if we want to enter `coef`, we enter `round(coef, EQ_PRECISION)` instead
- `__eq__` is implemented by checking that all the coefficients are exactly the same (no tolerance)
- `__hash__` is implemented by hashing an ordered collection of the coefficients (or something like that)

I believe these definitions are consistent.

idk3:
While this would give consistent `__eq__` and `__hash__`, the problem is that the user needs to be able to decide when to round. If we round after every computation we'll introduce a lot of floating point weirdness which I don't think is justified by the need for `__hash__` consistency. My suggestion would be to avoid rounding, and use `isclose` for the approximate equality checks.

kevinsung:
No, I'm proposing that the rounding be done internally by the SymbolicOperator class, so the user never has to think about rounding.

Also, what do you think of defining `__eq__` to do what `isclose` does now, and setting `__hash__` to `None`?

idk3:
My point is that this is dangerous - we get the regular floating point issues, but magnified by orders of magnitude (depending on `EQ_TOLERANCE`). For example (x / 100) * 100 will now be different from the original x by as much as 100 * `EQ_TOLERANCE` for each term in the original x. Across a large number of terms this difference could be significant, and the differences also accumulate with every operation performed.

I think my preference is for the reverse, having `__eq__` be an aggressive "true" equality, and having `isclose` be where we might do checking against `EQ_TOLERANCE`.

kevinsung:
I'm not convinced that this is that dangerous. Similar things can already happen whenever some code checks `EQ_TOLERANCE`, and if we are really concerned about small differences like this we can make `EQ_TOLERANCE` even smaller.

My other proposal is that we just make SymbolicOperator unhashable (I think this makes sense). Assuming we do this, I don't see any reason not to use `==` for approximate equality check. That is a very natural syntax, and it's the sort of thing that overriding `__eq__` is meant to be used for. Since we are dealing with floating point numbers, there is no reason to ever check for exactly equality that I can think of.

babbush:
I suppose using == to mean "isclose". But I'd like to see a PR that gets rid of .isclose() and adds "==" before merging this. I think making the data structure unhashable is a reasonable idea.

kevinsung:
Ok I'm making the PR to define `__eq__`. I just noticed that `isclose` takes in two extra parameters, `rel_tol` and `abs_tol`. There are some tests in `_low_depth_trotter_error_test.py` which actually use `isclose` with ` rel_tol=1e-8`, and they fail if the tolerance is set to `EQ_TOLERANCE`.

Do you really want to get rid of `isclose`? In that case we can't use these tolerance parameters anymore. I'm fine with that; we'd just need to rewrite this test or make it pass in a hacky way (e.g. just redefine `isclose` for that test). Please advise @babbush .

babbush:
I think we should get rid of isclose and rel_tol and just have one "tolerance" parameter. @idk3 what is going on with rel_tol in the trotter error module?

kevinsung:
`__eq__` can't take any extra parameters so we'll have to use the global `EQ_TOLERANCE` in all cases. I think that makes sense.

To be clear there is just one place in the entire repository that this breaks, and that is following test:
```
    def test_all_terms_in_dual_basis_jellium_hamiltonian(self):
        grid_length = 4
        dimension = 1

        # Generate the Hamiltonian.
        hamiltonian = dual_basis_jellium_hamiltonian(grid_length, dimension)

        terms = ordered_low_depth_terms_no_info(hamiltonian)
        FO = FermionOperator

        expected_terms = []
        for i in range(grid_length ** dimension):
            expected_terms.append(FO(((i, 1), (i, 0)),
                                     0.018505508252))
            expected_terms.append(FO(((i, 1), ((i + 1) % 4, 0)),
                                     -0.0123370055014))
            expected_terms.append(FO(((i, 1), ((i + 2) % 4, 0)),
                                     0.00616850275068))
            expected_terms.append(FO(((i, 1), ((i + 3) % 4, 0)),
                                     -0.0123370055014))
            expected_terms.append(normal_ordered(
                FO(((i, 1), ((i + 1) % 4, 1), (i, 0), ((i + 1) % 4, 0)),
                   3.18309886184)))
            if i // 2:
                expected_terms.append(normal_ordered(
                    FO(((i, 1), ((i + 2) % 4, 1), (i, 0), ((i + 2) % 4, 0)),
                       22.2816920329)))

        for term in terms:
            found_in_other = False
            for term2 in expected_terms:
                if term.isclose(term2, rel_tol=1e-8):
                    self.assertFalse(found_in_other)
                    found_in_other = True
            self.assertTrue(found_in_other, msg=str(term))
        for term in expected_terms:
            found_in_other = False
            for term2 in terms:
                if term.isclose(term2, rel_tol=1e-8):
                    self.assertFalse(found_in_other)
                    found_in_other = True
self.assertTrue(found_in_other, msg=str(term))
```

babbush:
I believe this is the work of @idk3

kevinsung:
@idk3 would you mind rewriting that test so that it passes with all tolerances set to `EQ_TOLERANCE`? Perhaps you just need to add some more digits to the hard-coded values.

idk3:
@kevinsung https://github.com/quantumlib/OpenFermion/pull/264 should address :)

kevinsung:
@idk3 is there still anything about this PR you don't approve?

idk3:
@kevinsung nope :P but I think your approval would override my disapproval (not totally sure though)

kevinsung:
@babbush we need you to merge this.

PR #236, opened 2018-03-03
Clean up import statements and fix sphinx warnings

kevinsung:
- Removed some in-line `import` statements that are no longer necessary because of improved import ordering in some `__init__.py` files
- Alphabeticized some imports
- Made it so that `from openfermion import *` doesn't also import `EQ_TOLERANCE` from `config.py`. I think generally users generally expect `*` imports to import only functions and not constants. Anyway I originally only added it because I thought it was necessary to make some things work but I think I was mistaken.
- Fix sphinx warnings

### DIFF
--- a/src/openfermion/ops/__init__.py
+++ b/src/openfermion/ops/__init__.py
@@ -13,15 +13,15 @@
 from ._binary_operator import SymbolicBinary
 from ._code_operator import (BinaryCode,
                              linearize_decoder)
-from ._polynomial_tensor import general_basis_change, PolynomialTensor
+from ._polynomial_tensor import PolynomialTensor, general_basis_change
 from ._quadratic_hamiltonian import QuadraticHamiltonian
 from ._symbolic_operator import (SymbolicOperator,
                                  prune_unused_indices)
 
 # Imports out of alphabetical order to avoid circular dependency.
 from ._fermion_operator import (FermionOperator,
-                                normal_ordered,
-                                freeze_orbitals)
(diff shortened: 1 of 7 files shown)


src/openfermion/ops/_interaction_rdm.py
```diff
@@ -17,10 +17,10 @@
 import numpy
 
 from openfermion.ops import (FermionOperator,
-                             PolynomialTensor,
                              InteractionOperator,
-                             normal_ordered,
-                             QubitOperator)
+                             PolynomialTensor,
+                             QubitOperator,
+                             normal_ordered)
```
idk3: Is this out of alphabetical order to avoid a circular import issue?

kevinsung: Capital letters come before lower case letters (at least according to Pythons `ord` function :P)

src/openfermion/ops/__init__.py
```diff
@@ -13,13 +13,15 @@
 from ._binary_operator import SymbolicBinary
 from ._code_operator import (BinaryCode,
                              linearize_decoder)
+from ._polynomial_tensor import general_basis_change, PolynomialTensor
```
idk3: Then this should be changed ;)

idk3: @kevinsung oops

kevinsung: Haha ok.

PR #237, opened 2018-03-03
rename files for BinaryCode and SymbolicBinary

kevinsung:
I renamed the files containing BinaryCode and SymbolicBinary to follow our practice of naming these data structure files by the main data structure they define.


kevinsung:
@conta877 I dunno if you have something to say about this.

conta877:
totally fine!
suggestions for the code_transform_functions while I'm at it - code_transformers?

jarrodmcc:
Looks like @conta877  is on board as is everyone else, so I'll approve these changes.

PR #239, opened 2018-03-04
Adjustments to sparse eigenvalue and expecation

jarrodmcc:
Saw that some recent issues were copying our code exactly from sparse_tools which has a lot of small defects and inefficiencies.  I added the optional parameter to encourage the use of initial guesses, which matter a lot for efficiency in eigenvalue routines based on sparse techniques.  Also modified expectation to allow non-sparse matrix arrays.  Frankly all the use of sparse matrices in place of ndarray vectors should be stripped out.  For most use cases it will be way less efficient and the code is more complex.  Sparse matrices make sense in most cases, using sparse matrices for general vectors typically does not.


babbush:
Tests seems to fail for python 3.

idk3:
@jarrodmcc could you also add a few more tests to PolynomialTensor? In particular there are gaps in the tests for `__eq__` and `__neq__` is not tested at all. One of the untested lines https://coveralls.io/builds/15795520/source?filename=src%2Fopenfermion%2Fops%2F_polynomial_tensor.py#L179 doesn't seem correct (assumes the output of `keys()` will be sorted).

PR #242, opened 2018-03-05
updated arXiv reference

babbush:
Do not merge until revision has been uploaded to arXiv.


babbush:
I also updated the demo on the recent contributions to display the output, so people can see what it does without needing to run the code.

babbush:
If coveralls doesn't do its thing in a couple minutes just force it...

PR #247, opened 2018-03-11
Implement LCU coefficient preprocessing used for constructing efficient prepare circuits

Strilanc:
This is *almost* exactly what we want. The discretized probabilities don't have a denominator guaranteed to be a power of 2.


src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Utility methods for LCU circuits."""
+
+from __future__ import absolute_import
+from __future__ import division
```
babbush: For consistency with our style guide, we'd prefer you import them both in the same line. Also, what is the precise reason you need to import these things?

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+    total = 0
+    for v in vals:
+        yield total
+        total += v
+    yield total
+
+
+def _differences(weights):
+    p = None
+    have_p = False
```
babbush: I know what p and have_p are because I know how this is all supposed to work. But it might be good to have a little more description of what this function is doing and what p / have_p are so somebody looking at this code without our paper could have a chance at figuring out what is going on,.

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+            yield w - p
+        p = w
+        have_p = True
+
+
+def _discretize_probability_distribution(unnormalized_probabilities, epsilon):
+    """Approximates probabilities with integers over a common denominator.
+
+    Args:
+        unnormalized_probabilities: A list of non-negative numbers proportional
```
babbush: just non-negative, or also real?

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+
+        It is guaranteed that the common denominator is a multiple of the
+        number of probabilities in the distribution.
+    """
+    n = len(unnormalized_probabilities)
+    bin_count = math.ceil(1.0 / (epsilon * n)) * n
+
+    cumulative = list(_partial_sums(unnormalized_probabilities))
+    total = cumulative[-1]
+    discretized_cumulative = [int(math.floor(c/total*bin_count + 0.5))
```
babbush: please have spaces around / and *. Did you run pylint?

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+    """Approximates probabilities with integers over a common denominator.
+
+    Args:
+        unnormalized_probabilities: A list of non-negative numbers proportional
+            to probabilities from a probability distribution. The numbers may
+            not be normalized (they do not have to add up to 1).
+        epsilon: The absolute error tolerance.
+
+    Returns:
+        A tuple containing:
```
babbush: We can discuss here whether this is a poor practice, but in other parts of our code it has been common for us to talk about functions just returning two different objects. For instance, we would write under the returns line:

numerators_list(list): a list of of floats giving the numerators.
common_denominator(float): the common denominator.

unless you have a specific reason, I think this would be preferred just for consistency with other parts of the library. I'm not actually sure that sphinx will link the docs correctly if you do it your way.

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+    superior.
+
+    Args:
+        discretized_probabilities: A list of probabilities approximated by
+            integer numerators (with an implied common denominator). In order
+            to operate without floating point error, it is required that the
+            sum of this list is a multiple of the number of items in the list.
+
+    Returns:
+        A tuple containing:
```
babbush: Same comment as above regarding how to discuss returns.

src/openfermion/utils/_lcu_util.py
```diff
@@ -0,0 +1,181 @@
...
+    an efficient preparation method of a state where the computational basis
+    state |k> has an amplitude proportional to the coefficient.
+
+    Args:
+        lcu_coefficients: A list of non-negative floats, with the i'th float
+            corresponding to the i'th coefficient of an LCU decomposition
+            of the Hamiltonian (in an ordering determined by the caller).
+        epsilon: Absolute error tolerance.
+
+    Returns:
```
babbush: Same comment as above.

src/openfermion/utils/_lcu_util_test.py
```diff
@@ -0,0 +1,181 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Tests for _lcu_util.py."""
+
+from __future__ import absolute_import
```
babbush: import in one line? Also, again why do you feel this is necessary? Travis would be fine without it.

babbush:
@Strilanc incidentally, if you could figure out how to make the Google CLA bot stop doing this on literally every other pull request, it would be a huge help.

maffoo:
It looks like the problem with the cla bot is that master got merged into this branch, and there are commits by other people on master. If the branch had been rebased on master instead, then the commits would all still be by the same person and the bot would not have complained. It should be possible for the bot to recognize that the other commits are on the branch that this PR is going to be merged into and can therefore be ignored, but clearly it doesn't do that currently.

babbush:
Thanks @maffoo - what is the simplest way then for me to update branches that would avoid CLA bot getting mad. I suppose I should rebase on master myself? That sounds a lot harder than pressing the "update branch button".

kevinsung:
@maffoo with your suggestion the bot would continue to post on PR's submitted from out-of-date branches. That may be reasonable, but even still, if the user left the box "Allow edits from maintainers" checked, then they are giving us their consent to edit their branch so I think a better bot would not chime in.

maffoo:
@babbush I hadn't realized this was due to someone other than the PR submitter (e.g. an admin) updating the branch. We should check to see whether this still happens if the PR submitter pushes the "update branch" button instead, since in that case the merge commit will be authored by the PR submitter, and the CLA bot might be smart enough to figure out what is going on. If that doesn't work, the bot itself would probably have to be fixed...

Strilanc:
Yeah, CLA bot always seems unhappy after Ryan merges updates into the branch.

Personally, I think the workaround is not to touch other people's branches. You should rely on the PR author to keep their changes synced and update to master at their leisure, unless the PR is ready to go and approved and about to be merged so no more changes are expected. Otherwise you're focusing them to do a redundant merge after their next `git push` call fails because the local copy has gone down a different path than the remote copy.

src/openfermion/utils/_lcu_util_test.py
Strilanc: `numers[i] / denom` would return 0 instead of a proper result without this import.

src/openfermion/utils/_lcu_util.py
Strilanc: Done

src/openfermion/utils/_lcu_util.py
Strilanc: Done.

src/openfermion/utils/_lcu_util.py
Strilanc: Done

src/openfermion/utils/_lcu_util.py
Strilanc: Done.

No I didn't run pylint. Continuous integration tools are supposed to do that! I tried the linter suggested in the documentation (`pycodestyle`) but it doesn't return any errors for these files regardless of how I do the math expressions...

src/openfermion/utils/_lcu_util.py
Strilanc: Changed from `numbers` to `floats`.

src/openfermion/utils/_lcu_util.py
Strilanc: Done.

src/openfermion/utils/_lcu_util.py
Strilanc: Done.

I imported division because I do divisions in this file, and without this import they would potentially be integer divisions and round down. Actually, I think we should unconditionally include this future import in every file to avoid possible compatibility issues between python 2 and python 3.

src/openfermion/utils/__init__.py
```diff
@@ -70,3 +70,5 @@
                             qubit_operator_sparse,
                             sparse_eigenspectrum,
                             variance)
+
+from ._lcu_util import preprocess_lcu_coefficients_for_reversible_sampling
```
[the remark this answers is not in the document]
Strilanc: Done.

Strilanc:
I redid the merge into master and force-pushed over Ryan's, and responded to comments. Hopefully CLA bot will be happy now.

Strilanc:
Oh, I see that cla bot is saying that the PR is cla-borked forever now. Oh well.

babbush:
We usually force it when this happens. Which is why our PR history looks
crazy

babbush:
I will merge once tests pass. Thanks Craig.

babbush:
@Strilanc why is coveralls saying 0% test coverage!?

Strilanc:
I'm going to go out on a limb and say that the problem has something to do with this NaN in the line count of data/:

image

Strilanc:
Ran it again. Worked this time. I'm starting to think the CI tools just don't like Ryan.

I also did some... magic... to make the `cla/google` check pass.

Issue #251, opened 2018-03-12
Binary code transforms map creation operators to annihilation operators


kevinsung:
Executing the lines
```
ferm_op = FermionOperator('2')
n_modes = 5
qubit_op = binary_code_transform(ferm_op, parity_code(n_modes))
print(qubit_op)
```
yields
```
0.5 [Z1 X2 X3 X4] +
(-0-0.5j) [Y2 X3 X4]
```
However, according to Eq. (19) of arXiv:1208.5986, the second term should have a plus rather than minus sign. @conta877 or @msteudtner do you have any idea why?

conta877:
Yeah I agree with you based on eq 19. 
I'll wait @msteudtner to comment on this.

msteudtner:
Hi, I can take a look at it tomorrow, but I forsee that there isn't much I
can do: the parity code gives isospectral results (we have a test) and I
used the same matrix as in their paper. Are we sure the sign isnt flipped
in eq. 20 and 21 ?

Am 12.03.2018 22:19 schrieb "Isil Ozfidan" <notifications@github.com>:

kevinsung:
I'm quite sure that you're mapping creation operators to annihilation operators and vice versa. This cannot be tested for by looking at the spectrum (particle-hole transformations are just a change of basis). However, if we map all the way to sparse matrices, then we can initialize a state with only mode 1 occupied, and try to annihilate mode 1. This should give the vacuum state, but instead it just gives the zero vector:

```
annihilate_1 = binary_code_transform(FermionOperator('1'), parity_code(2))
annihilate_1_sparse = get_sparse_operator(annihilate_1)

state = jw_configuration_state(occupied_orbitals=[1], n_qubits=2)
result = annihilate_1_sparse * state
print(result.toarray())
```
yields
```
[[0.+0.j]
 [0.+0.j]
 [0.+0.j]
 [0.+0.j]]
```
Note that in this situation the initial state is represented by the string `01` in both the Jordan-Wigner and parity bases.

conta877:
it is isospectral. sign is not flipped in eq21. (the annihilation operator)
This is actually a pretty deep bug - awesome catch @kevinsung 

In the parity operator way it is defined currently the decoder is not the inverse of the encoder because its mod2. 
I believe the way this should be handled is when there is a -1 in a decoder matrix, in linearization one should look at its absolute value to introduce the qubit variable and its sign to add 1. 

I'll push a suggestion that I think @msteudtner should review.

[...]

PR #253, opened 2018-03-13
parity code bug chase

conta877:
addressing bug reported in #251

I'm not 100% if this is the right way of fixing this, as in its not gonna break some other code in the future. @msteudtner should review this change.

### DIFF
--- a/src/openfermion/transforms/_code_transform_functions.py
+++ b/src/openfermion/transforms/_code_transform_functions.py
@@ -164,10 +164,8 @@ def parity_code(n_modes):
 
     Returns (BinaryCode): The parity transform BinaryCode
     """
-    dec_mtx = numpy.reshape(([1] + [0] * (n_modes - 1)) +
-                            ([1, 1] + (n_modes - 1) * [0]) * (n_modes - 2) +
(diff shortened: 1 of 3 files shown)


src/openfermion/transforms/_code_transform_functions.py
```diff
@@ -164,10 +164,8 @@ def parity_code(n_modes):
 
     Returns (BinaryCode): The parity transform BinaryCode
     """
-    dec_mtx = numpy.reshape(([1] + [0] * (n_modes - 1)) +
-                            ([1, 1] + (n_modes - 1) * [0]) * (n_modes - 2) +
-                            [1, 1], (n_modes, n_modes))
     enc_mtx = numpy.tril(numpy.ones((n_modes, n_modes), dtype=int))
+    dec_mtx = numpy.linalg.inv(enc_mtx)
```
kevinsung: Since the decoder matrix should have a simple form here, I'd rather have it coded directly rather than perform a relatively expensive matrix inversion.

kevinsung:
I see that the function `linearize_decoder` takes in a matrix. Should this always be a matrix of zeros and ones?

PR #254, opened 2018-03-13
transform operator mixup fix

conta877:
ok this time its a legit fix for #251 

this was a bug introduced in refactoring of the transform function.
XZ = -ZX, was multiplying the transform_term with the update from right instead of left.

Also added @kevinsung s tests.

### DIFF
--- a/src/openfermion/transforms/_code_transform_functions_test.py
+++ b/src/openfermion/transforms/_code_transform_functions_test.py
@@ -14,7 +14,7 @@ import unittest
 
 from openfermion.transforms._code_transform_functions import *
 from openfermion.transforms import jordan_wigner
-from openfermion.ops import FermionOperator
+from openfermion.ops import FermionOperator, QubitOperator
 from openfermion.hamiltonians import MolecularData
 from openfermion.transforms import binary_code_transform
 from openfermion.transforms import get_fermion_operator
@@ -53,9 +53,10 @@ class CodeTransformTest(unittest.TestCase):
                           ((2, 'Y'), (3, 'X'), (4, 'X')): -0.5j})
 
         ferm_op = FermionOperator('5^')
-        op1 = jordan_wigner(ferm_op)
-        op2 = binary_code_transform(ferm_op, jordan_wigner_code(6))
-        self.assertEqual(op1.terms,op2.terms)
+        op2 = QubitOperator('Z0 Z1 Z2 Z3 Z4 X5',0.5)\
+              -QubitOperator('Z0 Z1 Z2 Z3 Z4 Y5',0.5j)
+        op1 = binary_code_transform(ferm_op, jordan_wigner_code(6))
+        self.assertTrue(op1.isclose(op2))
 
     def test_checksum_code(self):
         hamiltonian, gs_energy = lih_hamiltonian()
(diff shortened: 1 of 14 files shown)


src/openfermion/transforms/_code_transform_functions_test.py
```diff
@@ -36,6 +38,25 @@ def lih_hamiltonian():
...
+        qubit_op = binary_code_transform(ferm_op, parity_code(n_modes))
+        self.assertEqual(qubit_op.terms,
+                         {((1, 'Z'), (2, 'X'), (3, 'X'), (4, 'X')): 0.5,
+                          ((2, 'Y'), (3, 'X'), (4, 'X')): -0.5j})
+
+        ferm_op = FermionOperator('5^')
+        op1 = jordan_wigner(ferm_op)
+        op2 = binary_code_transform(ferm_op, jordan_wigner_code(6))
+        self.assertEqual(op1.terms,op2.terms)
+
```
kevinsung: Can you change these tests to conform to the following procedure?
1. Inititialize `op1` using `binary_code_transform`
2. Initialize `op2` as a the correct QubitOperator
3. `self.assertTrue(op1.isclose(op2))

This is clearer and more correct; your tests are actually testing for exact equality of floating point numbers, but we only want to check for approximate equality.

src/openfermion/transforms/_code_transform_functions_test.py
```diff
@@ -36,6 +38,26 @@ def lih_hamiltonian():
...
+        ferm_op = FermionOperator('2')
+        n_modes = 5
+        qubit_op = binary_code_transform(ferm_op, parity_code(n_modes))
+        self.assertEqual(qubit_op.terms,
+                         {((1, 'Z'), (2, 'X'),(3, 'X'),(4, 'X')): 0.5,
+                          ((2, 'Y'),(3, 'X'), (4, 'X')): 0.5j})
+        ferm_op = FermionOperator('2^')
+        n_modes = 5
+        qubit_op = binary_code_transform(ferm_op, parity_code(n_modes))
+        self.assertEqual(qubit_op.terms,
```
kevinsung: I meant these two tests too (instead of asserting something about `qubit_op.terms`, create a new QubitOperator called `correct_op` and do `self.assertTrue(qubit_op.isclose(correct_op))`.

conta877: ok I got a bunch of those in other tests I think, will change it all then.

kevinsung:
@babbush Perhaps in a situation like this, there really is no reason to click "Update branch" if all the changes can be merged automatically anyway. That would avoid this CLA bot from chiming in.

babbush:
yeah probably true. Its just so tempting to press the button when I'm looking at the PR already!

kevinsung:
@conta877 , thanks for taking care of this so quickly!

babbush:
@kevinsung @conta877 is this reviewed and ready to merge? Kevin feel free to merge stuff like this (I believe you have merge rights).

kevinsung:
Yes, this is ready. I would merge but I don't know how to override the CLA thing.

babbush:
The override is something only admin can do, I see. Just request that I merge when things like this are ready. I'll merge once travis is finished.

PR #257, opened 2018-03-14
Fix uccsd_singlet_paramsize?

kevinsung:
I'm a bit unsure about this one, but hopefully my variable names are clear enough that someone can follow my reasoning and correct me if I'm wrong.

Singlet implies Sz = 0, which means there are the same number of spin up as spin down electrons, which means the number of electrons is even, right?

Previously we had `n_single_amplitudes = n_occupied * n_virtual`, but since `n_occupied` and `n_virtual` refer to spatial orbitals, we need to double it because there are two spins, right?

Then, we had `n_double_amplitudes = n_single_amplitudes ** 2`. But we wouldn't want to count double excitations of the same electron, or for two electrons into the same orbital, etc., right?

If I'm right about this I think we may want to change `uccsd_singlet_operator` as well (I don't mind doing it).

### DIFF
--- a/src/openfermion/utils/_unitary_cc.py
+++ b/src/openfermion/utils/_unitary_cc.py
@@ -118,13 +118,9 @@ def uccsd_singlet_paramsize(n_qubits, n_electrons):
         Number of independent parameters for singlet UCCSD with a single
         reference.
     """
-    if n_electrons % 2 != 0:
-        raise ValueError('A singlet state must have an even number of '
-                         'electrons.')
-
     # Since the total spin S^2 is conserved, we work with spatial orbitals
     n_spatial_orbitals = n_qubits // 2
-    n_occupied = n_electrons // 2
+    n_occupied = int(numpy.ceil(n_electrons / 2))
     n_virtual = n_spatial_orbitals - n_occupied
 
     n_single_amplitudes = n_occupied * n_virtual
@@ -157,12 +153,8 @@ def uccsd_singlet_generator(packed_amplitudes, n_qubits, n_electrons):
         generator(FermionOperator): Generator of the UCCSD operator that
             builds the UCCSD wavefunction.
     """
-    if n_electrons % 2 != 0:
-        raise ValueError('A singlet state must have an even number of '
-                         'electrons.')
-
     n_spatial_orbitals = n_qubits // 2
-    n_occupied = n_electrons // 2
+    n_occupied = int(numpy.ceil(n_electrons / 2))
     n_virtual = n_spatial_orbitals - n_occupied
 
     # Unpack amplitudes
--- a/src/openfermion/utils/_unitary_cc.py
+++ b/src/openfermion/utils/_unitary_cc.py
@@ -130,8 +130,8 @@ def uccsd_singlet_paramsize(n_qubits, n_electrons):
     n_single_amplitudes = n_occupied * n_virtual
 
     # Below is equivalent to
-    #     n_single_amplitudes + (n_single_amplitudes choose 2)
-    n_double_amplitudes = n_single_amplitudes * (n_single_amplitudes + 1) // 2
+    #     2 * n_single_amplitudes + (n_single_amplitudes choose 2)
+    n_double_amplitudes = n_single_amplitudes * (n_single_amplitudes + 3) // 2
 
     return n_single_amplitudes + n_double_amplitudes
 
@@ -170,9 +170,9 @@ def uccsd_singlet_generator(packed_amplitudes, n_qubits, n_electrons):
     # Single amplitudes
     t1 = packed_amplitudes[:n_single_amplitudes]
     # Double amplitudes associated with one spatial occupied-virtual pair
-    t2_1 = packed_amplitudes[n_single_amplitudes:2 * n_single_amplitudes]
+    t2_1 = packed_amplitudes[n_single_amplitudes:3 * n_single_amplitudes]
     # Double amplitudes associated with two spatial occupied-virtual pairs
-    t2_2 = packed_amplitudes[2 * n_single_amplitudes:]
+    t2_2 = packed_amplitudes[3 * n_single_amplitudes:]
 
     # Initialize operator
     generator = FermionOperator()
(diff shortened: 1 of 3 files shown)


jarrodmcc:
I'll look more carefully at this tomorrow, but I just wanted to clarify
something definition wise first.  Singlet in this sense refers to an action
of the operator.  That is, singlet operators don't change the spin subspace
one is in, and the naming is related to how the operators transform with
respect to irreducible representations.  The practical implication of this
is that it is a symmetry restriction that forces excitations to match in a
certain way between the alpha and beta spaces of a given spin orbital.
That is why there are only n_occupied * n_virtual single excitations.  They
act symmetrically on the alpha and beta spaces of each of the spatial
orbitals, and working in the language of spatial orbitals is a little more
natural here.

Squaring the single amplitudes for the doubles was more of a programming
simplicity and convenience thing.  Eventually one would want take out the
redundant excitations, but this complicates indexing and doesn't provide a
complete fix regardless, as there are also certain symmetry disallowed
excitations you would want to remove.  If you want to change the indexing
for the double excitations to be more efficient, I think it would be great,
just make sure all the other routines that convert to and from it from
electronic structure packages match.

kevinsung:
Thanks Jarrod, that was very helpful. My mistake was that I was only concerned with preserving Sz, but the operator also needs to preserve S^2. I have updated the code; the only change now is in the counting of double excitations. I believe I removed all the symmetry-disallowed excitations as well, but I'm not sure.

Also, which other conversion routines are you talking about? Are you talking about something in the electronic structure plugins that might be affected by this?

jarrodmcc:
Sure, the tests seem to be failing at the moment though, as it looks like you changed the counting but not the way the tensor is accessed.  As far as the storage routines I meant, you might have to check on the routines like conversion in Psi4 (e.g. [here ](https://github.com/quantumlib/OpenFermion-Psi4/blob/master/openfermionpsi4/_psi4_conversion_functions.py) ) Alternatively, if storage is left the same, I think you'd want to modify the routines that allow one to use the stored amplitudes in convert_amplitude_format, otherwise one can't use stored data anymore

kevinsung:
Ok, I have rewritten the function for generating the UCCSD singlet generator. Also, I renamed the functions `uccsd_operator` and `uccsd_singlet_operator` to `uccd_generator` and `uccsd_singlet_generator`. Is that okay?

@jarrodmcc It appears that `uccsd_convert_amplitude_format` is supposed to be used with the function `uccsd_operator` and not `uccsd_singlet_operator`. The input parameters to `uccsd_operator` contain explicit information about the excitation operators (either as an index list or because the parameters are contained in a 2- (singles) or 4- (doubles) tensor).  However, the input parameters to `uccsd_singlet_operator` are in the form of a one-dimensional array `packed_amplitudes`, with the singles amplitudes preceding the doubles amplitudes, and the total length of the list equal to the appropriate output of `uccsd_singlet_paramsize`. The correspondence between the ordering of the list and the excitation operators is implicitly determined and is a little complex because only the minimal number of parameters required is used; I think minimizing the number of parameters is important and is one of the main reasons one would use singlet UCCSD instead of the general one.

It seems to me that currently there is no functionality anywhere for converting amplitudes for use with `uccsd_singlet_operator`, so I'm not sure that I need to change any other code. I could, however, write a function that takes in the singles and doubles amplitudes as an N x N and N x N x N x N tensor and output some `packed_amplitudes`, but this would only really make sense if the input tensors were guaranteed to have the right symmetries. As an example, for 4 spatial orbitals (8 spin-orbitals) and 4 electrons, the coefficients of "4^ 0 7^ 3", "6^ 2 5^ 1", "4^ 0 6^ 2", and "7^ 3 5^ 1" should all be the same.

My understanding is that the reason for having the connection with electronic structure plugins is that classical procedures can generate natural initial guesses for UCCSD (I don't know how this works though). Is the same true for _singlet_ UCCSD?

jarrodmcc:
The renaming seems fine.  You are correct that the packing is a little complicated and conversion routines are not in place yet.  The initial guesses provided by electronic structure codes in the case of restricted spin are usually what we would already call singlet UCCSD.  They are provided by spatial orbital indexing though, so it would still require conversion to our packed format if we wanted to use it there.  I think this pull request looks good overall now, but I will take a minute to go through the code momentarily.

kevinsung:
I realized that I missed one type of symmetry-allowed double excitation. I'm fixing it now.

src/openfermion/utils/_unitary_cc.py
```diff
@@ -117,17 +118,25 @@ def uccsd_singlet_paramsize(n_qubits, n_electrons):
         Number of independent parameters for singlet UCCSD with a single
         reference.
     """
-    n_occupied = int(numpy.ceil(n_electrons / 2.))
-    n_virtual = n_qubits // 2 - n_occupied
+    if n_electrons % 2 != 0:
```
jarrodmcc: This is actually not strictly a requirement, a singlet operator can be applied to a doublet state for example.  However I'm not sure how thoroughly we've tested those cases, so perhaps it would be okay to change the error message to a warning instead saying something like "untested for odd number of electrons"?

kevinsung: No, I realize now there should be no problem if the input is odd. I'll fix this.

jarrodmcc: Sounds good.  I realize the existing code would probably need to be generalized to allow application to triplet states, though I'm not sure about doublet states.  Maybe we can leave the error in for now and change the message to "not implemented".  What are your thoughts?

kevinsung: How should we handle an odd value of `n_qubits`? The old code seems inconsistent on this.

jarrodmcc: It should at least be a not implemented error, as an odd number of qubits is incompatible with any encoding we've specified  I think.  That would have needed to result from a manual truncation of qubits, which if done, should have been done after calling these routines I would think.

kevinsung:
Ok I've made the changes. I think this is ready.

PR #262, opened 2018-03-18
Change normalization and initialization of Majorana operators.

### LINKED ISSUE #258: Change normalization convention of Majorana operators
Right now our convention is (a^\dagger + a) / sqrt(2), but I think omitting the normalization factor is more common in the literature and also more natural from a coding point of view (the coefficient can be passed to the function). I'd like to change the definition of Majorana operators to omit the normalization factor by default. People cool with that? @babbush? Not sure if anyone actually cares about this...

kevinsung:
Resolves #258 .

Also switched the meaning of "Type 0" and "Type 1" operators, and allows initialization using strings, e.g. `majorana_operator('c3')` now means Type 0 operator on mode 3.


kevinsung:
You'll need to merge this one due to the CLA thing.

Issue #259, opened 2018-03-16
Which Bravyi-Kitaev transform?


kevinsung:
As discussed at #232 , our current implementation of the Bravyi-Kitaev transform comes from arXiv:1701.07072, and this definition is actually different from the one given by Bravyi and Kitaev in arXiv:quant-ph/0003137, which is the same as the one in arXiv:1208.5986. Apparently this fact had gone unnoticed until that discussion so I thought we should discuss now whether or not we should change our implementation. If we want to implement both, then I'm asking which one should be our "canonical" one and be assigned the function name `bravyi_kitaev`.

The two transforms actually correspond to different variants of the Fenwick tree data structure; the one in arXiv:quant-ph/0003137 is also called a binary indexed tree and it is the one that is actually used in practice because it can be implemented efficiently as an implicit data structure (elements are stored in an array and parent-child relationships computed using bitwise operations on the indices of the array).

One reason to change our implementation is that the one from quant-ph/0003137 can be performed faster. That's because the parity and update sets can be computed from the mode index using arithmetic (actually, bitwise) operations, rather than actually traversing a tree data structure.

Another nice property about the binary indexed tree version is that the form of the transformed operators depends less on the number of modes. For instance, the annihilation operator a_5 looks the same whether the total number of modes is, 9, 10, ..., 15, and then at 16 (the next power of 2) just one more X operator gets attached. On the other hand, with the current implementation, a_5 looks different for all these different total number of modes.

kevinsung:
Now that I think of it, if one of them gives slightly lower weight operators in the cases in which they differ, then we should probably go with that one. I haven't investigated that yet though.

msteudtner:
@kevinsung  Operator weight depends on the structure of the tree, right ? (the width and depth or however it is actually called)  For a number of modes that is 2^integer, Fenwick and binary BK seem the same - I think when you are in between the 2^integer and 2^(integer-1) in the binary BK, you basically delete the upper part (I mean thepart closer to the root)  of the tree. This clips the update operators (the part that is X or Y).  I guess comparing the shape of the trees for non-integer power cases is going to be an answer.  I assume the binary trees are going to be wider and not as deep  (have more Zs and less Xs) then the Fenwick trees. So what is better, a broader or higher tree ?

kevinsung:
Yes, a binary-indexed tree for non-powers of 2 can be thought of as taking the Fenwick tree from the next power of 2 and omitting the higher indices.

[...]

Issue #222, opened 2018-02-26
More demos and comprehensive description of features


babbush:
Pull request #221 starts a comprehensive list of features which we will merge once completed (help on that welcome, especially if making additions which explain your contribution). This should help raise awareness about the many functionalities currently implemented in OpenFermion.

However, the best way to show people a new feature is to write an example. I believe these are best demonstrated through ipython notebooks. Currently we have one huge demo but I think we should creating smaller, more specific demos. For instance, perhaps we should have a demo called "transforms" which discusses the many transforms currently implemented. Perhaps we should have a demo called "models" which discusses how to generate different Hamiltonian models. Any help in this regard would be appreciated.

@kevinsung @idk3 @kanavsetia @conta877

babbush:
We have a specific request from @Konjkov to add a demo for compiling phase estimation circuits.

conta877:
is there a plan to divide the huge one into smaller pieces?

babbush:
Not currently but there should be such a plan!

kevinsung:
I'm working on a transforms demo! See my work in progress here: https://github.com/kevinsung/OpenFermion/blob/transforms_demo/examples/transforms.ipynb. Comments welcome. I'm taking a very (too?) pedagogical approach.

kevinsung:
Ok I made my demo only about Jordan-Wigner and Bravyi-Kitaev. If others want to make a demo about Bravyi-Kitaev Superfast and/or the Verstraete-Cirac, please go ahead. It might make sense to put those in the same notebook.

mhucka:
Closing because this seems to be about as done as it's likely to be.

PR #263, opened 2018-03-19
Add demo for Jordan-Wigner and Bravyi-Kitaev transforms

kevinsung:
Partially addresses #222 . I think it makes sense to have the Jordan-Wigner and Bravyi-Kitaev transforms in their own demo notebook, with the other more exotic transforms in another one. I took a very pedagogical approach which I think will be helpful to people new to the topic. Of course, feedback on the presentation is welcome.

Some content might need to be changed depending on how #259 is resolved.


babbush:
@kevinsung I guess this needs to be merged after the eq PR?

kevinsung:
Indeed, and preferably after resolving #259 too. I understand #259 might need more input from people though and we don't need to wait.

babbush:
Is this good to go?

kevinsung:
Not yet, I'll make the necessary changes and let you know when.

kevinsung:
Ready once checks pass!

kevinsung:
Wait, I realized this and the binary transforms demo are not being tested, I'll fix that...

kevinsung:
Ok ready.

Issue #252, opened 2018-03-13
issues in tutorial examples


michaelkaicher:
In "Basic OpenFermion example" in the tutorial there is an issue with

`from openfermion.ops import hermitian_conjugated`
I think it should be 
`from openfermion.utils import hermitian_conjugated`

babbush:
Thanks for pointing this out Michael! Indeed, I don't think we've checked the code in the "examples" in a while. We do keep the "demo" up to date. I wonder if we shouldn't just remove the "tutorial" part of the docs and instead refer users to the jupyter notebook demo, which is both more comprehensive and more up-to-date. What do others think?

conta877:
I agree that having a single place for people to go to as a reference is a good idea. 
That can also be the push to create better demos.

kevinsung:
I agree that instead of the out-of-date examples section we should have a page with links to the demo notebooks on GitHub, at least for now.

Strilanc:
I just wrote an execute-documentation-snippets-and-expect-it-to-pass test for another project. It should do a good job of catching these kinds of mistakes, though it comes with its own downsides. Does that sound interesting?

kevinsung:
Sounds interesting to me! What are the downsides?

babbush:
As nice as this would be, I think that our demos are much more pedagogical and complete anyways. I think we should just refer people there as a rule and remove the text tutorial.

Strilanc:
The downside is that if you put expensive code into the documentation, your tests take forever. Also the doc code may touch the network, and other unfortunate things.

babbush:
Closed by #292

PR #265, opened 2018-03-20
Define __eq__ for SymbolicOperator and eliminate isclose

kevinsung:
This change was discussed at #233 . After this is merged, `op1.isclose(op2)` is no longer valid syntax and `op1 == op2` should be used instead, where `op1` and `op2` are SymbolicOperators such as FermionOperator or QubitOperator.

I did not define `__neq__` yet. I will do that once this and #233 are merged, then use EqualsTester to reformulate some tests for `__eq__` (which will also test `__neq__`).

### DIFF
--- a/examples/openfermion_demo.ipynb
+++ b/examples/openfermion_demo.ipynb
@@ -201,7 +201,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The print function prints each term in the operator on a different line. Note that the line my_operator = term_1 + term_2 creates a new object, which involves a copy of term_1 and term_2. The second block of code uses the inplace method +=, which is more efficient. This is especially important when trying to construct a very large FermionOperator. FermionOperators also support a wide range of builtins including, str(), repr(), *=, *, /, /=, +, +=, -, -=, - and **. Note that instead of supporting != and ==, we have the method .isclose(), since FermionOperators involve floats. We demonstrate some of these methods below."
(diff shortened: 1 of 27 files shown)


babbush:
Is this PR ready for review?

kevinsung:
Yes.

babbush:
I imagine this PR will break plugins like OpenFermion-ProjectQ and OpenFermion-Psi4. This means that we'll either need to bump the version on OpenFermion after this PR, or we'll need to leave __isclose__ in until our next version bump. What do you think?

kevinsung:
I think we should merge this now and bump the version soon. I'm willing to
update the plugins tomorrow (can't today).

babbush:
I think there is also a question about whether we want to implicitly "compress" when doing certain arithmetic operations. For instance, when adding two FermionOperators, should we just delete entries when we find that the absolute values of their coefficients are smaller than the tolerance?

babbush:
Anyways, I looked through the PR and it's great, thanks :+1:

However, I don't want to break the plugins for longer than about an hour. So I want to have the plugin PRs that would fix these changes ready-to-go before I merge. When all that is ready, we will merge, then bump version number. @jarrodmcc does this sound good to you?

@kevinsung are there other breaking changes we should make sure to incorporate before the next version bump?

kevinsung:
Ok that's fine with me, I'll probably get those PR's in tomorrow (or maybe
later today). The compress thing is a separate issue.

Have you looked at #259? It's best to resolve that before the next version.
I'm quite convinced that we should change our Bravyi-Kitaev implementation.

kevinsung:
We should also fix #252 before the next version (that's easy).

kevinsung:
@babbush it turns out that the only plugin that needs to be updated is OpenFermion-ProjectQ. It also turns out that that plugin is broken by the changes made at #257 and #267 . The tests that are broken are named at https://github.com/quantumlib/OpenFermion-ProjectQ/pull/42. I'll probably need some help to fix those.

kevinsung:
Given that OpenFermion-ProjectQ is already broken by the development version of OpenFermion, perhaps we can just merge this as is? It would make it easier to fix the plugins. It's not as if we're breaking the versions people download from pip.

babbush:
Okay

PR #267, opened 2018-03-20
Remove redundant double excitation from UCCSD singlet generator

kevinsung:
@jarrodmcc I messed up yesterday and added a redundant double excitation. I momentarily forgot that electrons are indistinguishable!

### DIFF
--- a/src/openfermion/utils/_unitary_cc.py
+++ b/src/openfermion/utils/_unitary_cc.py
@@ -188,8 +188,8 @@ def uccsd_singlet_generator(packed_amplitudes, n_qubits, n_electrons):
         virtual_down = down_index(virtual_spatial)
         occupied_up = up_index(occupied_spatial)
         occupied_down = down_index(occupied_spatial)
-        
-        # Generate single excitations 
+
+        # Generate single excitations
         coeff = t1[i]
         # Spin up excitation
         generator += FermionOperator((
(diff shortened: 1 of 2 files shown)


jarrodmcc:
This seems okay, but a thought I had earlier this morning was that if you added back in the index over spins that was in the original uccsd_singlet_operator you might be able to cut the length of the code in half and make it a little easier to understand what's going on.  What are your thoughts on that?

kevinsung:
I'm not sure we could make the code shorter because each set of independent double excitations does not depend on a spin and actually involves both spins. I think *not* explicitly iterating over spin makes it more clear that we are taking advantage of the total spin S^2 symmetry, and also has the advantage of corresponding more closely to the counting of the parameters.

kevinsung:
On second thought, I think you are right. Let me look into it further.

kevinsung:
Okay, @jarrodmcc I've made that change!

kevinsung:
Actually we only needed to loop over the indexing functions. The code is shorter; I hope it's clearer too (not completely sure).

kevinsung:
Ok I made the loop to be over the spin to be more clear. I think I'm really done here; let me know if you have other comments.

jarrodmcc:
The code looks good I think, thanks for the changes! I'll approve after the tests pass

PR #272, opened 2018-03-21
fix coefficients for compatibility with Psi4

kevinsung:
@jarrodmcc @babbush This fixes the ProjectQ UCCSD issue. The issue was that #257 did not double-count amplitudes like the old code did. After this is merged, https://github.com/quantumlib/OpenFermion-ProjectQ/pull/42 should fix all the issues with that plugin. We should probably think about what is the best convention though and how to make it clear to the user.


kevinsung:
Wait, don't merge this yet...

kevinsung:
Ok I think this is ready. Only one of the coefficients needed to be doubled. Here is an explanation:
In the old code, there was one loop for singles excitations and a separate loop for doubles excitations:
```
    # Define a compound space that is partitioned into occupied, virtual, spins
    # the spin component assumes alpha spins are even, beta spins are odd
    spaces = range(n_virtual), range(n_occupied), range(2)

    # Generate all spin-conserving single excitations between occupied-virtual
    for i, j, s in itertools.product(*spaces):
        uccsd_generator += FermionOperator((
            (2 * (i + n_occupied) + s, 1),
            (2 * j + s, 0)),
            t1[t1_ind(i, j)])

        uccsd_generator += FermionOperator((
            (2 * j + s, 1),
            (2 * (i + n_occupied) + s, 0)),
            -t1[t1_ind(i, j)])

    # Generate all spin-conserving double excitations between occupied-virtual
    for i, j, s, i2, j2, s2 in itertools.product(*spaces, repeat=2):
        uccsd_generator += FermionOperator((
            (2 * (i + n_occupied) + s, 1),
            (2 * j + s, 0),
            (2 * (i2 + n_occupied) + s2, 1),
            (2 * j2 + s2, 0)),
            t2[t2_ind(i, j, i2, j2)])

        uccsd_generator += FermionOperator((
            (2 * j2 + s2, 1),
            (2 * (i2 + n_occupied) + s2, 0),
            (2 * j + s, 1),
            (2 * (i + n_occupied) + s, 0)),
            -t2[t2_ind(i, j, i2, j2)])
```
In the new code, the loop for the doubles excitations goes through a different, smaller space due to spin symmetry:
```
    # Generate all spin-conserving double excitations derived
    # from two spatial occupied-virtual pairs
    for i, ((p, q), (r, s)) in enumerate(
            itertools.combinations(
                itertools.product(range(n_virtual), range(n_occupied)),
                2)):
```
However, some of the doubles excitations (those derived from one spatial occupied-virtual pair) were moved to the loop that also handled singles excitations, and furthermore they are only counted once. In the old code they would have been counted twice, once when the loop was at (i, j, k, l) and another time when the loop was at (k, l, i, j), as these doubles excitations are equivalent.

kevinsung:
There must be a bug in coveralls because I did not add any uncovered lines!

jarrodmcc:
I think I'm confused, but I feel like the code has converged back to what it had originally but slightly longer.  I thought the objective of the modification was to remove redundant excitations of the form
a_i^\dagger a_j a^i\dagger a_j which contribute 0 due to repetition of the indices and distinctness between the set of virtuals and occupied, but it seems we have the same set as originally.  Also, it seems the double excitations are counted twice in a sort of unconnected part in the singles and connected part in the doubles?  The convention on the doubles should always be 2, and I'm not sure we should mix the singles and doubles in different loops for clarity, which might be causing some of the counting/coefficient discrepancy.  Maybe we can just step through this in detail later.

kevinsung:
One way to see that the new code is different is that the second loop accesses a fewer number of parameters from `packed_amplitudes`. The original loop accessed `(n_occupied * n_virtual) ** 2` different indices (the number of different `i, j, i2, j2`), while the new loop accessed
`(n_occupied * n_virtual) choose 2` different indices (the number of different `(p, q), (r, s)`). In addition to not looping over the zero excitations you mentioned such as `0^ 1 0^ 1` (which means don't loop over `(p, q), (p, q)`), I also chose the following behavior: if we loop over `(p, q), (r, s)`, then don't loop over `(r, s), (p, q)`. Due to symmetry considerations (specifically, conservation of total spin S^2), that would be redundant too.

If I had simply removed the zero excitations (and not the additional redundant ones) then it would have reduced the number of parameters to `(n_occupied * n_virtual) ** 2 - (n_occupied * n_virtual)`, which is still larger than `(n_occupied * n_virtual) choose 2`. I realize now that this whole business was probably premature optimization haha.

The doubles excitations that I added in the singles excitation loop have the form `p^ q r^ s` where `p` and `r` are in the same spatial orbital and `q` and `s` are in the same spatial orbital. For instance, `2^ 0 3^ 1`. These are not counted in the second loop.

My main motivation for these changes was to have the fewest number of parameters possible but again that was probably premature optimization and if this makes things more confusing I won't insist that we keep the changes.

kevinsung:
I have rewritten the loop parts in a way that is less redundant and more closely matches the original code. I think it's more transparent now how I reduced the number of parameters: basically I used `itertools.combinations` instead of `itertools.product` to generate most of the doubles excitations. However, that means that I had to move the doubles excitations associated with a single occupied-virtual pair elsewhere, and I think it made sense to just put them in the same loop as with the singles. I also rewrote that loop to put an inner loop over spin, and now I don't need to double any coefficients.

To check this, I added a test (at Jarrod's suggestion) to compute the classical H2 CCSD energy using only the excitations relevant for singlet; for that purpose I wrote a conversion function `uccsd_singlet_get_packed_amplitudes`. The energies match.

Since H2 only has 2 spatial orbitals, this test does not test the code that generates most of the doubles amplitudes. In fact, that part of the code for the regular UCCSD generator function `uccsd_generator`, which I did not modify, was also not tested. So I added tests for LiH as well. However, these tests do not pass, even for the regular UCCSD generator function (which I did not modify).
For the regular UCCSD, the energies are
`-7.880968276932223` vs the correct `-7.88098231`.
For the singlet UCCSD (new code), the energies are
`-7.877374111269659` vs the correct `-7.88098231`.
I guess it's not good that the singlet energy does not match the regular energy, but I think we should only expect it to if the classical CCSD amplitudes are guaranteed to have the correct symmetries, since my conversion routine ignores many of the amplitudes.

jarrodmcc:
It depends what you mean by match the regular energy here.  Singlet UCCSD should always match UCCSD under constraints that make the UCCSD operator a singlet (this will mean some of the amplitudes are the same, which we remove as a redundancy in the singlet generator).  It should not always match the FCI or CCSD energy unless there are only 2 particles in the system.  A caveat here is that if you ran LiH with a frozen core, there would be only 2 active electrons, and it should be exact for both, but you have to make that constraint compatible everywhere (same calculation method, same active space etc)

kevinsung:
LiH has 4 electrons and 12 qubits (6 spatial orbitals). Does frozen core mean
`occupied_indices = range(2)` and `active_indices = range(2, 12)`?
If I do that then `get_active_space_integrals` throws an error.

jarrodmcc:
I believe the active space routines are based on spatial orbital indexing.
Frozen core would mean freezing the first spatial (or first two spin)
orbitals.  To enable that in Psi4, you'll need to change that option in the
template file that says freeze_core false to freeze_core true. We could
eventually make that part of a suite of optional parameters to pass to
electronic structure packages.

kevinsung:
Hmm I think adding a new molecule computation to the library might be outside the scope of the PR and also a bit beyond my expertise at this point. I ask that you accept my changes as they stand, especially since it's now tested even more than the old code was. Perhaps a future PR which adds more electronic structure features can add more tests for this.

PR #275, opened 2018-03-22
Add method to compute many-body order of SymbolicOperator

kevinsung:


src/openfermion/utils/_operator_utils.py
```diff
@@ -32,6 +32,27 @@ class OperatorUtilsError(Exception):
     pass
 
 
+def weight(symbolic_operator):
```
babbush: should we perhaps call this many_body_order? Or k_locality? Or something like that?

kevinsung: Also, should it be a method? Currently SymbolicOperator has one lonely method, `induced_norm`.

kevinsung: Those are okay but I still like "weight" the best. I think "many_body_order" is a bit jargon-y and "k_locality" might possibly be interpreted as some kind of geometric locality.

babbush: I am inclined to think so.

babbush: I agree with you about making it a method.

But I disagree with you about the name; "weight" is not a particularly common (or precise/clear) name for this property. It might be jargon, but "many_body_order" is the correct jargon. I agree that "k-locality" could be confused with a geometric meaning.

kevinsung: Yeah, I think you're right about "many_body_order" being correct and more in line with the physics literature. I like "weight" because it's the same as the Hamming weight of the bitstring of length `n_modes` where you have a `1` if it's acted on and `0` otherwise.

kevinsung: But now that I think of it, I probably just made that up. We can go with "many_body_order".

babbush: haha. Yes, weight makes sense in the context of a state but it is a little too ambiguous in the context of an operator, in my view.

kevinsung: Ryan I want to make one final case for the term "weight". In fact, this term is used in the quantum information literature (I knew I wasn't just making it up). For instance, Nielsen and Chuang defines the weight of Pauli strings in the error correction section, and there the connection with Hamming weight is natural. Some recent papers like https://arxiv.org/abs/1609.00735 also use the term. It's true that they define it for single operator strings, but I think the generalization to sums of operators as the maximum weight of a term is natural.

Another thing is that SymbolicOperator is not explicitly or necessarily a physics concept. I think its abstract ... er ... symbolic ... nature makes the connection with strings and Hamming weight most natural. I think the name "many_body_order" might be better for a function which can also accept a PolynomialTensor as an input.

babbush: Sorry but I really don't like "weight". I think it is too ambiguous. It sounds too much like a type of norm to me. Look up any paper on perturbative gadgets, which spend the whole paper trying trying to reduce the many body order, and they either use this term or "k-locality".

babbush: the many body order of a PolynomialTensor is hard-coded by construction. I can't imagine a case when somebody doesn't know the many-body order of their PolynomialTensor.

kevinsung: We already use properties for PolynomialTensor and its subclasses.

babbush: Just for one attribute (in PolynomialTensor), which really is an attribute that isn't computable from the rest of the data. There is a good reason for making it a property there. I don't see one in this case.

kevinsung: Okay I've changed the name and made it a method.

babbush:
Please resolve conflicts

PR #276, opened 2018-03-22
Don't "raise" warning

kevinsung:
I messed up; "raising" a warning causes the program to terminate and the correct way to do it is to use the `warnings` module.


kevinsung:
Great suggestion, I did it!

idk3:
Looks great! This is even more comprehensive than I was imagining ;) we should try to maintain this level across all warnings we use.

kevinsung:
Haha, I simply copied it straight from the documentation of `warnings`!

PR #279, opened 2018-03-23
Use EQ_TOLERANCE in equality comparison in JW for InteractionOperator

kevinsung:
Also removed an inline import statement.


kevinsung:
Actually I found a bug so I'm gonna fix that and submit a separate PR.

PR #282, opened 2018-03-23
Rewrite random_interaction_operator and allow complex option

kevinsung:
I'm investigating a bug and found it useful to understand this function. I couldn't so I rewrote it and added an option to generate a complex operator, and tested it as well.

However, I'm a little confused about the exposition of symmetries in InteractionOperator from the OpenFermion paper, specifically the `h_{pqrs} = h_{sqrp} = ...`. Are these symmetries derived only from the hermiticity of the Hamiltonian? If so, then I don't see why, for instance, `h_{pqrs} = h_{prqs}`; if r and q are different I think these should be the negative of each other. I think the symmetries implied from my code are correct.

### DIFF
--- a/src/openfermion/utils/_testing_utils.py
+++ b/src/openfermion/utils/_testing_utils.py
@@ -76,56 +76,50 @@ def random_quadratic_hamiltonian(n_qubits,
                                 antisymmetric_mat, chemical_potential)
 
 
-def random_interaction_operator(n_qubits):
+def random_interaction_operator(n_qubits, real=True):
     """Generate a random instance of InteractionOperator."""
+    if real:
+        dtype = float
+    else:
+        dtype = complex
 
-    # Initialize.
+    # The constant has to be real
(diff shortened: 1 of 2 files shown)


jarrodmcc:
Without looking specifically at that function, the integral symmetries come
from the interaction operator as well as the assumption that the basis
functions are real.  A more detailed exposition can be found here:
http://vergil.chemistry.gatech.edu/notes/permsymm/permsymm.pdf
which talks about the reduced symmetry in the complex case as well

kevinsung:
Thanks for the reference Jarrod, I think that will be helpful. I wrote my code only on the assumption that the output should be Hermitian. Perhaps that's not exactly what we want.

kevinsung:
I don't understand why coveralls is saying coverage descreased even though I added 0 untested lines.

babbush:
who knows? I'll merge it.

PR #283, opened 2018-03-23
Change implementation of Bravyi-Kitaev

### LINKED ISSUE #259: Which Bravyi-Kitaev transform?

kevinsung:
Changes the implementation of the Bravyi-Kitaev transform from the one described in arXiv:1701.07072 to the original one described in arXiv:quant-ph/0003137, which is the same as the one described in arXiv:1208.5986.

Keeps the old implementation under the name `bravyi_kitaev_tree`.

Resolves #259 .

This change is good for four reasons:
1. The new implementation results in QubitOperators with (in many cases strictly) smaller many-body order (this was verified numerically for all systems up to size 1000 but not proven).
2. The new implementation is faster (uses bitwise operations to compute parity and update sets instead of traversing a literal tree data structure).
3. The new implementation has simpler and shorter code (no need for a separate file to store a classical data structure).
4. In the new implementation, the indices that a given annihilation operator acts on is less dependent on the total number of modes.

I realize this might take some time to review and I'm happy to answer any questions. I plan to eventually write something up explaining the subtleties here.


babbush:
I don't think we should delete the Fenwick tree implementation. I think we should keep both. Perhaps we can call the new code Bravyi_Kitaev and call the old code "Fenwick_Tree" or something like that.

kevinsung:
What do you think of having the "Fenwick tree" implementation available only through the binary code transforms functionality?

babbush:
We already have nice code for it. I suggest we keep what we have. Some might find it useful.

kevinsung:
Ok I propose the name `bravyi_kitaev_tree` in recognition of the literal tree data structure it uses and that Fenwick never actually proposed to implement his data structure that way (his original data structure uses bitwise operations like in the original Bravyi-Kitaev and my implementation here).

babbush:
that's fine

kevinsung:
Ok I've done that. I'll do think this is ugly though because the test files for the two transforms are identical, and in fact much of the code in the two implementations is identical too. I think the ideal way to make the old version available is only as a BinaryCode (but I won't do that now).

kevinsung:
Something's definitely wrong with coveralls... now it's penalizing me for increasing coverage!

src/openfermion/transforms/_bravyi_kitaev.py
```diff
@@ -11,15 +11,13 @@
...
 """Bravyi-Kitaev transform on fermionic operators."""
-from __future__ import absolute_import
 
 from openfermion.ops import QubitOperator
-from openfermion.transforms._fenwick_tree import FenwickTree
+from openfermion.utils import count_qubits
 
 
 def bravyi_kitaev(operator, n_qubits=None):
     """Apply the Bravyi-Kitaev transform and return qubit operator.
```
babbush: Especially given all the confusion we've had, please give more description of which BK transform and ideally provide references to both Seeley and Bravyi+Kitaev

babbush:
Are you actually increasing test coverage though? It seems like the error might be in the "+" sign?

src/openfermion/transforms/_bravyi_kitaev.py
kevinsung: Done.

PR #286, opened 2018-03-24
Fix Jordan Wigner for InteractionOperator

### LINKED ISSUE #281: Bug in Jordan Wigner for InteractionOperators
Executing
```
from openfermion import *
from openfermion.utils._testing_utils import random_interaction_operator

iop = random_interaction_operator(5)
op1 = jordan_wigner(iop)
op2 = jordan_wigner(get_fermion_operator(iop))

assert op1 == op2
```
results in an AssertionError. That's a bug, right?

I suspect there's something wrong with how the indices are handled in `jordan_wigner_interaction_op`. A consistency test like this is actually not present in `transforms/_jordan_wigner_test.py`, though there are similar ones for the helper functions `jordan_wigner_one_body` and `jordan_wigner_two_body`.

kevinsung:
Fixes #281 .

Instead of finding the bug I just rewrote the code. It's also now about 4x faster than the old (incorrect) code.

One thing I noticed and documented is that based on the way the helper functions `jordan_wigner_one_body` and `jordan_wigner_two_body` are written, this only supports InteractionOperators with real entries. Maybe a future PR can expand support to complex ones.

### DIFF
--- a/src/openfermion/transforms/_bksf.py
+++ b/src/openfermion/transforms/_bksf.py
@@ -110,8 +110,9 @@ def bravyi_kitaev_fast_interaction_op(iop):
                                 continue
                         # Handle case of 3 unique indices
                         elif len(set([p, q, r, s])) == 3:
-                            transformed_term = two_body(edge_matrix_indices,p, q, r, s)
-                            transformed_term *= coefficient/2.
+                            transformed_term = two_body(edge_matrix_indices,
+                                                        p, q, r, s)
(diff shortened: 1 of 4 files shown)


kevinsung:
This cannot be merged until #287 is resolved.

babbush:
Why are we waiting on #287 ?

babbush:
Will this PR be ready if it passes?

PR #294, opened 2018-03-27
Move prune_unused_indices, freeze_orbitals, and linearize_decoder

### LINKED ISSUE #271: Reorganization
I think some reorganization is in order:
- I think `ops/__init__.py` should only import classes, and that the following moves make sense anyway:
-- `general_basis_change`, `prune_unused_indices`, `freeze_orbitals`, `normal_ordered` -> `utils/_operator_utils.py`.
-- `linearize_decoder` -> `transforms/_binary_codes.py`
- I think `transforms/_conversion.py` actually belongs in `utils`. It is the only file in `transforms` that is not actually concerned with fermionic transforms in the usual sense. If we put it in `utils` then we can also move the functions `jordan_wigner_sparse` and `qubit_operator_sparse` into it.

If people agree with these changes it would be best to make them before the next version release, and any help would be appreciated.

kevinsung:
Much less ambitious refactoring than requested at #271 . I decided I don't think I care enough to do the rest anyway so let's say this fixes #271.

@max-radin @conta877 just letting you know I moved some of your functions!

### DIFF
--- a/src/openfermion/ops/__init__.py
+++ b/src/openfermion/ops/__init__.py
@@ -13,14 +13,11 @@
 from ._binary_polynomial import BinaryPolynomial
 from ._polynomial_tensor import PolynomialTensor, general_basis_change
 from ._quadratic_hamiltonian import QuadraticHamiltonian
-from ._symbolic_operator import (SymbolicOperator,
-                                 prune_unused_indices)
+from ._symbolic_operator import SymbolicOperator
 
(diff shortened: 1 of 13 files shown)


kevinsung:
Again, coveralls is acting up... it says
" 87 of 87 new or added lines in 7 files covered. (100.0%) "

kevinsung:
Also, I believe ready ready for a version upgrade after this PR.

PR #297, opened 2018-03-29
fixed bug in travis config

babbush:
This should cause travis to fail until we've fixed the future dependency in setup. This should prevent such problems from occurring in the future. Once we've fixed the setup script we should merge this and then bump the version again (unfortunately).


babbush:
What I did is that I removed the explicit install of the requirements and made travis actually install openfermion. But it was installing numpy/scipy/matplotlib twice and took longer to run. Hopefully this time around it is a little faster.

PR #300, opened 2018-04-01
Add method to QuadraticHamiltonian for obtaining diagonalizing circuit

kevinsung:
- Added the `diagonalizing_circuit` method to QuadraticHamiltonian.
- Changed the order of arguments in the `__init__` of QuadraticHamiltonian and made passing in the constant optional.
- Modified the `diagonalizing_bogoliubov_transform` method of QuadraticHamiltonian to return an NxN matrix instead of a (2N) x (2N) matrix in the case that the Hamiltonian conserves particle number.

Example: if `one_body` is the coefficient matrix for a one-body Hamiltonian, then
```
quad_ham = QuadraticHamiltonian(one_body)
circuit = quad_ham.diagonalizing_circuit()
```
gives you a circuit that performs the basis change from the occupation number basis to the basis in which `one_body` is diagonal using Givens rotations. This circuit is different from the ones returned by the Slater determinants routines which take advantage of additional symmetries that only apply during state preparation. The circuit returned here is a general basis transformation that can be used anywhere in an algorithm.

In order to avoid circular dependencies, this required me to move much of the Givens rotations helper code from `utils/_slater_determinants.py` to `ops/_quadratic_hamiltonian.py` which is why this PR looks big.


src/openfermion/transforms/_conversion.py
```diff
@@ -263,11 +263,12 @@ def get_quadratic_hamiltonian(fermion_operator,
...
         # Hamiltonian conserves particle number
         quadratic_hamiltonian = QuadraticHamiltonian(
-            constant, hermitian_part, chemical_potential=chemical_potential)
+            hermitian_part, constant=constant,
+            chemical_potential=chemical_potential)
     else:
         # Hamiltonian does not conserve particle number
         quadratic_hamiltonian = QuadraticHamiltonian(
-            constant, hermitian_part, antisymmetric_part, chemical_potential)
+            hermitian_part, antisymmetric_part, constant, chemical_potential)
```
babbush: kind of a trivial comment, but shouldn't you call it "hermitian_part" and "anti_hermitian_part".

babbush:
Nevermind - I am missing something. It was keeping changes to QuadraticHamiltonian collapsed. I'll have another look...

src/openfermion/ops/_quadratic_hamiltonian.py
```diff
@@ -253,36 +254,108 @@ def diagonalizing_bogoliubov_transform(self):
...
-        # Get the orthogonal transformation that puts majorana_matrix
-        # into canonical form
-        canonical, orthogonal = antisymmetric_canonical_form(majorana_matrix)
+            # Compute the unitary and return
+            diagonalizing_unitary = majorana_basis_change.T.conj().dot(
+                orthogonal.dot(majorana_basis_change))
+
+            return diagonalizing_unitary
+
+    def diagonalizing_circuit(self):
```
babbush: I'm somewhat thinking that this should actually be a function that is called on the class as opposed to a method. One could imagine wanting to use some of this functionality without initializing the class. Also, the single responsibility principle is a general design pattern that we try to follow.

src/openfermion/transforms/_conversion.py
kevinsung: No, "anti-hermitian" means A = A^\dagger but this matrix satisfies A = A^T.

src/openfermion/ops/_quadratic_hamiltonian.py
kevinsung: I just tried to be consistent with what we already have, which seems to be something like "methods are okay if they are really specific to the class and can't be called on other classes." For instance, SymbolicOperator, BinaryPolynomial, and PolynomialTensor all have their own methods, and QuadraticHamiltonian already has methods like `majorana_form` and `diagonalizing_bogoliubov_transform` which I think should be treated the same way.

kevinsung:
I think what I did here is the most consistent with what we already have, and if we want to make this a function instead of a method then we should do the same with the other methods of QuadraticHamiltonian at the very least, and probably with all the other methods that we have in our classes (`induced_norm`, `many_body_order`, etc.). I don't feel very strongly either way.

src/openfermion/ops/_quadratic_hamiltonian.py
kevinsung: Also, the new method `diagonalizing_circuit` depends on the method `diagonalizing_bogoliubov_transform`, which in the case of non-particle-number-conservation depends on `majorana_form`, so it seemed to make most sense to require a QuadraticHamiltonian to be instantiated, especially if we want to have one function that works for both particle-number-conserving and not.

kevinsung:
It definitely makes sense to have the Givens rotations routines collected in a separate file though; I'll do that.

Issue #301, opened 2018-04-03
Does PolynomialTensor mistakenly implement __neq__ instead of __ne__?


singular-value:
Hey there, Pranav here. I'm hoping to get involved with OpenFermion, so I started taking a look at https://github.com/quantumlib/OpenFermion/issues/274 which seems like a good first issue (thanks @kevinsung and @idk3 for the clear description and background discussion).

I was reading through the code for `PolynomialTensor` and noticed that it implements a `__neq__` method. Should this method be `__ne__`? (https://docs.python.org/2/reference/datamodel.html#object.__ne__ and
https://docs.python.org/3/reference/datamodel.html#object.__ne__)

If so, I can submit a pull request.

kevinsung:
Yes, you are correct. It would be great if you can fix that and test it better, perhaps using the EqualsTester utility in `utils/_testing_utils.py`.

kevinsung:
Note that for the EqualsTester tests to pass you should set `__hash__` to `None` for PolynomialTensor. That's a change that should be made anyway.

singular-value:
Resolved by pull request #303

babbush:
@singular-value thanks for this contribution. If you'd like to be on the OpenFermion release paper, please email me at [my github name]@google.com and I can send you the link to our overleaf.

PR #303, opened 2018-04-06
Generalize equality and arithmetic operations for PolynomialTensor to allow differing keys

### LINKED ISSUE #274: Generalize __add__ and __eq__ methods of PolynomialTensor
As discussed in #240, we should be able to add and compare PolynomialTensors with different sets of keys (if the key is not present then it counts as the zero tensor). Currently, attempting these operations raises an error.

singular-value:
Resolves #274 and #301 .

### DIFF
--- a/src/openfermion/ops/_polynomial_tensor.py
+++ b/src/openfermion/ops/_polynomial_tensor.py
@@ -196,8 +196,8 @@ class PolynomialTensor(object):
 
         return diff < EQ_TOLERANCE
 
-    def __neq__(self, other_operator):
-        return not (self == other_operator)
+    def __ne__(self, other):
+        return not (self == other)
 
     def __iadd__(self, addend):
         if not issubclass(type(addend), PolynomialTensor):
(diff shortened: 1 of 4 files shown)


singular-value:
I signed it!

kevinsung:
Thanks for the great PR! Ideally you would have tested this using EqualsTester (it would have caught the `__hash__` thing) but I won't require it. The tests for this class are kind of messy so that would be something of a hassle.

@babbush something is definitely up with Coveralls because it says "47 of 47 new or added lines in 2 files covered." I think after the one comment I made is addressed this should be good to go.

babbush:
@kevinsung I agree. I am not sure what the problem is. Let me know once you feel the PR is good to go and I will override coveralls if necessary.

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -173,22 +173,34 @@ def __setitem__(self, args, value):
...
             diff = max(diff, discrepancy)
+
         return diff < EQ_TOLERANCE
 
-    def __neq__(self, other_operator):
-        return not (self == other_operator)
+    def __ne__(self, other):
+        return not (self == other)
+
+    def __hash__(self):
```
kevinsung: Sorry, I should have been more clear. It's `__hash__ = None`.

babbush:
coveralls appears to have gotten its act together.

singular-value:
Awesome!

PR #305, opened 2018-04-09
fix bug in is_hermitian for FermionOperators

kevinsung:
It was giving false negatives due to not normal-ordering.

### DIFF
--- a/src/openfermion/utils/_operator_utils.py
+++ b/src/openfermion/utils/_operator_utils.py
@@ -25,7 +25,8 @@ from openfermion.hamiltonians._jellium import (grid_indices,
                                                orbital_id,
                                                position_vector)
 from openfermion.ops import (FermionOperator, InteractionOperator,
-                             InteractionRDM, PolynomialTensor, QubitOperator)
+                             InteractionRDM, PolynomialTensor, QubitOperator,
+                             normal_ordered)
 from scipy.sparse import spmatrix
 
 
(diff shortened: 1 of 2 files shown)


idk3:
@babbush @kevinsung this was originally left out intentionally, to avoid possible surprises with the cost of normal ordering. The idea was that `normal_order` should only get called when the user explicitly calls it. In the FermiLib days we even discussed normal ordering by default with a flag to disable it. Why are we changing it now? Also, if we're doing it for `is_hermitian`, why not for other cases like `isclose`?

kevinsung:
I think @idk3 has a good point about being consistent with `isclose`. I actually think we should change both to always normal order beforehand because I can't think of a situation where true equality or hermiticity is not what you want to know. Evaluating these predicates is an intrinsically expensive task, but I think in practice it may not be so bad. We should think about the expense of evaluating these predicates in the same way that we think about the expense of normal-ordering.

babbush:
I may have been hasty in merging this. I am actually very worried about routines that implicitly normal order. Normal ordering can be very expensive. Our users would be quite surprised to learn that is_hermitian now has exponential runtime in general. Perhaps we should rethink our implementation.

kevinsung:
Wait... to check Hermiticity, can we not perform the JWT and check if the resulting QubitOperator is Hermitian?

babbush:
We could, but that also has exponential runtime in general.

kevinsung:
Which part, the JWT or checking Hermiticity of a QubitOperator? It seems to me that neither is exponential.

babbush:
Perform the JWT on a single FermionOperator that has N raising/lowering operators. That maps to an exponential number of terms. This is also related to what can happen when you try to normal order such terms.

kevinsung:
Oh, of course. Dang. Okay, well I'm fine with reverting the changes in this PR.

idk3:
Perhaps this can be an opportunity to revisit the original decision :)

In principle, I'm actually fine with normal ordering for these checks regardless of the possibly-exponential cost. The big thing for me is that the worst case for normal ordering is a fairly contrived one - it isn't one that comes up in any natural Hamiltonian or series of operations. Second, to my knowledge our use cases limit us to 3-body operators - these come up in the double commutators, but  so far we haven't gone further than that. So normal ordering for us is 1) far from the worst case, and furthermore 2) restricted to FermionOperators on 6 modes, and because of this may not be so bad after all. Thoughts?

kevinsung:
Yes, I'd rather keep this change too, because I believe
- `is_hermitian` should always return True for Hermitian operators.
- Checking whether an operator is Hermitian is intrinsically costly.
- In practice, it's probably not that bad.

PR #306, opened 2018-04-09
Create DiagonalCoulombHamiltonian class

### LINKED ISSUE #246: Diagonal interaction Hamiltonian class
Many of our recent papers including [arXiv:1706.00023](https://arxiv.org/abs/1706.00023) and [arXiv:1711.04789](https://arxiv.org/abs/1711.04789) have focused on simulating the electronic structure Hamiltonian in a basis where the Coulomb operator is diagonal. Thus, the Hamiltonian can be expressed as:
\sum_{pq} T_{pq} a^\dagger_p a_q + \sum_{pq} V_{pq} n_p n_q
where I have subsumed the usual "U_p n_p" term in the diagonal part of T_{pq} a^\dagger_p a_q.

Anyways, this is clearly a very essential and important form of the electronic structure Hamiltonian. As we build more routines for compiling to chips it will be essential to use this form. Likewise, this form can be stored with less memory and mapped to qubits almost instantly. Thus, we should have a class specialized for this purpose.

@idk3 @kevinsung @jarrodmcc thoughts?

kevinsung:
Includes conversion routines to/from FermionOperator and an optimized Jordan-Wigner Transform.

I'm going to say this resolves #246 ; if we want more features we can create a new issue.


kevinsung:
Reverting the changes made in #305 (seems likely to happen) will resolve the conflicts.

babbush:
please resolve conflicts. @kevinsung your last few PRs have been out of sync with the repo.

kevinsung:
The conflict is only due to the change made in #305 so if we revert that first then this will go through.

src/openfermion/transforms/_jordan_wigner.py
```diff
@@ -61,9 +65,45 @@ def jordan_wigner(operator):
                     z_factors + ((ladder_operator[0], 'Y'),), 0.5j)
             transformed_term *= pauli_x_component + pauli_y_component
         transformed_operator += transformed_term
+
     return transformed_operator
 
 
+def jordan_wigner_diagonal_coulomb_hamiltonian(operator):
```
babbush: Perhaps this should be called by the standard jordan_wigner function when it is passed a DiagonalCoulombHamiltonian. That way one does not need to keep track of multiple different jordan_wigner functions.

src/openfermion/ops/__init__.py
```diff
@@ -11,6 +11,7 @@
 #   limitations under the License.
 
 from ._binary_polynomial import BinaryPolynomial
+from ._diagonal_coulomb_hamiltonian import DiagonalCoulombHamiltonian
```
babbush: I am not sure I like "DiagonalCoulombHamiltonian" because it seems to exclude other operators that might not have any explicit connection to Coulomb operators or Hamiltonians, which also fit this form. Sorry if I agreed to this name in the past...

what about DiagonalTwoBodyOperator?

babbush: Maybe this name is okay. If you like, let's just keep it.

src/openfermion/ops/_diagonal_coulomb_hamiltonian.py
```diff
@@ -0,0 +1,39 @@
...
+    where
+
+        - :math:`T` is a Hermitian matrix.
+        - :math:`V` is a real symmetric matrix.
+
+    Attributes:
+        one_body(ndarray): The Hermitian matrix :math:`T`.
+        two_body(ndarray): The real symmetric matrix :math:`V`.
+        constant(float): The constant.
+        n_qubits(int): The number of qubits.
```
babbush: why does n_qubits need to be an attribute? I don't necessarily like that because our other classes like FermionOperator and QubitOperator don't have this attribute and n_qubits is easily computed from the dimension of one_body or two_body.

src/openfermion/transforms/_jordan_wigner.py
kevinsung: I did do it that way. However, this reminds me that I should also update `get_sparse_operator`.

src/openfermion/ops/_diagonal_coulomb_hamiltonian.py
kevinsung: I only did it because PolynomialTensor has this attribute and this class is more similar to that than to FermionOperator and QubitOperator. Do you think we should remove it from PolynomialTensor too?

kevinsung: Ok I removed this attribute from DiagonalCoulombHamiltonian.

kevinsung:
Ok I've addressed these comments and also slightly modified the conversion routine from FermionOperator to check that the two-body coefficients are real and discard small imaginary parts.

src/openfermion/ops/__init__.py
kevinsung: I had the same thought, but I think DiagonalTwoBodyOperator sounds too much like it doesn't contain one-body terms.

PR #307, opened 2018-04-11
Fix PolynomialTensor multiply, extend to floats, and allow dividing by numbers

kevinsung:
- Allows multiplying and dividing PolynomialTensors and DiagonalCoulombHamiltonians by numbers
- Fixed incorrect behavior in multiplication by a PolynomialTensor with different key (a missing key should count as the zero tensor; previous behavior was the same as adding zeros instead of multiplying zeros).

### DIFF
--- a/src/openfermion/ops/_polynomial_tensor.py
+++ b/src/openfermion/ops/_polynomial_tensor.py
@@ -277,6 +277,11 @@ class PolynomialTensor(object):
         product *= multiplier
         return product
 
+    def __rmul__(self, multiplier):
+        product = copy.deepcopy(self)
+        product *= multiplier
+        return product
+
     def __itruediv__(self, dividend):
         if isinstance(dividend, (int, float, complex)):
             for key in self.n_body_tensors:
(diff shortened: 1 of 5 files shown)


kevinsung:
Hold on, I forgot I wanted to do this for DiagonalCoulombHamiltonian too...

src/openfermion/ops/_polynomial_tensor.py
```diff
@@ -250,26 +250,47 @@ def __sub__(self, subtrahend):
         return r
 
     def __imul__(self, multiplier):
-        if not issubclass(type(multiplier), PolynomialTensor):
+        if isinstance(multiplier, (int, float, complex)):
+            for key in self.n_body_tensors:
+                self.n_body_tensors[key] *= multiplier
+
+        elif isinstance(multiplier, PolynomialTensor):
```
singular-value: ah, thanks for this.

kevinsung: No problem, I should have caught it the first time around.

PR #310, opened 2018-04-14
Marginal maps

ncrubin:
RDM mapping functions with extensive testing
    
This PR provides functionality to convert marginals in various forms to one another.  All inputs attempt to obey the tensor ordering convention used throughout openfermion.
    
In order to have quality tests we've also added 2-hole-marginals, particle-hole-marginals, and the code that generated them.
    
I am completely open to moving the functions somewhere else in the package if everyone thinks the `measurements` submodule is an inappropriate place for the methods.


src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Mapping RDMs to other RDMs"""
+import numpy
+from itertools import product
+
+
+def krond(i, j):
+    return 1 if i == j else 0
```
idk3: `float(i == j)` or `int(i == j)` would be cleaner here - I'd slightly prefer the direct float if you don't care otherwise.

src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Mapping RDMs to other RDMs"""
+import numpy
+from itertools import product
+
+
+def krond(i, j):
```
idk3: `kron_delta` would be 100% clear

src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+    Args:
+        tpdm (numpy.ndarray): The 2-RDM as a 4-index tensor. Indices follow the
+            internal convention of tpdm[p, q, r, s] ==
+            :math:`a_{p}^{\dagger}a_{q}^{\dagger}a_{r}a_{s}`.
+        particle_number (float): number of particles in the system
+
+    Returns:
+        opdm (numpy.ndarray): The 1-RDM contracted from the tpdm.
+    """
+    return numpy.einsum('prrq', tpdm) / (particle_number - 1)
```
idk3: I don't think we can reasonably expect `particle_number` to be a float, cast `(particle_number - 1)` explicitly

src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+
+def tqdm_to_oqdm(tqdm, hole_number):
+    """
+    Map from 2-hole-RDM to 1-hole-RDM
+
+    Args:
+        tqdm (numpy.ndarray): The 2-hole-RDM as a 4-index tensor. Indices
+            follow the internal convention of tqdm[p, q, r, s] ==
+            :math:`a_{p}a_{q}a_{r}^{\dagger}a_{s}^{\dagger}`.
+        hole_number (float):
```
idk3: Say what `hole_number` is (yeah it's pretty obvious), also the float cast should be done for `(hole_number - 1)` to be safe

src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+            :math:`a_{p}^{\dagger}a_{q}^{\dagger}a_{r}a_{s}`.
+        opdm (numpy.ndarray): The 1-RDM as a 2-index tensor. Indices follow the
+            internal convention of opdm[p, q] ==
+            :math:`a_{p}^{\dagger}a_{q}`.
+
+    Returns:
+        phdm (numpy.ndarray): The particle-hole matrix.
+    """
+    ldim = opdm.shape[0]
+    tgdm = numpy.zeros_like(tpdm)
```
idk3: Why is this `tdgm` instead of `phdm`?

src/openfermion/measurements/_rdm_mapping_functions.py
```diff
@@ -0,0 +1,200 @@
...
+            Indices follow the internal convention of phdm[p, q, r, s] ==
+            :math:`a_{p}^{\dagger}a_{q}a_{r}^{\dagger}a_{s}`.
+        num_particles: number of particles in the system.
+        num_basis_functions: number of spin-orbitals
+            (usually the number of qubits)
+
+    Returns:
+        opdm (numpy.ndarray): the 1-RDM transformed from a 1-RDM.
+    """
+    return numpy.einsum('prrq', phdm) / (num_basis_functions -
```
idk3: Cast divisor to float

src/openfermion/measurements/_rdm_mapping_functions_test.py
```diff
@@ -0,0 +1,187 @@
...
+            THIS_DIRECTORY, 'tests/tqdm_H1-Li1_sto-3g_singlet_1.45.hdf5')
+        with h5py.File(tqdm_lih_sto3g, 'r') as fid:
+            self.tqdm_lih_sto3g = fid['tqdm'][...]
+
+        phdm_lih_sto3g = os.path.join(
+            THIS_DIRECTORY, 'tests/phdm_H1-Li1_sto-3g_singlet_1.45.hdf5')
+        with h5py.File(phdm_lih_sto3g, 'r') as fid:
+            self.phdm_lih_sto3g = fid['phdm'][...]
+
+    def test_krond(self):
```
idk3: Add a test like `krond(3, 3) == 1`, and split these up a little

src/openfermion/measurements/_rdm_mapping_functions_test.py
```diff
@@ -0,0 +1,187 @@
...
+        for file in filter(lambda x: x.endswith(".hdf5"),
+                           os.listdir(DATA_DIRECTORY)):
+            molecule = MolecularData(
+                filename=os.path.join(DATA_DIRECTORY, file))
+            if molecule.fci_one_rdm is not None:
+                true_oqdm = numpy.eye(molecule.n_qubits) - molecule.fci_one_rdm
+                test_opdm = oqdm_to_opdm(true_oqdm)
+                assert numpy.allclose(test_opdm, molecule.fci_one_rdm)
+
+    def test_tqdm_conversions(self):
```
idk3: This test has 9 asserts - please break them up / make them smaller. Also, it would be safer not to overwrite the builtin `file` - it looks like you can use `filename` instead everywhere it appears.

src/openfermion/measurements/__init__.py
```diff
@@ -17,3 +17,8 @@
 
 from ._rdm_equality_constraints import (one_body_fermion_constraints,
                                         two_body_fermion_constraints)
+
+from ._rdm_mapping_functions import (krond, tpdm_to_tqdm, tpdm_to_opdm,
```
idk3: I think some of these names might be a bit terse, `opdm` and `tpdm` are clear enough IMO but I don't know what `oqdm`, `tqdm`, or `phdm` are. This might be clearer to someone who works with RDMs more, so maybe this isn't needed - @babbush / @jarrodmcc ?

src/openfermion/measurements/_rdm_mapping_functions_test.py
```diff
@@ -0,0 +1,187 @@
...
+
+        true_oqdm = numpy.eye(molecule.n_qubits) - molecule.fci_one_rdm
+        test_oqdm = tqdm_to_oqdm(true_tqdm,
+                                 molecule.n_qubits - molecule.n_electrons)
+        assert numpy.allclose(true_oqdm, test_oqdm)
+
+        test_tpdm = tqdm_to_tpdm(true_tqdm, molecule.fci_one_rdm)
+        assert numpy.allclose(test_tpdm, molecule.fci_two_rdm)
+
+    def test_phdm_conversions(self):
```
idk3: This test also has 9 asserts and redefines the builtin `file`.

src/openfermion/measurements/__init__.py
babbush: I agree with Ian that these names are too terse.

src/openfermion/measurements/_rdm_mapping_functions.py
babbush: kronecker_delta would be even more clear

src/openfermion/measurements/_rdm_mapping_functions.py
babbush: Indeed, in many cases the user will provide an int which would make this function perform int division.

src/openfermion/measurements/_rdm_mapping_functions_test.py
babbush: Indeed, it is much easier to identify the problem when these are separate tests.

src/openfermion/measurements/_rdm_mapping_functions.py
kevinsung: I think the best way to deal with division is to add `from __future__ import division` at the top of the file. Then we can just use Python 3 syntax.

ncrubin: 👍

ncrubin:
@idk3 thanks for the review!

babbush:
My feeling is that measurements is not the best folder. Perhaps in transforms or utils would be more appropriate?

ncrubin:
I will move the methods into `utils`.  `utils` is becoming the catch-all for anything outside of representing fermions as qubits, operators and their algebras, and hamiltonian utilities.  We might want to brainstorm on how to expand the 5 main modules.

src/openfermion/tests/_rdm_generation.py
```diff
@@ -0,0 +1,101 @@
...
+            sparse_fermi_op = get_sparse_operator(fermi_op,
+                                                  n_qubits=molecule.n_qubits)
+            true_phdm[p, q, r, s] = numpy.conj(state).T.dot(
+                sparse_fermi_op.dot(state)).real
+            true_phdm[s, r, q, p] = true_phdm[p, q, r, s]
+
+    return true_phdm
+
+
+if __name__ == "__main__":
```
babbush: What is this? We run tests using unittest or pytest modules. Please see other test files for an example. No modules in OpenFermion have a "main".

ncrubin: It's sample code to generate marginals in a very naive way.  It's mostly an explanation of where the `tpdm_*.hdf5` files come from.  I'd be happy to remove it or move the file to the `examples/` folder in the top level directory.

babbush: Please either add it to a demo ipython notebook in the examples folder, or remove it.

ncrubin: removed

babbush:
LGTM

Issue #122, opened 2017-12-11
Saving and loading operators


kevinsung:
Currently, the `save_operator` function in `utils/_operator_utils.py` saves Fermion and QubitOperators using the data serialization module `marshal`. I think it's desirable to save these objects in a human-readable format, like JSON. What do people think? This would best be implemented after resolving issue #43.

Also, I'm wondering why `marshal` was chosen rather than `pickle` or `cPickle`.

babbush:
We actually had a large discussion about this back in the early days of FermiLib. See
https://github.com/ProjectQ-Framework/FermiLib/issues/91

The long story short is that pickle has security issues and also is less efficient. Myself and some collaborators are working on a project that involves computation of FermionOperators that are tens of gigabytes so this became a real issue for us. However, I think there are also some concerns about marshal that were never really resolved.

Looping @Spaceenter into the conversation since he implemented marshal for us.

idk3:
Direct text representations are also possible and easy to implement. We could allow for different modes in our save_to_file function that save in different formats - thoughts? Would that address things for you @kevinsung ?

kevinsung:
Yeah, I think adding an option to the `save_operator` to save it in JSON format would be good.

Spaceenter:
I think the reason we chose marshal over JSON a few months ago is performance:

1. I think we originally implemented using JSON, but @idk3 found it uses too much memory, and switching to marshal reduced the memory usage by an order of magnitude.

2. Another issue was that JSON requires every key (and value?) must be string, so we end up having lots of conversions from number to string and string to number, which is not efficient. 

That being said, I agree JSON is more readable for very small operators when performance is not a concern. However, if you want human-readable format, is the __str__ method of the operators sufficient? @kevinsung

idk3:
One thing is that we have str but can't convert from string. A possible
nice solution would be expanding init to allow longer input strings (as is
only the term operators are possible), and adding "as string" modes for
load/save (which look at a file). I think we could do the first without the
second but it's nice to have it encapsulated.

kevinsung:
@Spaceenter Yeah the `str` method prints it out in a nice format. I like @idk3 's suggestion to expand `init` to allow longer strings, rather than just a single term.

babbush:
Yes, that is a great suggestion. I think we should open a new issue about that functionality.

babbush:
The string initialization functionality has now been expanded. Accordingly, if anyone wants to work on this, go ahead!

[...]

PR #311, opened 2018-04-14
Added human-readable file IO for SymbolicOperators

quid256:
Proposed resolution to #122

### DIFF
--- a/src/openfermion/utils/_operator_utils.py
+++ b/src/openfermion/utils/_operator_utils.py
@@ -396,13 +396,15 @@ def inverse_fourier_transform(hamiltonian, grid, spinless):
                                      vec_func_2=momentum_vector)
 
 
-def load_operator(file_name=None, data_directory=None):
+def load_operator(file_name=None, data_directory=None, human_readable=False):
     """Load FermionOperator or QubitOperator from file.
 
     Args:
         file_name: The name of the saved file.
         data_directory: Optional data directory to change from default data
                         directory specified in config file.
+        human_readable: Whether the operator should be loaded from a
+                        human-readable format
 
     Returns:
         operator: The stored FermionOperator or QubitOperator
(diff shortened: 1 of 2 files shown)


quid256:
I signed it!

src/openfermion/utils/_operator_utils.py
```diff
@@ -396,13 +396,15 @@ def inverse_fourier_transform(hamiltonian, grid, spinless):
                                      vec_func_2=momentum_vector)
 
 
-def load_operator(file_name=None, data_directory=None):
+def load_operator(file_name=None, data_directory=None, human_readable=False):
```
kevinsung: Instead of `human_readable` let's call this `plain_text`.

src/openfermion/utils/_operator_utils.py
```diff
@@ -396,13 +396,15 @@ def inverse_fourier_transform(hamiltonian, grid, spinless):
...
     """Load FermionOperator or QubitOperator from file.
 
     Args:
         file_name: The name of the saved file.
         data_directory: Optional data directory to change from default data
                         directory specified in config file.
+        human_readable: Whether the operator should be loaded from a
+                        human-readable format
```
kevinsung: "Whether the input file is plain text."

src/openfermion/utils/_operator_utils_test.py
```diff
@@ -333,12 +333,27 @@ def test_save_and_load_fermion_operators(self):
                          msg=str(self.fermion_operator -
                                  loaded_fermion_operator))
 
+    def test_save_and_load_fermion_operators_readably(self):
+        save_operator(self.fermion_operator, self.file_name,
+                      human_readable=True)
+        loaded_fermion_operator = load_operator(self.file_name,
+                                                human_readable=True)
+        self.assertEqual(self.fermion_operator.terms,
+                         loaded_fermion_operator.terms)
```
kevinsung: `self.assertEqual(self.fermion_operator, loaded_fermion_operator)`

src/openfermion/utils/_operator_utils.py
```diff
@@ -462,10 +478,14 @@ def save_operator(operator, file_name=None, data_directory=None,
     else:
         raise TypeError('Operator of invalid type.')
 
-    tm = operator.terms
-    with open(file_path, 'wb') as f:
-        marshal.dump((operator_type, dict(zip(tm.keys(),
-                                              map(complex, tm.values())))), f)
+    if human_readable:
+        with open(file_path, 'w') as f:
+            f.write(operator_type + ": " + str(operator))
```
kevinsung: I think we'll want to do `:\n` instead of `: `. Can you add a test with small operator that explicitly includes what the saved format looks like? I.e., the test should compare the saved file with an explicit string.

src/openfermion/utils/_operator_utils.py
```diff
@@ -412,27 +414,39 @@ def load_operator(file_name=None, data_directory=None):
...
+        if operator_type == 'FermionOperator':
+            operator = FermionOperator()
+            for term in operator_terms:
+                operator += FermionOperator(term, operator_terms[term])
+        elif operator_type == 'QubitOperator':
+            operator = QubitOperator()
+            for term in operator_terms:
+                operator += QubitOperator(term, operator_terms[term])
+        else:
+            raise TypeError('Operator of invalid type.')
```
kevinsung: It should not be necessary to add terms individually. It should be possible to simply do something like `operator = FermionOperator(some_string)` where `some_string` is probably going to be straight from the file, minus the initial heading which says what type it is.

src/openfermion/utils/_operator_utils_test.py
quid256: Should the rest of the test cases be comparing the original operator and the loaded operator in this way? Including the cases that have already been written?

src/openfermion/utils/_operator_utils.py
quid256: I think this code is constructing the operators as you suggest. It uses the `split` function only to separate the type of the operator from the remaining data, which is stored as a string when the data is loaded from plain text

src/openfermion/utils/_operator_utils_test.py
kevinsung: That would be nice but I won't require it. Those tests were actually written before this was possible; we didn't always implement `__eq__` in SymbolicOperator.

src/openfermion/utils/_operator_utils.py
kevinsung: Ah, I was looking at the wrong part of the code. Yes, you did do it that way. Great!

kevinsung:
Thanks for the contribution! Feel free to add your name and affiliation to the NOTICE and README, if you like.

babbush:
... in the alphabetical section, please!

PR #313, opened 2018-04-18
General cells

jarrodmcc:
- Adds general supercells to grid class
- Fixes non-hermitian issue in one-body potential while being consistent for even and odd grids between plane wave and dual by re-defining the one-body potential to have the dual as the base case and FT defining the plane wave case
- Fixes some preconditions for success of certain routines and disables use of background constant in cases where it doesn't apply.


jarrodmcc:
The pull request is on the large side, as a lot of things had to change in concert to have intermediate stages that aren't broken.  Understandable if it takes a while to get to it.  Probably still needs additional tests for non-uniform cells and the constant term needs to be generalized to arbitrary unit cells as well, but that's disabled by exception for now.

src/openfermion/hamiltonians/_jellium.py
```diff
@@ -405,12 +291,14 @@ def jordan_wigner_dual_basis_jellium(grid, spinless=False,
         grid (Grid): The discretization to use.
         spinless (bool): Whether to use the spinless model or not.
         include_constant (bool): Whether to include the Madelung constant.
+            Note: constant is unsupported for non-uniform, non-cubic cells with
```
babbush: I think that this indentation and writing "Note:" is going to confuse the automatic docs. Because it is likely going to think now that "Note" is an argument that is supposed to be passed to this function. Perhaps you can just remove the colon and write on this line "note that the constant is not..."

src/openfermion/hamiltonians/_jellium.py
```diff
@@ -268,12 +147,14 @@ def dual_basis_jellium_model(grid, spinless=False,
         kinetic (bool): Whether to include kinetic terms.
         potential (bool): Whether to include potential terms.
         include_constant (bool): Whether to include the Madelung constant.
+            Note: constant is unsupported for non-uniform, non-cubic cells with
```
babbush: See my comment below: I think that the way you've written this note will confuse the sphinx documentation.

src/openfermion/hamiltonians/_plane_wave_hamiltonian.py
```diff
@@ -58,6 +57,13 @@ def wigner_seitz_length_scale(wigner_seitz_radius, n_particles, dimension):
 def dual_basis_external_potential(grid, geometry, spinless):
     """Return the external potential in the dual basis of arXiv:1706.00023.
 
+    The external potential resulting from electrons interacting with nuclei
+        in the plane wave dual basis.  Note that a cos term is used which is
+        strictly only equivalent under aliasing in odd grids, and amounts
+        to the addition of an extra term to make the diagonals real on even
+        grids.  This approximation is not expected to be significant and allows
+        for use of even and odd grids on an even footing.
```
babbush: would it be fair to say an "odd but even footing"?

src/openfermion/utils/_grid.py
```diff
@@ -13,80 +13,285 @@
...
+import scipy
+import scipy.linalg
+
+
+# Exceptions.
+class OrbitalSpecificationError(Exception):
+    pass
 
 
 class Grid:
```
babbush: Very good idea building out this Grid class to encapsulate some of the tricky reciprocal lattice stuff.

src/openfermion/utils/_low_depth_trotter_error.py
```diff
@@ -15,9 +15,8 @@
 from future.utils import iteritems, itervalues
 
 import numpy
+import openfermion.hamiltonians
 
-from openfermion.config import *
-from openfermion.hamiltonians import jellium_model, wigner_seitz_length_scale
```
babbush: I'm just curious why you prefer "import openfermion.hamiltonians" instead of from openfermion.hamiltonians import blah? I understand not liking the wildcard import but the other direct imports are convenient and make the code a bit cleaner. I'm not telling you to change anything, just wondering why you made this change.

src/openfermion/utils/_low_depth_trotter_error.py
```diff
@@ -350,11 +349,14 @@ def dual_basis_jellium_hamiltonian(grid_length, dimension=3,
...
 
     # Compute appropriate length scale.
-    length_scale = wigner_seitz_length_scale(
+    length_scale = openfermion.hamiltonians.wigner_seitz_length_scale(
         wigner_seitz_radius, n_particles, dimension)
 
     grid = Grid(dimension, grid_length, length_scale)
-    hamiltonian = jellium_model(grid, spinless=spinless, plane_wave=False)
+    hamiltonian = (
+        openfermion.hamiltonians.jellium_model(grid,
```
babbush: see, now you have ugly lines like this as a result of the import.

src/openfermion/utils/_low_depth_trotter_error.py
jarrodmcc: I'll just document my response here for future reference, but the import from config was removed due to not being actually used.  My IDE shows when imports aren't used and actually we have a fair number in OpenFermion that are not.  I think people are not removing imports as functions are moved around so we just have an ever growing list of them in some files.

The from removal was due to the fact that we have some fragile circular dependency problems in OpenFermion that can be exacerbated by the use of "from x import y" due to the fact that this executes the module in question as opposed to import.  Supposedly these problems are mostly fixed in Python3, but since we support 2, this was the easiest fix.  See more here:

https://stackoverflow.com/questions/744373/circular-or-cyclic-imports-in-python
and the many other stack overflow posts it links to

PR #315, opened 2018-04-24
Projection method and small fix to controlled trotterization

obriente:
Two small things here:
1) In a previous update I'd made a mistake in my definition of a 'C-Phase' gate in the Trotterization method, this is now fixed.
2) I've made a small method to project a qubit operator onto the Z=+/-1 sector of an individual qubit. I'm not sure if this already exists somewhere in OpenFermion (given that it was done in arXiv:1512.06860), but I couldn't find it.


src/openfermion/transforms/_projection.py
```diff
@@ -0,0 +1,61 @@
...
+        # qubits has an expectation value of zero
+        if [t for t in term if t[0] in qubits
+                and t[1] in ['X', 'Y']]:
+            error += abs(factor)**2
+            continue
+
+        new_term = tuple((t[0]-len([q for q in qubits 
+                                    if q < t[0]]), t[1])
+                         for t in term if t[0] not in qubits)
+        new_factor =\
```
babbush: How does the linter feel about this?

src/openfermion/transforms/__init__.py
```diff
@@ -34,3 +34,4 @@
 from ._jordan_wigner import jordan_wigner
 from ._reverse_jordan_wigner import reverse_jordan_wigner
 from ._verstraete_cirac import verstraete_cirac_2d_square
+from ._projection import project_onto_sector
```
babbush: make imports alphabetically unless you run into a circular dependency.

src/openfermion/transforms/_projection.py
```diff
@@ -0,0 +1,61 @@
...
+
+'''
+Functions to reduce the number of qubits involved
+in modelling a given system
+'''
+from openfermion.ops import QubitOperator
+import numpy
+
+
+def project_onto_sector(operator, qubits, sectors):
```
babbush: isn't this almost identical to the routines for freezing orbitals in FermionOperator. Should we think about having this routine work for both FermionOperator and QubitOperator?

src/openfermion/transforms/_projection.py
```diff
@@ -0,0 +1,61 @@
...
+        operator: the QubitOperator to work on
+        qubits: a list of indices of qubits in
+           operator to remove
+        sectors: for each qubit, whether to project
+            into the 0 subspace (<Z>=1) or the
+            1 subspace (<Z>=-1).
+
+    Returns:
+        projected_operator: the resultant operator
+        error: the trace norm of the removed term.
```
babbush: why is this error returned automatically? It doesn't seem like a crucial thing.

obriente: Fair point, perhaps we should flag for it? Error provides an upper bound to the error in estimating the eigenvalue of O by the eigenvalue of the projected operator, so I think providing the capability to calculate it is useful, but perhaps returning it always isn't the best idea.

src/openfermion/transforms/_projection.py
obriente: Ah, thanks for pointing the freezing routine out, I hadn't been able to find it when I looked before.

The routines should be effectively identical, so combining them shouldn't be an issue (although I think we'll need different routines for both classes). But regarding terminology - to me it feels a bit odd talking about 'freezing orbitals' for a qubit system. What do you think? If you think it's fine I'll change it, otherwise perhaps we can find a name that seems obvious regardless of the type of operator.

As a general statement: given a Hermitian operator O with two eigenspaces, one can write any other operator A=B+C where [B,O]=0 and {C,O}=0, and then diagonalise B on either eigenspace of O, with error in the eigenvalues (or at least, the lowest eigenvalue) bounded by the Frobenius norm of C. If someone wanted to work out the code to do this it might be useful, but this sounds like a bit of a challenge.

src/openfermion/transforms/_projection.py
obriente: The linter didn't complain, but it doesn't read that well. Rewritten.

src/openfermion/transforms/__init__.py
obriente: Thanks for finding that.

src/openfermion/transforms/_projection.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def project_onto_sector(operator, qubits, sectors):
+    '''
+    Takes a QubitOperator, and projects out a list
+    of qubits, into either the +1 or -1 sector.
+    Note - this requires knowledge of which sector
+    we wish to project into.
+
+    Input:
```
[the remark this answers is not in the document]
obriente: Fixed

src/openfermion/transforms/_projection.py
obriente: Splitting the code into a second function is easy enough, so that works well.

babbush: Great. I think the doc strings on the first function are out of date now though.

obriente: Fixed now.

PR #317, opened 2018-04-26
Style fixes in utils

idk3:


src/openfermion/utils/_grid_test.py
```diff
@@ -103,40 +103,41 @@ def test_momentum_vector(self):
...
-        g2 = Grid(dimensions=3, length=(3,3,3), scale=scale_matrix)
+        g2 = Grid(dimensions=3, length=(3, 3, 3), scale=scale_matrix)
         self.assertEqual(g1, g2)
 
-    def test_preconditions(self):
-        nan = float('nan')
-
+    def test_no_errors_in_call(self):
         # No exception
-        _ = Grid(dimensions=1, length=1, scale=1.0)
```
kevinsung: I think it's actually better to leave these. For instance, common configurations of pylint (including mine) would complain about pointless statements if we make your changes.

kevinsung:
Thanks @idk3 ! I can review this tomorrow. As it so happens, I was just doing a similar thing for `ops` (PR #318 ) so maybe you can review that one? Funny enough, I made some changes there which are the exact opposite of what you did here (the one I mentioned in the comment about pointless statements.

idk3:
and good thing we didn't start on the same part :o

src/openfermion/utils/_rdm_mapping_functions.py
```diff
@@ -32,7 +32,7 @@ def map_two_pdm_to_one_pdm(tpdm, particle_number):
     Returns:
         opdm (numpy.ndarray): The 1-RDM contracted from the tpdm.
     """
-    return numpy.einsum('prrq', tpdm) / (particle_number - 1)
+    return numpy.einsum('prrq', tpdm) / float(particle_number - 1)
```
kevinsung: Don't cast in these situations. It's unnecessary because `/` always means floating point division when `division` has been imported from `__future__`.

src/openfermion/utils/_rdm_mapping_functions.py
```diff
@@ -108,7 +108,7 @@ def map_two_hole_dm_to_one_hole_dm(tqdm, hole_number):
     Returns:
         oqdm (numpy.ndarray): The 1-hole-RDM contracted from the tqdm.
     """
-    return numpy.einsum('prrq', tqdm) / (hole_number - 1)
+    return numpy.einsum('prrq', tqdm) / float(hole_number - 1)
```
kevinsung: Same comment as above.

src/openfermion/utils/_rdm_mapping_functions.py
```diff
@@ -206,5 +206,5 @@ def map_particle_hole_dm_to_one_pdm(phdm, num_particles, num_basis_functions):
     Returns:
         opdm (numpy.ndarray): the 1-RDM transformed from a 1-RDM.
     """
-    return numpy.einsum('prrq', phdm) / (num_basis_functions -
-                                         num_particles + 1)
+    return numpy.einsum('prrq', phdm) / float(
```
kevinsung: See above.

src/openfermion/utils/_slater_determinants.py
```diff
@@ -19,7 +19,7 @@
 from openfermion.config import EQ_TOLERANCE
 from openfermion.ops import QuadraticHamiltonian
 from openfermion.ops._givens_rotations import (
-        fermionic_gaussian_decomposition, givens_decomposition)
+    fermionic_gaussian_decomposition, givens_decomposition)
```
kevinsung: Are you using non-default settings for pycodestyle? Mine does not complain about having an extra four spaces in these situations, and I actually think  the extra space helps with readability sometimes.

idk3: This is the (optional) PEP8 E126, continuation line over-indented for hanging indent - I prefer only doing the extra spaces when it would otherwise raise E129, visually indented line with same indent as next logical line. I do it with my own PRs (I find it a little harder to read sometimes) but don't mind in other people's, I guess.

babbush:
Is this ready to merge?

PR #318, opened 2018-04-26
Fix docs errors and lint in ops

kevinsung:
- Fixed an error in the documentation of `QuadraticHamiltonian.bogoliubov_transform`.
- Removed lint caught by `pycodestyle` and my `pylint` in `ops`.

### DIFF
--- a/src/openfermion/ops/_givens_rotations_test.py
+++ b/src/openfermion/ops/_givens_rotations_test.py
@@ -54,7 +54,7 @@ class GivensMatrixElementsTest(unittest.TestCase):
         """Test bad input."""
         with self.assertRaises(ValueError):
             v = numpy.random.randn(2)
-            G = givens_matrix_elements(v[0], v[1], which='a')
+            _ = givens_matrix_elements(v[0], v[1], which='a')
 
 
 class GivensRotateTest(unittest.TestCase):
@@ -171,7 +171,7 @@ class GivensDecompositionTest(unittest.TestCase):
         Q = Q[:m, :n]
 
         with self.assertRaises(ValueError):
-            givens_rotations, V, diagonal = givens_decomposition(Q)
+            _, _, _ = givens_decomposition(Q)
 
     def test_identity(self):
         n = 3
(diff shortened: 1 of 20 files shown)


src/openfermion/ops/_givens_rotations_test.py
```diff
@@ -171,7 +171,7 @@ def test_bad_dimensions(self):
         Q = Q[:m, :n]
 
         with self.assertRaises(ValueError):
-            givens_rotations, V, diagonal = givens_decomposition(Q)
+            _, _, _ = givens_decomposition(Q)
```
idk3: You don't need to unpack for these tests, `_ = ...` should be enough

src/openfermion/ops/_givens_rotations_test.py
```diff
@@ -322,12 +322,10 @@ def test_bad_dimensions(self):
...
-                fermionic_gaussian_decomposition(rand_mat))
+            _, _, _ = fermionic_gaussian_decomposition(rand_mat)
 
     def test_bad_constraints(self):
         n = 3
         ones_mat = numpy.ones((n, 2 * n))
         with self.assertRaises(ValueError):
-            decomposition, left_unitary, antidiagonal = (
-                fermionic_gaussian_decomposition(ones_mat))
+            _, _, _ = fermionic_gaussian_decomposition(ones_mat)
```
idk3: `_ = ...` should be fine (also on L325)

src/openfermion/ops/_interaction_rdm_test.py
```diff
@@ -13,9 +13,10 @@
 """Tests for interaction_rdms.py."""
 from __future__ import absolute_import
 
+import os
```
idk3: `os` should be after `unittest` here I suspect

kevinsung: They are both standard library modules so they should be placed in alphabetical order.

kevinsung:
@idk3 I staled your review by removing unnecessary imports of `absolute_import`.

PR #320, opened 2018-04-26
moving standardized jellium Hamiltonian; associated changes

### LINKED ISSUE #316: Jellium model generation
I asked @idk3 about generating Jellium models and he pointed me to the function `dual_basis_jellium_hamiltonian` in `utils/_low_depth_trotter_error.py`. Doesn't this belong in `hamiltonians/_jellium.py`?

idk3:
Fixes #316


idk3:
I don't really like the name I've given it, feedback on that appreciated! I want to highlight that it does all the lifting normally associated with `jellium_model` or `dual_basis_jellium_model` for you, but don't have good ideas for what to call it.

src/openfermion/hamiltonians/_jellium.py
```diff
@@ -15,7 +15,42 @@
 
 import numpy
 
-from openfermion.ops import FermionOperator, QubitOperator
+from openfermion.ops import FermionOperator, normal_ordered, QubitOperator
+from openfermion.utils._grid import Grid
+
+
+def wigner_seitz_length_scale(wigner_seitz_radius, n_particles, dimension):
```
kevinsung: You should delete this from the file it was moved from.

src/openfermion/hamiltonians/_jellium_test.py
```diff
@@ -22,6 +22,51 @@
 from openfermion.utils import count_qubits, eigenspectrum, Grid, is_hermitian
 
 
+class WignerSeitzRadiusTest(unittest.TestCase):
```
kevinsung: You took these from `_plane_wave_hamiltonian_test.py`, right? Delete them there.

kevinsung:
Like @babbush , I'm not sure I agree with the premise of this function. It seems more appropriate as a function you'd define in your own code for your own convenience.

I actually think that ideally, the "boilerplate" you're trying to avoid with this function go into a Jupyter notebook example on generating model Hamiltonians. That would be very helpful for explaining what is going on here and it would be more transparent than hiding the inner workings in a convenience function.

src/openfermion/hamiltonians/_jellium.py
idk3: oops, should've noticed the lack of red

babbush:
I guess I am okay with this. I'd ask @kevinsung to take a closer look and approve though. He seemed to have some comment about deleting tests that I didn't understand.

idk3:
It was more that I forgot to delete some tests while transferring code :P
This function straddles the boundary for a convenience function IMO - it's
fairly simple, but still possible to make small surprising mistakes with
it. I don't think demo code should be code that we intend to have users
basically copy-paste for their own use, but there are good examples we
could give with this function together with say the Fourier transform code
(perhaps better explaining Grid, the Wigner-Seitz radius, etc).

babbush:
Is it ready to be merged or do you still need to address that test issue?

idk3:
I've addressed it, but I'm still not totally happy with the function name.
If nobody has a better one it's good to merge though.

kevinsung:
I think we should definitely change the name. This function seems to be about setting the proper length scale given an electron density, so I think the name should reflect that. Furthermore, I think it should not normal order and compress the Hamiltonian before returning.

kevinsung:
I think it's better to have this function have one purpose, and it seems that that purpose is to help the user set the proper length scale. It should do that, and in all other respects be similar to the other Hamiltonian generation routines. I think that would be a better separation of function responsibility.

PR #328, opened 2018-05-01
Better Grid separation in "standardized" jellium

idk3:


idk3:
This needs better tests but shows the general idea.

kevinsung:
Thanks @idk3, this is exactly what I asked for ... except I also think we should just completely delete `standardized_dual_basis_jellium_hamiltonian` now. I won't insist on it though.

idk3:
No worries :) I'm actually cleaning the rest of it up now and thinking about good tests to add.

PR #331, opened 2018-05-03
Allow more than two arguments in BCH expansion

kevinsung:


src/openfermion/utils/_bch_expansion.py
```diff
@@ -16,11 +16,50 @@
...
+
+    Raises:
+        ValueError: invalid order parameter.
+        TypeError: operator types are not all the same.
+    """
+    order = kwargs.get('order', 6)
+    if (not isinstance(order, int)) or order < 0:
+        raise ValueError('Invalid order parameter.')
+    if len(set(type(op) for op in ops)) != 1:
+        raise ValueError('Operators must all be of the same type.')
```
idk3: This should be a `TypeError` (saw you changed it in the docs!). The other thing is the message would be confusing if someone tried `bch_expand()`, I think TypeError is still appropriate there but the message should be different.

PR #335, opened 2018-05-11
Linear op

sheilaliuxl:


src/openfermion/utils/_sparse_tools.py
```diff
@@ -203,6 +203,71 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
...
+    """
+    if n_qubits is None:
+        n_qubits = count_qubits(qubit_operator)
+    if n_qubits < count_qubits(qubit_operator):
+        raise ValueError('Invalid number of qubits specified.')
+
+    n_hilbert = 2 ** n_qubits
+
+    def matvec(vec):
+        if (len(vec) != n_hilbert):
```
Spaceenter: My understand is that len(vec) could be larger than n_hilbert. As long as len(vec) equals to 2**m (where m >= n_qubits), it is OK. We can simply think "X0 Y1 Z4" as "X0 Y1 I2 I3 Z4 I5 ..."  (where I is identity op)?
@babbush @jarrodmcc

jarrodmcc:
Something weird going on with Travis failing to import scipy modules.  Will try restarting

src/openfermion/utils/_sparse_tools.py
```diff
@@ -203,6 +203,64 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
...
+    """
+    if n_qubits is None:
+        n_qubits = count_qubits(qubit_operator)
+    if n_qubits < count_qubits(qubit_operator):
+        raise ValueError('Invalid number of qubits specified.')
+
+    n_hilbert = 2 ** n_qubits
+
+    def matvec(vec):
+        """matvec for the LinearOperator class."""
```
jarrodmcc: Probably worth specifying which types vec supports.  My assumption is probably standard 1-d numpy vectors.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -203,6 +203,64 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
...
+        retvec = numpy.zeros(vec.shape, dtype=complex)
+        # Loop through the terms.
+        for qubit_term in qubit_operator.terms:
+            vecs = [vec]
+            tensor_factor = 0
+            coefficient = qubit_operator.terms[qubit_term]
+
+            for pauli_operator in qubit_term:
+                # Split vector by half and half for each bit.
+                num_splits = 2 ** (pauli_operator[0] - tensor_factor)
```
jarrodmcc: Is there anything you can do to elaborate on the logic here of why you are adding and subtracting tensor factor separately from the index on the qubit term itself?

src/openfermion/utils/_sparse_tools_test.py
```diff
@@ -108,6 +108,60 @@ def test_qubit_operator_sparse_n_qubits_not_specified(self):
...
+        """Testing with multiple terms."""
+        qubit_operator = QubitOperator(((0, 'Z'), (1, 'X')), 10.0) + 2.0 * \
+            QubitOperator('Y2')
+
+        vec = numpy.array([1, 2, 3, 4, 5, 6, 7, 8])
+        matvec_expected = 10 * numpy.array([3, 4, 1, 2, -7, -8, -5, -6]) + \
+            2.j * numpy.array([-2, 1, -4, 3, -6, 5, -8, 7], dtype=complex)
+
+        self.assertTrue(numpy.allclose(
+            get_linear_qubit_operator(qubit_operator) * vec, matvec_expected))
```
jarrodmcc: Instead of hard coding all the tests, let's have at least one that checks consistency between our other methods for building qubit operators and the LinearOperator defined either by their action on a random normalized vector, or their action on the identity matrix (for small cases) to check the matrix representations.  That way we also detect any convention changes between different features that happen by accident later.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -203,6 +203,64 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
...
+                             %(len(vec), n_hilbert))
+
+        retvec = numpy.zeros(vec.shape, dtype=complex)
+        # Loop through the terms.
+        for qubit_term in qubit_operator.terms:
+            vecs = [vec]
+            tensor_factor = 0
+            coefficient = qubit_operator.terms[qubit_term]
+
+            for pauli_operator in qubit_term:
```
jarrodmcc: Within each term all applications should commute, so if you can think of a way of parallelizing this loop it might be beneficial in the long term, but this is definitely just an added bonus.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -203,6 +203,64 @@ def qubit_operator_sparse(qubit_operator, n_qubits=None):
...
+                # Split vector by half and half for each bit.
+                num_splits = 2 ** (pauli_operator[0] - tensor_factor)
+                if num_splits >= 2:
+                    vecs = [v for iter_v in vecs
```
jarrodmcc: I know what you're doing here, but if you could add a few more lines of comment on your general algorithmic approach to implementing the action of the operators so it's easier to maintain in the future, it would be very helpful.  If there was any reference you used that might suffice, or if you derived it from scratch, just describe a bit the logic behind the approach?

src/openfermion/utils/_sparse_tools.py
sheilaliuxl: It should be more clear with newly added comments.
Since vector is split, and its size keeps getting smaller, there is a "- tensor_factor" because of the cache.

src/openfermion/utils/_sparse_tools.py
sheilaliuxl: Leaving it as a TODO item for a change in the future.

src/openfermion/utils/_sparse_tools.py
sheilaliuxl: Yes, it's from scratch. Comments are added, thanks for the suggestion!

src/openfermion/utils/_sparse_tools_test.py
sheilaliuxl: Good point, added compare test with another function.

jarrodmcc:
I think the changes look quite good.  Do you have any theories on the failures of the tests though? I would guess other parts might be breaking due to some circular dependencies extra imports introduced but I'm not sure.  Are the tests passing locally for you?

PR #336, opened 2018-05-17
Use ndarray instead of sparse matrices for states

### LINKED ISSUE #329: Consistently use numpy arrays instead of sparse matrices for states
@jarrodmcc has brought up something similar before, the main point being that using sparse matrices incurs a performance penalty and generally doesn't save space anyway because states will be dense. I also think this would be good because numpy and scipy always return numpy arrays for eigenvectors, so we wouldn't have to convert between datatypes to compare them.

I'm proposing that functions like `jw_get_gaussian_state`, `jw_hartree_fock_state`, and `jw_configuration_state` should return Numpy arrays instead of Scipy sparse matrices. @babbush is this too interface-breaking? It would be an easy change, and basically the only difference would be that states would have shape (N,) instead of (N, 1), which does not affect matrix-vector multiplication.

kevinsung:
Resolves #329 .


sheilaliuxl:
Looks like a minor type error left then for the shape var in _sparse_tools_test.py, only works for python 2.7, trying again with my other cl.

kevinsung:
I fixed it another way here.

kevinsung:
I think leaving it as float is fine because manually setting an array value is a pretty rare use case too. I think the most common operations are matrix-vector multiplications or scipy function calls and those handle conversions smoothly. For now I'd rather save space where possible but we can change this later.

PR #340, opened 2018-05-19
New function which returns MolecularData geometry of input PubChem names.

daochenw:
This allows easy geometry specification by non-specialists as the PubChem names are very flexible, e.g. 'H2O', 'water' or 'dihydrogen oxide' for the same molecule. Colloquial names like 'dry ice' or 'sugar' are also accepted. Requires the PubChemPy library.

### DIFF
--- a/examples/openfermion_tutorial.ipynb
+++ b/examples/openfermion_tutorial.ipynb
@@ -1018,6 +1018,33 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


babbush:
So this is starting to look pretty good. However, since this will be contributed to OpenFermion core instead of OpenFermion-Psi4, we need to enforce test coverage. You will need to write some unit tests in a separate file. Please look at the rest of OpenFermion for an idea of how to do that.

daochenw:
Sure, I've just done that, thanks for your help!

PR #342, opened 2018-05-21
added lambda norm function

babbush:


src/openfermion/utils/_lcu_util.py
```diff
@@ -17,6 +17,24 @@
 import math
 
 
+def lambda_norm(diagonal_operator):
```
idk3: This name isn't very descriptive, something like 'triangle_inequality_norm_bound' would be clearer.

src/openfermion/utils/_lcu_util.py
```diff
@@ -17,6 +17,24 @@
...
+        diagonal_operator: instance of DiagonalCoulombHamiltonian.
+
+    Returns:
+        lambda_norm: A float giving the lambda norm.
+    """
+    lambda_norm = 0.
+    n_qubits = diagonal_operator.one_body.shape[0]
+    for p in range(n_qubits):
+        for q in range(n_qubits):
+            lambda_norm += 0.75 * abs(diagonal_operator.two_body[p, q])
```
idk3: Explain the 0.75 in a comment! If it's from what I think it is you can do the same improvement in certain cases of the next line too :)

idk3: (This may be the reason for your test failures! But it's a little tricky to tell with a random test.)

Issue #339, opened 2018-05-17
Optimize get_linear_qubit_operator


kevinsung:
This is a place where speed is critical. I just wanted to record here some suggestions that came up so they're not lost:
- As mentioned at #335 , one of the loops can probably be parallelized
- We could use Cython or the Numpy C API or something like that

jarrodmcc:
Both loops can likely be parallelized, but with different strategies.  It's easier to parallelize over the sum between terms, then do a reduce operation, but that has a higher memory overhead and only speeds up QubitOperators with multiple terms.  The inner loop may actually be a bit tricky due to operations touching the same part of memory.

sheilaliuxl:
I guess the inner loop is a good candidate for Cython, while the outer one is good for multithreading.

jarrodmcc:
At some point we will want C++, Cython, or other bindings to optimize some of the routines.  The first addition of this will be the major hurdle, because we have to change the CI, build system, testing etc, but once it's there it has the potential to lower the barrier for others.  For any C++ change we make, I think the Python simulator should be left in place so that if the build fails on someone's system it can be defaulted back to.  Projectq worked this way and I think it's a reasonable choice.

babbush:
Did @sheilaliuxl 's pull requests sufficiently close this issue?

sheilaliuxl:
@babbush Inner loop (C(ython)) is still a TODO item, while the outer loop is done in #357.

PR #343, opened 2018-05-23
Parallelize H * v for the outer loop with multiprocessing

sheilaliuxl:
Partially resolves [#339](https://github.com/quantumlib/OpenFermion/issues/339) for the outer for loop.


kevinsung:
Please add a benchmark for this functionality to `examples/performance_benchmarks.py`.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -241,45 +284,21 @@ def matvec(vec):
...
-                }
-                vecs = [v for vp in xyz[pauli_operator[1]](vec_pairs)
-                        for v in vp]
-                tensor_factor = pauli_operator[0] + 1
-
-            # No need to check tensor_factor, i.e. to deal with bits left.
-            retvec += coefficient * numpy.concatenate(vecs)
-        return retvec
+        pool = multiprocessing.Pool(processes)
+        vecs = pool.map(term_on_vec, [(numpy.copy(vec), item[0], item[1])
```
jarrodmcc: My main worry for this is that for large systems, this is going to carry a heavy memory footprint, which will represent a tradeoff between the decreased computation time.  This is because this will store a copy for every operator in the sum, which could be in the 10's of thousands. 

The ideal fix would be some kind of adaptive parallel map-reduce that works through the map reduce tree with fixed chunk sizes.  However in lieu of something that complex, I think we have to set a parallelism limit to how many of these we store and work through the vector in parallel chunks.  For example, compute on 10 in parallel, reduce in parallel to 1, compute on next 10 in parallel, reduce those 10 with first etc.

jarrodmcc:
I can't tell if the test for 2.7 is failing because of the memory issues I mentioned on the test device, a core issue with using MP with 2.7, or a random load fluke on the testing machine.  I'll restart the tests to see if it repeats.

babbush:
@sheilaliuxl  since you are now officially a developer, you should add yourself to the authors list (in alphabetical section) and to the NOTICE with this PR.

PR #350, opened 2018-05-24
Add davidson aglorithm to find out the lowest eigen values & vectors.

sheilaliuxl:


sheilaliuxl:
@babbush @jarrodmcc 
Looks like the coverage is running infinitely, have you ever seen this/ do you have any idea why? Thanks!

src/openfermion/utils/_davidson.py
```diff
@@ -0,0 +1,273 @@
...
+# Exceptions.
+class DavidsonError(Exception):
+    """Exceptions."""
+    pass
+
+
+class Davidson(object):
+    """Davidson algorithm to get the n states with smallest energies."""
+
+    def __init__(self, linear_operator, linear_operator_diagonal, eps=1e-6):
```
jarrodmcc: If I recall correctly, Davidson also has another parameter which is the number of vectors to use in the subspace solves, should we pass that in here as well?

src/openfermion/utils/_davidson.py
```diff
@@ -0,0 +1,273 @@
...
+
+    def get_lowest_n(self, n_lowest=1, initial_guess=None, max_iterations=300):
+        """
+        Returns `n` smallest eigenvalues and corresponding eigenvectors for
+            linear_operator.
+
+        Args:
+            n(int): The number of states corresponding to the smallest eigenvalues
+                and associated eigenvectors for the linear_operator.
+            initial_guess(numpy.ndarray[complex]): Initial guess of eigenvectors
```
jarrodmcc: If I'm guessing right, this should have n columns, each of which is a guess at an eigenvector.  Do we need to put any constraints on this?  For example, do they need to be orthogonal for success?

We should maybe also check if initial_guess.shape[1] < n but not none, and pass a warning that we need to generate random vectors for the rest.

babbush:
Looks like restarting the Travis build solved the coveralls problem.

src/openfermion/utils/_davidson.py
sheilaliuxl: 1. The only constraints is that all guess_v vectors are not zero. Added check.
2. Added random vectors in this case, though now really required.

src/openfermion/utils/_davidson.py
sheilaliuxl: Done, there was a TODO comment in the previous version, now removed.

babbush:
Hi Sheila, usually, yes, we try to pass the coverage check.

sheilaliuxl:
Now only one line is not covered, which is kind of expected, since it should be very (or even extremely) rare to see all new directions are linearly dependent with original vectors.

PR #354, opened 2018-05-31
Add support for real eigenvectors when specified, up to the user to decide when it should work.

sheilaliuxl:
Real eigen-vectors or not is the property of a matrix, therefore put it in the constructor.


sheilaliuxl:
A follow up with @jarrodmcc's suggestion in [#350](https://github.com/quantumlib/OpenFermion/pull/350).

jarrodmcc:
Even with my google email it won't let me override the cla notices.  I'll just wait till it passes all the tests and merge it in

babbush:
that's very strange Jarrod. I don't know why I can do it but you cannot.

PR #356, opened 2018-06-01
Added function to remove qubits using symmetries

sammcardle30:
Hi OpenFermion developers,

I’ve made a function to reduce 2 qubits from Bravyi-Kitaev mapped Hamiltonians using symmetries, as described in arXiv:1701.08213. It’s been useful for my research so far, so I thought that it would be a good file to contribute.
I’ve included unit tests of the function, and have tried to adhere to the style guide- but please do let me know if there’s anything which needs changing.
I’d also be happy to add a tutorial, perhaps of using this function and NMO’s to reduce the LiH Hamiltonian from 12 to 6 qubits, as described in arXiv:1803.10238. 

Best regards,

Sam McArdle


src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+
+        Returns:
+                qubit_hamil: The qubit Hamiltonian corresponding to
+                             the supplied fermionic Hamiltonian, with
+                             two qubits removed using spin symmetries.
+
+        Notes: This function reorders the spin orbitals as all spin-up, then
+               all spin-down. It uses the OpenFermion bravyi_kitaev_tree
+               mapping, rather than the bravyi-kitaev mapping.
+
```
babbush: extra space here should be deleted.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+                              is less than the number of electrons in the
+                              molecule if some orbitals have been assumed
+                              filled).
+
+        Returns:
+                qubit_hamil: The qubit Hamiltonian corresponding to
+                             the supplied fermionic Hamiltonian, with
+                             two qubits removed using spin symmetries.
+
+        Notes: This function reorders the spin orbitals as all spin-up, then
```
babbush: The convention in OpenFermion is to keep the even-odd convention in spin orderings. So I object to this returning all then down by default. There is a function called "reorder" in utils/_operator_utils.py which assists with reorderings.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+               all spin-down. It uses the OpenFermion bravyi_kitaev_tree
```
babbush: Why? The tree mapping actually has higher many-body order when the number of orbitals is not a binary power.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+               mapping, rather than the bravyi-kitaev mapping.
+
+    """
+
+    def edit_hamil_for_spin(qubit_hamiltonian, spin_orbital, orbital_parity):
```
babbush: It would be a better design pattern to define this function outside the namespace of the other. Please do that. Also please write out "hamiltonian".

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+        """ Removes the Z terms acting on the orbital from the Hamiltonian.
+        """
+
+        new_qubit_dict = {}
+        for key, val in qubit_hamiltonian.terms.items():
```
babbush: better to call this "term, coefficient" instead of "key, val" or something like that. It makes the code more readable and easy to understand.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+                else:
+                    old_val = new_qubit_dict.get(key)
+                    new_qubit_dict[key] = val + old_val
+
+        qubit_hamiltonian.terms = new_qubit_dict
+        qubit_hamiltonian.compress()
+
+        return qubit_hamiltonian
+
+
```
babbush: delete extra spaces. If you run the pep8 linter on this code it would help you find all the style guide issues like this.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+    # Catch errors if inputs are of wrong type.
+    if (type(fermion_hamil) is not
+        openfermion.ops._fermion_operator.FermionOperator):
```
babbush: since you import FermionOperator, can't you just write "FermionOperator" instead of "openfermion.ops._fermion_operator.FermionOperator"?

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
+""" Module to remove two qubits from the problem space
+    using conservation of electron number and conservation
+    of electron spin. As described in arXiv:1701.08213.
+"""
+
+
+import openfermion
```
babbush: if possible, it is best if you just import everything you need explicitly, as you do below. This way, you shouldn't need to "import openfermion".

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+from openfermion.ops import FermionOperator, QubitOperator
+from openfermion.transforms import get_fermion_operator, bravyi_kitaev_tree
+from openfermion.utils import reorder, up_then_down, prune_unused_indices
+import numpy
```
babbush: python package imports go before openfermion imports. Leave a space between import numpy and the other openfermion imports.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
+""" Module to remove two qubits from the problem space
+    using conservation of electron number and conservation
+    of electron spin. As described in arXiv:1701.08213.
+"""
+
+
+import openfermion
+from openfermion.ops import FermionOperator, QubitOperator
+from openfermion.transforms import get_fermion_operator, bravyi_kitaev_tree
```
babbush: our style guide mandates alphabetical imports. So bravyi_kitaev_tree should come before get_fermion_operator.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+        openfermion.ops._fermion_operator.FermionOperator):
+        raise ValueError('Supplied operator should be an instance'
+                         'of FermionOperator class')
+    if type(active_orbitals) is not int:
+        raise ValueError('Number of active orbitals should be an integer.')
+    if type(active_electrons) is not int:
+        raise ValueError('Number of active electrons should be an integer.')
+
+    # Arrange spins up then down, then BK map to qubit Hamiltonian.
+    ferm_hamil_reorder = reorder(fermion_hamil, up_then_down)
```
babbush: why is this necessary?

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Catch errors if inputs are of wrong type.
+    if (type(fermion_hamil) is not
+        openfermion.ops._fermion_operator.FermionOperator):
+        raise ValueError('Supplied operator should be an instance'
+                         'of FermionOperator class')
+    if type(active_orbitals) is not int:
+        raise ValueError('Number of active orbitals should be an integer.')
+    if type(active_electrons) is not int:
+        raise ValueError('Number of active electrons should be an integer.')
```
babbush: Since you raise errors in the function you should add a "raises" section to the docstrings (similar to the "args" section).

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+        parity_final_orb = -1
+        parity_middle_orb = 1
+
+    # Removes the final qubit, then the middle qubit.
+    qubit_hamil = edit_hamil_for_spin(qubit_hamil, active_orbitals,
+                                      parity_final_orb)
+    qubit_hamil = edit_hamil_for_spin(qubit_hamil, active_orbitals/2,
+                                      parity_middle_orb)
+    qubit_hamil = prune_unused_indices(qubit_hamil)
+
```
babbush: you should have at most one empty space between code in a function.

src/openfermion/utils/_remove_symmetry_qubits.py
```diff
@@ -0,0 +1,118 @@
...
+    if remainder == 0:
+        parity_final_orb = 1
+        parity_middle_orb = 1
+    elif remainder == 1:
+        parity_final_orb = -1
+        parity_middle_orb = 1
+    elif remainder == 2:
+        parity_final_orb = 1
+        parity_middle_orb = -1
+    elif remainder == 3:
```
babbush: this can be an "else"

src/openfermion/utils/_remove_symmetry_qubits_test.py
```diff
@@ -0,0 +1,171 @@
+""" Unit tests of remove_symmetry_qubits().
+    Test that it preserves the correct
+    ground state energy, and reduces the number
+    of qubits required by 2.
+"""
+
+import unittest
+
+import openfermion
+import openfermionpyscf
```
babbush: these are awesome tests but unfortunately you cannot import openfermionpyscf inside of openfermion (but you can import openfermion inside of openfermionpyscf). There are many reasons for this. Among them: some people want to use openfermion without installing openfermionpyscf / pyscf (for instance, if you use windows where it is impossible to install pyscf). And we can't have tests failing for those users. So if you want to have tests that depend on openfermionpyscf, you'd need to put them in the openfermionpyscf repo, which you are welcome to do. Otherwise, perhaps you should hardcode the outputs so that these tests can run without openfermionpyscf. Thanks!

src/openfermion/utils/_remove_symmetry_qubits_test.py
```diff
@@ -0,0 +1,171 @@
...
+from openfermion.hamiltonians import MolecularData
+from openfermionpyscf import run_pyscf
+from openfermion.utils import eigenspectrum
```
babbush: alphabetical imports: ops comes before utils.

src/openfermion/utils/_remove_symmetry_qubits_test.py
```diff
@@ -0,0 +1,171 @@
...
+from openfermion.ops import FermionOperator, QubitOperator
+from openfermion.transforms import get_fermion_operator
+
+import _remove_symmetry_qubits
```
babbush: you should edit the __init__ file in utils to import your function and then import it through utils.

src/openfermion/utils/_remove_symmetry_qubits_test.py
```diff
@@ -0,0 +1,171 @@
...
+
+
+
```
babbush: delete extra space.

src/openfermion/utils/_remove_symmetry_qubits_test.py
```diff
@@ -0,0 +1,171 @@
...
+                         h2_sto_numorb-2)
+
+        self.assertEqual(number_of_qubits(h2_631_qbt, h2_631_numorb),
+                         h2_631_numorb-2)
+
+        self.assertEqual(number_of_qubits(lih_sto_qbt, lih_sto_numorb),
+                         lih_sto_numorb-2)
+
+
+if __name__ == '__main__':
```
babbush: we do not have __main__ in any of our files. The tests are run through the pytest module.

babbush:
note that tests fail since we do not build openfermionpyscf during the continuous integration testing of openfermion, for reasons I explain in the review.

sammcardle30:
Hi Ryan,
Brilliant, thank you for all of your feedback and suggestions! What is the best way for me to make the changes that you've suggested and resubmit the files for another review? 
The reason I've reordered the orbitals and used the BK tree mapping rather than the default ordering and mapping is because only the former seemed to leave Hamiltonians where the middle and final qubits were acted on by the Z or I operators; using the latter combination didn't necessarily give 2 qubits which I could reduce. I can have a think about how the algorithm would change if the qubits are instead ordered up-down-up-down, if you'd prefer? I'll also try to fix the test function so that it doesn't include pyscf. Can I use the approach in this test https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/transforms/_binary_codes_test.py
where the LiH Hamiltonian is just loaded in? 
Thanks again,
Sam

babbush:
Hi Sam,

To make changes, just push additional commits to the git branch from which you submitted this pull request. It will automatically push the changes here. It sounds like you prefer the up-then-down ordering because it is convienient to perform the symmetry reduction. In that case, just transform the Hamiltonian to up-then-down, apply the symmetry reduction, then transform back to even-odd. Perhaps I misunderstand something though and the issue is trickier than that. Yes, it is fine to use the approach where the LiH Hamiltonian is loaded in. We keep those preloaded files around precisely for testing purposes like this. Thanks!

sammcardle30:
Hi Ryan,
Perfect, thank you. Ah ok, that sounds like the best approach then; I'll stick with the all up- all down ordering to make the symmetry reduction easier, and then transform the qubit Hamiltonian back into the default orbital ordering, as you suggest. That's ideal then, I'll change the tests to use the presaved files.
I'll make all of the changes over the next few days, and resubmit them on Monday.
Thanks,
Sam

sammcardle30:
Hi Ryan,

I'm just working on integrating the changes you've suggested, and have run into a complication with the ordering of spins.
The method I've implemented is that described in 'Tapering off qubits to simulate fermionic Hamiltonians' by Bravyi et al. It relies on ordering the orbitals all spin up, then all spin down. This is because the middle and final rows of the BK transform matrix have 1's as half or all of their entries, respectively. Therefore the middle and final entries of the BK transformed state vector are the number of spin up electrons (mod 2) and the total number of electrons (mod 2), respectively. Because these quantities are conserved, the Hamiltonian won't change them, so only acts on the middle and final qubits with I or Z operators. The effect of these operators can be precomputed, leaving those 2 qubits redundant. 
I've looked at your suggestion of using the "reorder" function to swap the ordering back to up-down. However, because the qubit Hamiltonian obtained is in the BK mapping, and so each qubit may store a combination of orbitals, it's in general non-trivial to just swap the indices on the qubit operators to reorder the spins. From what I can see, the best way to get back into the up-down ordering would be to apply the inverse BK transform to the reduced Hamiltonian, reorder the fermion operator, and then BK transform again. However, as OpenFermion doesn't currently have a reverse BK transform function, I haven't been able to check that this indeed works. 

I was wondering if you had any suggestions about the best way to proceed? Would you prefer a workaround to restore the up-down ordering, or is it sufficient to make it clear in the function description (and/or name) that it reorders the orbitals?

Thanks for your help,

Sam

babbush:
Hi Sam,

Shouldn't it be possible to just identify which qubits are acted upon trivially if BK is performed in the even-odd ordering? If not, let me ask this: is it even the case that each qubit encodes orbitals of a well defined spin post-symmetry reduction? If so, do we still have the same number of up spins as down spins?

sammcardle30:
Hi Ryan,
When working in the default even-odd ordering, only the final qubit is guaranteed to be acted upon trivially, as it stores the sum of the occupation numbers of all of the orbitals (mod2). However, no other qubit will necessarily be acted upon trivially. So we can in general only reduce one qubit, rather than two, if we don't initially reorder the orbitals (verified with H2 in ccpvdz basis). In some circumstances we are able to reduce two qubits without reordering the orbitals (this is true for LiH and H2 in the STO bases), but in general we'll need to reorder the orbitals to remove two qubits.
After symmetry reduction (in the all up- all down ordering) the qubits still encode orbitals of a well defined spin and we still have an equal number of up and down spins. The symmetry reduction makes use of the fact that in this ordering, the occupancy of the highest up and down orbitals (the final two orbitals in the default ordering) are stored only in the middle and final qubits. The middle qubit stores the sum of the occupancies of the up orbitals (mod2). The final qubit stores the sum of the occupancies of all of the orbitals (mod2). As the value of these sums stays fixed throughout the simulation, we eliminate these qubits, so I suppose in a loose sense we have dropped the highest up and down orbitals from the simulation, and are inferring their occupation by enforcing conservation of spin and electron number.
Best regards,
Sam

babbush:
Based on what you are saying I think we should just forget about the odd-even convention for the purposes of this function and just have it output things in whatever order is natural. Please just add a WARNING note about this in the docstrings for the function. Thanks!

sammcardle30:
Hi Ryan,
I think that sounds like the best approach. I've now made all of your suggested changes.
Thanks!
Sam

babbush:
Please fix all failing tests.

sammcardle30:
Travis is failing to build, but I'm not sure why. It gives several error messages of the form:
ImportError while importing test module '/home/travis/build/quantumlib/OpenFermion/src/openfermion/hamiltonians/_plane_wave_hamiltonian_test.py'.
Hint: make sure your test modules/packages have valid Python names.

What correction do I need to make to avoid these errors?

babbush:
Hmm, I don't know, I've been looking at this for a few minutes now. Perhaps there is a problem with circular imports. @kevinsung can you take a quick look at this. Any ideas?

kevinsung:
It's a problem with circular imports. In `utils/__init__.py`, move the line
```
from ._remove_symmetry_qubits import (symmetry_conserving_bravyi_kitaev,
                                      edit_hamiltonian_for_spin)
```
to the end of the file. Then, in in `remove_symmetry_qubits_test`, import the functions from `_remove_symmetry_qubits` directly, rather than trying to import the module `remove_symmetry_qubits` from `utils`.

kevinsung:
By the way, you can run tests locally using `pytest`. Just go into `src/openfermion` and execute `pytest`. If you don't have it, install it with `pip install pytest`.

babbush:
That CLA bot is really driving me nuts. What happened is that I merged master into your fork at some point because your branch got outdated. And that always confuses it because now it thinks the PR is authored by two people. Oh well.

sammcardle30:
Brilliant, thank you @kevinsung!

sammcardle30:
@kevinsung, sorry to be a pain, but having made the changes you suggested, there are still similar error messages coming up, such as:
____ ERROR collecting src/openfermion/hamiltonians/_chemical_series_test.py ____
ImportError while importing test module '/home/travis/build/quantumlib/OpenFermion/src/openfermion/hamiltonians/_chemical_series_test.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
src/openfermion/__init__.py:23: in <module>
    from openfermion.hamiltonians import *
src/openfermion/hamiltonians/__init__.py:17: in <module>
    from ._jellium import (
src/openfermion/hamiltonians/_jellium.py:19: in <module>
    from openfermion.utils._grid import Grid
src/openfermion/utils/__init__.py:100: in <module>
    from ._remove_symmetry_qubits import (symmetry_conserving_bravyi_kitaev,
src/openfermion/utils/_remove_symmetry_qubits.py:21: in <module>
    from openfermion.transforms import bravyi_kitaev_tree, get_fermion_operator
src/openfermion/transforms/__init__.py:27: in <module>
    from ._conversion import (get_boson_operator,
src/openfermion/transforms/_conversion.py:22: in <module>
    from openfermion.hamiltonians import MolecularData
E   ImportError: cannot import name 'MolecularData'

kevinsung:
The problem is that you're importing from `transforms` in a file from `utils`, but generally `utils` is lower in the import hierarchy. I'll think about the best way to resolve it.

By the way, why are you using `bravyi_kitaev_tree` instead of `bravyi_kitaev`? Is that really necessary?

kevinsung:
Are you sure that your observation generalizes? Most of the literature, including the papers you referenced, use the mapping `bravyi_kitaev` rather than `bravyi_kitaev_tree`, so it should be possible for you to use it too. Perhaps you're interested not in the middle and final qubits, but some qubits that are indexed by powers of 2. I'm not sure.

In any case, these files should be moved to the `transforms` module instead of `utils`. That should resolve the import errors.

sammcardle30:
From the checks that I've done, using bravyi_kitaev_tree seems to generalise. That's certainly something I'd be happy to look into, although the fact that there aren't 2 qubits acted upon trivially for LiH under the ordinary BK transform means the method would seemingly have to change from that of the IBM Tapering paper.

Brilliant, thank you for all of your help- I've moved the files to transforms now, as you suggest.

kevinsung:
You're saying that Table I of the "Tapering" paper uses `bravyi_kitaev_tree` instead of `bravyi_kitaev`?

babbush:
As a minor historical note - this idea to remove the two qubits acted upon trivially, is actually not due to the Bravyi paper. Bravyi's paper explains _why_ there are two qubits acted upon trivially, but the original observation was made by us in [Phys. Rev. X 6, 031007](https://journals.aps.org/prx/abstract/10.1103/PhysRevX.6.031007), which was on arXiv well over a year before Bravyi's paper. There, in Appendix A, we discuss this observation. Note that I used the standard BK transform, not the tree transformed, and it seemed to work at least for H2. It is rather mysterious to me why you'd need the tree transform for this to work for LiH but if you have examined each qubit in that Hamiltonian closely then I'll take your word for it.

sammcardle30:
I'm not sure about Table 1- the part of the paper I've been working from is section 7. They reference the original Bravyi and Kitaev paper, which I believe corresponds to the usual bravyi_kitaev function in OpenFermion (although they refer to it as the binary tree mapping, so perhaps they are using the BK_tree version) ? I would have expected the ordinary bravyi_kitaev function to work (it does for H2 in the sto basis, as Ryan has noted above), but for whatever reason it doesn't seem to with LiH (or ccpvdz H2, and presumably other molecules).

Ah, apologies Ryan- I'll make sure to reference both papers next time I change the file.

kevinsung:
They explicitly define their mapping in Eq. (50), and this corresponds to `bravyi_kitaev`, not `bravyi_kitaev_tree`.

sammcardle30:
In that case, I'm not sure why using bravyi_kitaev() on the fermionic LiH Hamiltonian doesn't produce 2 qubits that are acted upon trivially - I agree that we would expect it to.

sammcardle30:
Hi Ryan,
I'm just adding some more tests now, which will hopefully cover the majority of the remaining code.

sammcardle30:
Hi Ryan,
Coverage is now at ~92%, which is as high as I can get it without access to additional saved molecules- is that sufficiently high?

babbush:
Hi Sam,

Normally we don't like to merge PRs that lower test coverage by more than 0.01% and this one decreases overall coverage by 0.05%. However, this PR has gone on for a while so perhaps we can make an exception. But let me ask this: are you aware of which lines are not covered? If so, what functionality is not covered and why would you need additional molecule files to test them?

sammcardle30:
The two uncovered lines are if clauses which are entered when (the number of active electrons % 4) is equal to 1 or 3. I suppose I could enter into these clauses by changing the active space of LiH such that it has 1 or 3 active electrons, but as this would result in an imbalanced Hamiltonian, it wouldn't make sense to apply the spin reduction, so the test would be somewhat for the sake of just increasing the coverage.

babbush:
Okay, I merged the PR. If you want to be added to the release paper please email me so I can send you the link.

sammcardle30:
Fantastic, thank you for helping me to get this function added Ryan!

PR #358, opened 2018-06-04
 Add period cutoff to Jellium plane wave hamiltonian. 

Spaceenter:
Note the unit tests are just to ensure code coverage. We can add integration tests after figuring out the correct formula for dual basis hamiltonian with period cutoff.


src/openfermion/hamiltonians/_jellium.py
```diff
@@ -89,13 +89,16 @@ def plane_wave_kinetic(grid, spinless=False, e_cutoff=None):
...
+def plane_wave_potential(grid, spinless=False, e_cutoff=None,
+                         non_periodic=False, period_cutoff=None):
     """Return the e-e potential operator in the plane wave basis.
 
     Args:
         grid (Grid): The discretization to use.
         spinless (bool): Whether to use the spinless model or not.
         e_cutoff (float): Energy cutoff.
+        non_periodic (bool): If the system is non-periodic.
+        period_cutoff (float): Period cutoff.
```
jarrodmcc: Perhaps state what the default is here for people that want to specify non_periodic but not the cutoff

src/openfermion/hamiltonians/_jellium.py
```diff
@@ -185,6 +194,8 @@ def dual_basis_jellium_model(grid, spinless=False,
         include_constant (bool): Whether to include the Madelung constant.
             Note constant is unsupported for non-uniform, non-cubic cells with
             ions.
+        non_periodic (bool): If the system is non-periodic.
+        period_cutoff (float): Period cutoff.
```
jarrodmcc: Same as above

src/openfermion/hamiltonians/_jellium_test.py
```diff
@@ -418,3 +418,20 @@ def test_plane_wave_energy_cutoff(self):
...
+    def test_plane_wave_period_cutoff(self):
+        grid = Grid(dimensions=2, length=2, scale=1.0)
+        spinless = True
+        period_cutoff = 0.
+
+        hamiltonian_1 = FermionOperator()
+        jw_1 = jordan_wigner(hamiltonian_1)
+        spectrum_1 = eigenspectrum(jw_1)
+
+        hamiltonian_2 = jellium_model(grid, spinless, True, False, None, True,
```
jarrodmcc: I'm a little confused by this test, the difference should definitely be larger than 0 for any deviation.  I guess that this will be changed later in integration testing? It might make more sense to test against only the potential operator and see if its 0, to remove contributions by the kinetic operator.

src/openfermion/hamiltonians/_plane_wave_hamiltonian_test.py
```diff
@@ -126,3 +126,20 @@ def test_plane_wave_energy_cutoff(self):
...
+
+        h_1 = plane_wave_hamiltonian(grid, geometry, True, True, False, None)
+        jw_1 = jordan_wigner(h_1)
+        spectrum_1 = eigenspectrum(jw_1)
+
+        h_2 = plane_wave_hamiltonian(grid, geometry, True, True, False, None,
+                                     True, period_cutoff)
+        jw_2 = jordan_wigner(h_2)
+        spectrum_2 = eigenspectrum(jw_2)
+
```
jarrodmcc: Same comment for ths test, I think it only really makes sense to test against the potential operators and not have the kinetic included, but if this is only to placehold until integration testing, it might be okay.

src/openfermion/hamiltonians/_jellium.py
Spaceenter: Done - also other instances.

src/openfermion/hamiltonians/_jellium.py
Spaceenter: Done - also other instances.

src/openfermion/hamiltonians/_jellium_test.py
Spaceenter: This is really a placeholder for future real integration test, when we figure out the correct formula for dual basis cutoff, periodic_cutoff will be set to non-zero, and hamiltonian_1 will be replaced with a jellium model.
Added a TODO.

src/openfermion/hamiltonians/_plane_wave_hamiltonian_test.py
Spaceenter: Yes, this test is really only for code coverage, and a placeholder for future integration test. Added a TODO. 

Also, as mentioned in our email thread, since plane wave external potential is implemented as an inverse Fourier transformation of dual basis external potential:
https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/hamiltonians/_plane_wave_hamiltonian.py#L97 ,
We are not really able to do integration test for external potential.

Spaceenter:
@jarrodmcc It's hard to keep code coverage without the two 'fake' integration test. I guess it's OK to keep them as placeholders for future real integration tests? Added TODO there to remind us.

babbush:
@jarrodmcc @Spaceenter are there outstanding changes needed to merge this PR? If not, let's try to merge before we cut the new release (which we are doing today or tomorrow).

PR #359, opened 2018-06-06
rewrite get ground states by particle number

kevinsung:
This changes `jw_get_ground_states_by_particle_number` to `jw_get_ground_state_at_particle_number`. The reason I made the change is to have a consistent interface with `get_ground_state`: it returns one eigenvalue and one eigenvector, the eigenvector being a numpy array. I also removed some expensive checks that I think should be left up to the user.

@babbush it turns out that the method you gave previously (simply pass a state with the correct symmetries to the scipy solver) does not work all the time. I have found that sometimes it returns a lower eigenvalue than the correct one. For this reason, I'm not sure how one would implement this functionality using LinearOperator, but that would be useful.


babbush:
Hi Kevin, that is interesting. My understanding was that the Lanczos algorithm essentially applies the Hamiltonian to the initial vector over and over in order to project to an eigenstate. It seems to me then that if this eigenstate has a symmetry that is conserved by the Hamiltonian, it should only output solutions with that same symmetry. @jarrodmcc do you have an idea what might be going on?

jarrodmcc:
Removing the additional eigenvectors is a good start to removing that behavior.  It's not clear why we asked for 3 by default.  If you ask for more than one but provide only a single guess, the second two guesses will probably be random, and if they overlap with a lower energy state than the one found originally, it will supersede the original.

jarrodmcc:
It's also possible for numerical noise to bump you into other eigenstates, but this isn't usually as common.

kevinsung:
@jarrodmcc to be clear, I wrote the old version of this method and back then I cared about obtaining multiple eigenvectors but in hindsight it didn't really matter. This function always worked, I'm just editing it now to make it nicer.

What @babbush and I are talking about is that it seems that we should be able to implement this functionality simply as
```
eigs, vecs = scipy.sparse.linalg.eigsh(sparse_operator, v0=jw_hartree_fock_state(n_electrons, n_qubits), k=1)
return eigs[0], vecs[:, 0]
```
This would also allow it to work for LinearOperators. However, I have found that this does not work all the time.

kevinsung:
Here is an example of the alternative method failing.
```
import openfermion

# Generate Jellium model
# ----------------------
dimensions = 2
grid_length = 2
wigner_seitz_radius = 10.
filling_fraction = 0.5
spinless = True

n_modes = grid_length**dimensions * (2 - spinless)
n_electrons = int(filling_fraction * n_modes)

length_scale = openfermion.wigner_seitz_length_scale(
    wigner_seitz_radius, n_electrons, dimensions)
grid = openfermion.Grid(dimensions, grid_length, length_scale)

# Get FermionOperator
model = openfermion.jellium_model(grid, spinless, plane_wave=False)

# Get ground state and energy at correct particle number
# ------------------------------------------------------
hamiltonian_sparse = openfermion.get_sparse_operator(model)

energy, state = openfermion.get_ground_state(
    hamiltonian_sparse, initial_guess=openfermion.jw_hartree_fock_state(n_electrons, n_modes))

num = openfermion.get_sparse_operator(openfermion.number_operator(n_modes))
print('Wanted {} electrons'.format(n_electrons))
print('Got {} electrons'.format(openfermion.expectation(num, state).real))
```
Output:
```
Wanted 2 electrons
Got 3.999999999999998 electrons
```

jarrodmcc:
I was just suggesting an alternative mechanism by which this procedure can fail. It can certainly fail in the default case, especially depending on the internal implementation and its numerical robustness.

jarrodmcc:
Would it be possible to also add an option to use by default the new Davidson routines that Sheila put into place? If they are working they should be much faster than what is in scipy for these problems.  Possibly more stable as well but I'm not sure

kevinsung:
That's a good idea, but I ask that we leave it for another PR. I opened #362 for it.

kevinsung:
Ok I fixed that and also ran pylint on the `transforms` module. Need your approval again.

Issue #361, opened 2018-06-08
Initial guess is not actually optional for Davidson algorithm


kevinsung:
```
from openfermion import *

model = fermi_hubbard(2, 2, 1.0, 4.0)
sparse_operator = get_sparse_operator(model)
davidson = Davidson(linear_operator=sparse_operator,
                    linear_operator_diagonal=sparse_operator.diagonal())
_ = davidson.get_lowest_n()
```
Output:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-1-35a3a44d039f> in <module>()
      5 davidson = Davidson(linear_operator=sparse_operator,
      6                     linear_operator_diagonal=sparse_operator.diagonal())
----> 7 _ = davidson.get_lowest_n()

~/Projects/OpenFermion/src/openfermion/utils/_davidson.py in get_lowest_n(self, n_lowest, initial_guess, max_iterations)
    126         # 2. Checks for initial guess vectors' dimension is the same to that of
    127         # the operator.
--> 128         if initial_guess.shape[0] != len(self.linear_operator_diagonal):
    129             raise ValueError('Guess vectors have a different dimension with '
    130                              'linear opearator diagonal elements: {} != {}.'

AttributeError: 'NoneType' object has no attribute 'shape'
```

The definition of `get_lowest_n` is
```
    def get_lowest_n(self, n_lowest=1, initial_guess=None, max_iterations=None):
```
@sheilaliuxl should `initial_guess` be optional or not? If not, then it should not have a default value of `None`.

PR #364, opened 2018-06-08
Fix bug of optional args for get_lowest_n() in Davidson algo

sheilaliuxl:
As suggested by @kevinsung in #361.

1. Fix minor bug;
2. Add unit test for the new benchmark function with (Parallel)LinearQubitOperator;
3. Replace print with logging, so that one can control what to see with DEBUG, INFO, FATAL options;
4. Fix most pylint errors.


sheilaliuxl:
Seems coverage test is stuck again. @babbush

babbush:
Go for it

kevinsung:
I don't think my account has the permissions.

jarrodmcc:
According to travis, CI builds are now taking on the order of 35-40 minutes each, which is perhaps something we should consider looking into, as it makes even merging new branches into these for additional tests a bit awkward.

jarrodmcc:
Well, the versions are often done in parallel, so 10 minutes wall clock, but sometimes they don't seem to have available computers to do that, so it inches up in time considerably at random

PR #365, opened 2018-06-08
Increased test coverage for remove_symmetry_qubits

sammcardle30:
I've added additional tests for the function that I added the other day, so that the code coverage is increased. I also added additional description of its use with the Fermi-Hubbard model in the documentation.


sammcardle30:
I'm not sure why it's showing 32 commits, as I've only made 2 since the function was merged- seemingly the old commits are still showing. Is there a way for me to get rid of this, so that it doesn't wrongly show my branch as being a lot of commits ahead of master in future?
Thanks,
Sam

sammcardle30:
Thanks Ryan- I think it was the latter of those two options.

Issue #363, opened 2018-06-08
Davidson algorithm should allow sparse matrix input


kevinsung:
Currently it only works for LinearOperators, but it should also work for sparse matrices, right?

sheilaliuxl:
Absolutely, mostly within 20 iterations for dimension at the order of 10^6 or even higher.

Actually it works for not only sparse matrices, but also for more general matrices when they're diagonal dominant, so that the approximation (H - \lambda I)^{-1} \approx (D_H - \lambda I)^{-1} is valid.

-- Seems no easy way to insert equations in discussion :(

sheilaliuxl:
Davidson's key inputs are:
1. linear operator (scipy.sparse.linalg.LinearOperator);
2. linear operator's exact diagonal elements (without calculating the whole matrix of course);

With a sparse matrix as the starting point, I'd suggest converting it to a LinearOperator as well, just as the QubitOperator case, and then they can share the same underlying Davidson Algo implementation, which had better be decoupled for clarity.

PR #369, opened 2018-06-12
Sparse Davidson

sheilaliuxl:
A duplicate PR to #366 due to CLA failure.
As suggested in #363.


babbush:
@sheilaliuxl - in the future, just ignore the CLA bot. It gets upset for strange reasons all the time. We can easily reset it.

sheilaliuxl:
Ack. @babbush

babbush:
Ack?

src/openfermion/utils/_davidson.py
```diff
@@ -158,6 +162,8 @@ def get_lowest_n(self, n_lowest=1, initial_guess=None, max_iterations=None):
         while (num_iterations < max_iterations and not success):
             (eigen_values, eigen_vectors, mat_eigen_vectors, max_trial_error,
              guess_v, guess_mv) = self._iterate(n_lowest, guess_v, guess_mv)
+            logging.info("Eigenvalues for iteration %d: %s, error is %f.",
```
jarrodmcc: I'm not opposed to it, but have we been using logging anywhere else in the code? Should we start?

sheilaliuxl: In [examples/](https://github.com/quantumlib/OpenFermion/blob/master/examples/performance_benchmarks.py), I replaced `print` with `logging.info`, which I believe is a more standard way for tracing.

Though it's not present in examples/ yet, I'm sure it'll be useful:
* in production;
* for debugging (I tried to log it in unit test, it doesn't work yet, not sure why :( ).

PR #372, opened 2018-06-19
Function to normal order in chemistry convention

babbush:
This will be used for the low rank stuff. @kevinsung are my doc strings okay?

### DIFF
--- a/src/openfermion/utils/_operator_utils.py
+++ b/src/openfermion/utils/_operator_utils.py
@@ -46,9 +46,9 @@ def chemist_ordered(fermion_operator):
 
     The normal ordering convention for chemists is different.
     Rather than ordering the two-body term as physicists do, as
-    :math: a^\dagger a^\dagger a a
+    :math: `a^\dagger a^\dagger a a`
     the chemist ordering of the two-body term is
-    :math: a^\dagger a a^\dagger a
+    :math: `a^\dagger a a^\dagger a`
 
     TODO: This routine can be made more efficient.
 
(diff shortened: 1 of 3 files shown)


babbush:
@kevinsung I've added a test that absolutely checks for compliance with chemist ordering. I was asking about the ":math:" thing in the doc strings. Does that look okay / will it render okay in sphinx? If so, I think we're now good to merge.

src/openfermion/utils/_operator_utils.py
```diff
@@ -37,6 +37,54 @@ class OperatorUtilsError(Exception):
...
+class OperatorSpecificationError(Exception):
+    pass
+
+
+def chemist_ordered(fermion_operator):
+    """Puts a two-body fermion operator in chemist ordering.
+
+    The normal ordering convention for chemists is different.
+    Rather than ordering the two-body term as physicists do, as
+    :math: a^\dagger a^\dagger a a
```
kevinsung: It should be :math:`a^\dagger a^\dagger a a`

kevinsung: I mean,
```
:math:`a^\dagger a^\dagger a a`
```

src/openfermion/utils/_operator_utils.py
```diff
@@ -46,9 +46,9 @@ def chemist_ordered(fermion_operator):
 
     The normal ordering convention for chemists is different.
     Rather than ordering the two-body term as physicists do, as
-    :math: a^\dagger a^\dagger a a
+    :math: `a^\dagger a^\dagger a a`
```
kevinsung: No space between :math: and the backtick.

PR #376, opened 2018-06-22
test installed module

loriab:
### in this PR
- [ ] installs a few more files so that one can run pytest on an installation
- [ ] allows some of the heavier (i.e., jupyter, matplotlib) and noncore (i.e., pubchempy) dependencies to be optional by moving the imports and adding proper pytest skipping
- [ ] tested on py27, 35, 36
- [ ] fyi, the setup.py/MANIFEST.in handling of `examples/` works just fine for `python setup.py install` (that is, places the examples dir w/i the module). however, if you `python setup.py install --single-version-externally-managed`, that will place the examples dir into `PREFIX/openfermion/`.

### Q
- [ ] there's some repeated code (`_module_import()`). if you decide you like this PR, could you suggest a good shared access place for it?
- [ ] despite making some dependencies optional, I didn't interfere with `requirements.txt`, as I can control them independently through conda. but if you like this change, should be tweaked.

### longer term Q
- [ ] I tested this by conda packaging it (`conda install -c psi4/label/agg openfermion`, https://anaconda.org/psi4/openfermion/files, https://github.com/psi4/psi4meta/tree/master/conda-recipes/openfermion). Do you want that? If so, from your own channel, conda-forge (a community packaging channel), or psi4?
- [ ] If so (above) and if this PR gets pulled in, should master be packaged or this PR cherry-picked onto v0.6?
- [ ] The whole purpose of this exercise is to test openfermionpsi4 for the upcoming Psi4 1.2 release. What I'm finding for the psi plugin (https://github.com/quantumlib/OpenFermion-Psi4/tree/master/examples) are more in the nature of examples than tests. Are there tests out there?

### DIFF
--- a/src/openfermion/tests/_example_test.py
+++ b/src/openfermion/tests/_example_test.py
@@ -18,11 +18,29 @@ import sys
 import tempfile
 import unittest
 
-import nbformat
 import numpy
+import pytest
 
 from openfermion.config import THIS_DIRECTORY
 
+
+def _module_import(plug):
+    import sys
+    if sys.version_info >= (3, 4):
+        from importlib import util
+        plug_spec = util.find_spec(plug)
+    else:
+        import pkgutil
+        plug_spec = pkgutil.find_loader(plug)
+    if plug_spec is None:
+        print('no', plug)
+        return False
+    else:
+        print('yes', plug)
+        return True
+
+using_nbconvert = pytest.mark.skipif(_module_import('nbconvert') is False,
+                                     reason='Not detecting module nbconvert. Install package if necessary and add to envvar PYTHONPATH')
 
 class ExampleTest(unittest.TestCase):
     """Unit tests for example scripts."""
@@ -34,8 +52,11 @@ class ExampleTest(unittest.TestCase):
         self.binary_code_transforms_demo = 'binary_code_transforms_demo.ipynb'
         self.jw_bk_demo = 'jordan_wigner_and_bravyi_kitaev_transforms.ipynb'
 
+    @using_nbconvert
     def test_demo(self):
         """Unit test for demo."""
+        import nbformat
+
         # Determine if python 2 or 3 is being used.
         major_version, minor_version = sys.version_info[:2]
         if major_version == 2 or minor_version == 6:
@@ -66,8 +87,11 @@ class ExampleTest(unittest.TestCase):
             errors = []
         self.assertEqual(errors, [])
 
+    @using_nbconvert
     def test_binary_code_transforms_demo(self):
         """Unit test for demo."""
+        import nbformat
+
         # Determine if python 2 or 3 is being used.
         major_version, minor_version = sys.version_info[:2]
         if major_version == 2 or minor_version == 6:
@@ -98,8 +122,11 @@ class ExampleTest(unittest.TestCase):
             errors = []
         self.assertEqual(errors, [])
 
+    @using_nbconvert
     def test_jordan_wigner_and_bravyi_kitaev_transforms_demo(self):
         """Unit test for demo."""
+        import nbformat
+
         # Determine if python 2 or 3 is being used.
         major_version, minor_version = sys.version_info[:2]
         if major_version == 2 or minor_version == 6:
--- a/src/openfermion/tests/_example_test.py
+++ b/src/openfermion/tests/_example_test.py
@@ -33,14 +33,14 @@ def _module_import(plug):
         import pkgutil
         plug_spec = pkgutil.find_loader(plug)
     if plug_spec is None:
-        print('no', plug)
         return False
     else:
-        print('yes', plug)
         return True
 
-using_nbconvert = pytest.mark.skipif(_module_import('nbconvert') is False,
-                                     reason='Not detecting module nbconvert. Install package if necessary and add to envvar PYTHONPATH')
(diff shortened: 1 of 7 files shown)


loriab:
I signed it!

loriab:
Sorry, I didn't realize I was springing a PR on you at the end of a release cycle yesterday.

Looking at the coverage fail, I think a decision on optional deps and requirements.txt is needed before I could tweak one of the test matrix into covering the extra lines.

loriab:
Thanks for taking a look, @jarrodmcc. Didn't mean to make a surprise PR at the end of a release cycle.

My goals are modest — I just wanted to check that Psi4 is still working for OpenFermion (we're not great at API/ABI stability and some changes to pybind11 have required rebuilding of downstream projects). Beyond that, if there was an easy way to monitor ongoing compatibility, rather than annual testing, I'd like to set it up. Beyond that, since conda packages are free after the real work of devising tests, if you wanted conda packages built and/or distributed in the `psi4-rt` metapackage (collects optional run-time dependencies like Grimme's `dftd3` and psi4 plugins`), glad to do so.

As far as monitoring ongoing compatibility, on the psi side, I'd like a little smoke test case for [here](https://github.com/psi4/psi4/blob/master/tests/pytest/test_addons.py) so that _if_ openfermion/openfermion-psi4 are present, it'll check that psi4 is still working. On the openfermion-psi4 side, I can well understand the travis hurdles. Something like https://github.com/cdsgroup/resp/blob/master/.travis.yml may be helpful in avoiding a big compilation.

My focus on installed, as opposed to source-dir, testing is the result of trying to test multiple projects and also b/c conda packaging encourages it.

Btw, I'm always glad to rebase when you need it. That may help avoid CLA troubles.

src/openfermion/tests/_example_test.py
```diff
@@ -18,24 +18,52 @@
...
 import unittest
 
-import nbformat
 import numpy
+import pytest
 
 from openfermion.config import THIS_DIRECTORY
 
 
+def _module_import(plug):
```
jarrodmcc: As you asked about in your PR, I think we can put this _module_import helper function into utils/testing_utils.py

src/openfermion/tests/_example_test.py
```diff
@@ -18,24 +18,52 @@
...
+        plug_spec = util.find_spec(plug)
+    else:
+        import pkgutil
+        plug_spec = pkgutil.find_loader(plug)
+    if plug_spec is None:
+        return False
+    else:
+        return True
+
+using_nbtools = pytest.mark.skipif(not (_module_import('nbconvert') and _module_import('ipykernel')),
```
jarrodmcc: Minor style comment, but make sure your files pass pep8 checking (line length in the case of the lines below for the extended strings). It's mostly manual checking at the moment, but we will be converting to enforced pylinter checking soon for PR's

src/openfermion/tests/_example_test.py
jarrodmcc: Also, for possible future users of this function, can you add a docstring and specification for the plug parameter?

loriab:
Thanks for the review. I think that commit patches up the specific code issues.

> Which requirements did you imagine changing to optional based on the current changes?

I think jupyter and matplotlib are clear candidates for optional deps, as they're only used for the tests. I've also set up pubchempy as optional since it's neither part of the core scientific python stack nor core openfermion functionality. But I can see how you'd be more hesitant to classify it optional since the module itself imports it.

> As for conda packages, I think for now we are happy just maintaining the PyPI package, as installation in anaconda using it seems to work reasonably seamlessly as well. We are happy for you to have a separate conda build that you make sure runs with Psi4 as well in your own channel. Would it be helpful for us to include such a conda build file for optional use in this repo?

Sounds good to me, thanks. No need to include a conda recipe file here (unless you'd like to) as I keep all recipes in a [separate repo](https://github.com/psi4/psi4meta/tree/master/conda-recipes/openfermion).

> As for versions, I would recommend pinning it to release versions of 0.7 or greater rather than the Github master at the moment. This will probably keep the interface more stable, though like yourselves we tend to move things around a bit still.

Will do. Last week's release actually solves this problem nicely.

> We could work to add a few more tests before your next release to the OpenFermion-Psi4 plugin, do you have an expected time frame for that?

Next release was due last month :-) I'm finally about to cut `1.2rc3`. So 1.2 will probably not wait on openfermion integration tests. This is part of a general effort (before my ecosystem diagram gets completely unmanagable) to set up testing interlocks so Psi4 can refactor freely.

jarrodmcc:
Thanks for the changes! I'm going to go ahead and merge this in spite of coveralls hanging since the tests are passing and it looks good.

babbush:
@loriab I want to reiterate how much the OpenFermion developers appreciate consideration from the folks at Psi4! Thanks again!

Additionally, it is our policy that anybody with merged PRs on OpenFermion is entitled to be listed as an author in the README / NOTICE files and also on the OpenFermion release paper (https://arxiv.org/abs/1710.07629), which we keep on arXiv so that we can update whenever we want.

If you'd like to be listed as an author on the README / NOTICE, consider opening a PR to add yourself in the alphabetical section. If you'd like to be on the release paper, please email myself or Jarrod or so that we can send you an overleaf link to the paper draft.

PR #377, opened 2018-06-26
Allow taking expectation of LinearOperator and switch HF state to numpy array

kevinsung:


idk3:
What's wrong with the Hartree-Fock state being sparse? This prevents me from making the HF state beyond ~30 qubits.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -703,58 +703,63 @@ def sparse_eigenspectrum(sparse_operator):
...
+    elif isinstance(state, numpy.ndarray):
         # Handle state vector.
-        if scipy.sparse.issparse(state):
-            expectation = (state.getH() * sparse_operator * state)
-        else:
-            expectation = numpy.dot(numpy.conj(state.T),
-                                    sparse_operator.dot(state))
-        if expectation.shape != ():
-            expectation = expectation[0, 0]
+        expectation = numpy.dot(numpy.conjugate(state), operator * state)
```
jarrodmcc: Just to quickly check, is this new way of writing the expectation value compatible with numpy array states that are given as both arrays with dimension d and arrays with dimension (d, 1)?  I think in the latter case this may not yield what one expects, because of the lack of transpose, and even with the transpose will cause inconsistent data types in the return.

src/openfermion/utils/_sparse_tools.py
```diff
@@ -703,58 +703,63 @@ def sparse_eigenspectrum(sparse_operator):
...
     Returns:
-        A real float giving the variance.
+        A complex number giving the variance.
 
     Raises:
         ValueError: Input state has invalid format.
     """
-    return (expectation(sparse_operator ** 2, state) -
-            expectation(sparse_operator, state) ** 2)
+    return (expectation(operator ** 2, state) -
```
jarrodmcc: A thought as I'm reading this is that powers of linear operators should ideally not square the terms in the operator, but rather result in a new linear operator that applies the original twice to a state.  We may not need this optimization now, but if someone starts writing a lot of routines calling higher powers of operators, we will.

src/openfermion/utils/_sparse_tools_test.py
```diff
@@ -812,8 +814,8 @@ def setUp(self):
             hamiltonian, self.n_particles))
         self.hf_state_index1 = numpy.sum(2 ** occupied_states)
 
-        self.hf_state1 = csc_matrix(
-            ([1.0], ([self.hf_state_index1], [0])), shape=(2 ** n_qubits, 1))
+        self.hf_state1 = numpy.zeros(2 ** n_qubits)
```
jarrodmcc: Same thought for d vs (d, 1)

src/openfermion/utils/_sparse_tools.py
kevinsung: It does not work for dimension (d, 1). I can fix that though.

src/openfermion/utils/_sparse_tools.py
kevinsung: Squaring a LinearOperator gives an operator of type
```
<class 'scipy.sparse.linalg.interface._PowerLinearOperator'>
```
I think scipy already performs this optimization, but not sure.

src/openfermion/utils/_sparse_tools_test.py
kevinsung: We are already using the convention `(d,)` rather than `(d, 1)` in `get_ground_state`, `jw_configuration_state`, etc.

src/openfermion/utils/_sparse_tools.py
kevinsung: I verified this by inspecting the source code: https://github.com/scipy/scipy/blob/master/scipy/sparse/linalg/interface.py.

LinearOperators can only be raised to non-negative integer powers.

jarrodmcc: Awesome, didn't realize LinearOperator did that by default

src/openfermion/utils/_sparse_tools.py
```diff
@@ -703,58 +703,69 @@ def sparse_eigenspectrum(sparse_operator):
...
+        if len(state.shape) == 1:
+            # Row vector
+            expectation = numpy.dot(numpy.conjugate(state), operator * state)
         else:
-            expectation = numpy.dot(numpy.conj(state.T),
-                                    sparse_operator.dot(state))
-        if expectation.shape != ():
-            expectation = expectation[0, 0]
+            # Column vector
+            expectation = numpy.dot(numpy.conjugate(state.T),
```
jarrodmcc: Okay i think this looks good now.  As a final note if you wanted to include it before merging, you could probably do both with the transpose (and both would be correct), then you could extract [0,0] in one case and not the other.  But it's fine either way.

PR #378, opened 2018-06-27
Low rank circuits

babbush:
Adding final function that will go in OpenFermion core to help assist with low rank circuit simulation.

### DIFF
--- a/src/openfermion/utils/_low_rank.py
+++ b/src/openfermion/utils/_low_rank.py
@@ -199,7 +199,7 @@ def prepare_one_body_squared_evolution(one_body_matrix):
     """
     # Diagonalize the one-body matrix.
     eigenvalues, eigenvectors = numpy.linalg.eig(one_body_matrix)
-    basis_transformation_matrix = eigenvectors.transpose()
+    basis_transformation_matrix = numpy.conjugate(eigenvectors.transpose())
 
     # Obtain the diagonal two-body matrix.
     density_density_matrix = numpy.outer(eigenvalues, eigenvalues)
--- a/src/openfermion/utils/_low_rank.py
+++ b/src/openfermion/utils/_low_rank.py
@@ -198,7 +198,7 @@ def prepare_one_body_squared_evolution(one_body_matrix):
             storing the values of the basis transformation.
     """
     # Diagonalize the one-body matrix.
-    eigenvalues, eigenvectors = numpy.linalg.eigh(one_body_matrix)
+    eigenvalues, eigenvectors = numpy.linalg.eig(one_body_matrix)
     basis_transformation_matrix = eigenvectors.transpose()
 
     # Obtain the diagonal two-body matrix.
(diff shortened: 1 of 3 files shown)


babbush:
@jarrodmcc I am looking into this now, but an issue missed by tests on my last PR is that the one-body squares output by the low_rank_two_body_decomposition function are not necessarily Hermitian. That makes it less clear how we should diagonalize things.

babbush:
The issue seems to be fixed for an even number of qubits. More substantial tests will be forthcoming once these routines are used in the forthcoming circuit compilation routines.

src/openfermion/utils/_low_rank.py
```diff
@@ -174,3 +174,35 @@ def low_rank_two_body_decomposition(chemist_two_body_coefficients,
...
+
+    Returns:
+        density_density_matrix(ndarray of floats) an N by N array storing
+            the diagonal two-body coefficeints :math:`V_{pq}` above.
+        basis_transformation_matrix (ndarray of floats) an N by N array
+            storing the values of the basis transformation.
+    """
+    # Diagonalize the one-body matrix.
+    eigenvalues, eigenvectors = numpy.linalg.eig(one_body_matrix)
+    basis_transformation_matrix = eigenvectors.transpose()
```
jarrodmcc: What's the convention being used that it's the transpose, but not original or conjugate transpose?  Seems like generally there could be complex eigenvectors with non-hermitian matrices

src/openfermion/utils/_low_rank_test.py
```diff
@@ -190,3 +191,56 @@ def test_rank_reduction(self):
...
+
+    def test_one_body_square_decomposition(self):
+
+        # Initialize a random two-body FermionOperator.
+        n_qubits = 4
+        random_operator = get_fermion_operator(
+            random_interaction_operator(n_qubits))
+
+        # Convert to chemist tensor.
+        constant, one_body_coefficients, chemist_tensor = \
```
jarrodmcc: No backslashes to be consistent

src/openfermion/utils/_low_rank_test.py
```diff
@@ -190,3 +191,56 @@ def test_rank_reduction(self):
...
+            self.assertAlmostEqual(0., numpy.amax(numpy.absolute(difference)))
+            density_density_alternative = numpy.outer(diagonal, diagonal)
+            difference = density_density_alternative - density_density_matrix
+            self.assertAlmostEqual(0., numpy.amax(numpy.absolute(difference)))
+
+            # Test spectra.
+            one_body_squared_spectrum = eigenspectrum(one_body_squared)
+            two_body_spectrum = eigenspectrum(two_body_operator)
+            difference = two_body_spectrum - one_body_squared_spectrum
+            self.assertAlmostEqual(0., numpy.amax(numpy.absolute(difference)))
```
jarrodmcc: If we expect the resulting normal ordered operators to be the same also, maybe we can test that as well? Or is that too onerous?

PR #379, opened 2018-06-28
Add page in docs for collecting projects and papers

jarrodmcc:
Adds a page to collect project and papers using OpenFermion that hopefully both inspire new projects and let people advertise their own work using OpenFermion there.  I left all the links as ArXiv so that people don't need journal descriptions to see the papers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -47,8 +47,8 @@ You might also want to explore the alpha release of the
 `OpenFermion Cloud Library <https://github.com/quantumlib/OpenFermion/tree/master/cloud_library>`__
 where users can share and download precomputed molecular benchmark files.
 
-Check out other `projects and papers using OpenFermion <docs/other_projects.md>`__ for inspiration, and let us know if you've
-been using OpenFermion!
+Check out other `projects and papers using OpenFermion <docs/other_projects.md>`__ for inspiration,
+and let us know if you've been using OpenFermion!
 
 Developer install
 -----------------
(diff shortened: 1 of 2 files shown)


jarrodmcc:
Techniques implemented in OpenFermion seems like a different category, as the paper doesn't actually seem to use it

babbush:
Can we also have a section for papers with techniques implemented in OpenFermion then? The authors worked very hard to incorporate these techniques in OpenFermion and I think it is to the benefit of both papers like that and OpenFermion that we advertise such work as well.

jarrodmcc:
Sure, in a future PR. if I add it now that will be the only paper there

README.rst
```diff
@@ -47,6 +47,8 @@ You might also want to explore the alpha release of the
 `OpenFermion Cloud Library <https://github.com/quantumlib/OpenFermion/tree/master/cloud_library>`__
 where users can share and download precomputed molecular benchmark files.
 
+Check out other `projects and papers using OpenFermion <docs/other_projects.md>`__ for inspiration, and let us know if you've
```
sheilaliuxl: 80 Characters/ line?

jarrodmcc: I didn't see it too strict in other markdown files we've been using but I can update these lines to make sure they are

jarrodmcc:
Also, on thinking about it, feature's implemented in OpenFermion along with their citations is a bit too different to be in the same file.  That's something that should be featured prominently in the docs, the OpenFermion release paper, or both.

PR #381, opened 2018-07-02
Low rank options

babbush:
* Fixes an error where I was performing truncation only based on eigenvalue and failing to take into account the norm of the eigenvector.
* Allows one to truncate by specifying the final rank of the decomposed operator.

### DIFF
--- a/src/openfermion/utils/_low_rank.py
+++ b/src/openfermion/utils/_low_rank.py
@@ -111,14 +111,15 @@ def get_chemist_two_body_coefficients(two_body_operator):
 
 
 def low_rank_two_body_decomposition(chemist_two_body_coefficients,
-                                    truncation_threshold=None):
+                                    truncation_threshold=None,
+                                    final_rank=None):
     """Convert two-body operator into sum of squared one-body operators.
 
     This function decomposes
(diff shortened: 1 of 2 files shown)


src/openfermion/utils/_low_rank.py
```diff
@@ -111,37 +111,42 @@ def get_chemist_two_body_coefficients(two_body_operator):
...
+                                    truncation_threshold=None,
+                                    final_rank=None):
     """Convert two-body operator into sum of squared one-body operators.
 
     This function decomposes
     :math:`\sum_{pqrs} h_{pqrs} a^\dagger_p a_q a^\dagger_r a_s`
     as :math:`\sum_{l} \lambda_l (\sum_{pq} g_{lpq} a^\dagger_p a_q)^2`
     l is truncated to take max value L so that
-    :math:`\sum_{l=0}^{L-1} |\lambda_l| < x`
+    :math:`\sum_{l=0}^{L-1} (\sum_{pq} |g_{lpq}|)^2 |\lambda_l| < x`
```
jarrodmcc: This bound is non-standard enough for this type of truncation it's probably worth mentioning in the draft if we don't already and it works well for predicting errors.

Issue #396, opened 2018-07-15
SymbolicOperator.__init__ fails when given a unicode string


Strilanc:
The type check only allows `str`, not both `unicode` and `str`. Anyone using `from __future__ import unicode_literals` as part of being ready for python 3 will have a hard time initializing the operator.

```
In [1]: type('')
Out[1]: str

In [2]: from __future__ import unicode_literals

In [3]: type('')
Out[3]: unicode
```

Strilanc:
It looks like the way to fix this [is with the libary sixes](https://stackoverflow.com/a/11301392/52239):

```
from six import string_types
isinstance(s, string_types)
```

PR #397, opened 2018-07-15
Fix version-incompatible string checks using six

Strilanc:
Fixes https://github.com/quantumlib/OpenFermion/issues/396

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,3 +7,4 @@ jupyter
 networkx
 matplotlib
 pubchempy
+six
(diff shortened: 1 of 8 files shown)


babbush:
This seems like kind of a silly thing to need to add a dependency for. Is this really the only solution?

Strilanc:
The tricky thing about it is that the literal `unicode` exists in python 2 but not in python 3, so you can't just write `isinstance(x, (str, unicode))`.

We could implement the same thing six does and import it from ourselves:

```python
import sys
string_type = basestring if sys.version_info[0] == 2 else str

...

from openfermion.whatever import string_type
isinstance(x, string_type)
```

babbush:
Since this is only a couple lines I would prefer it to adding a dependency.

PR #398, opened 2018-07-15
Added Haar random function

babbush:
I find myself writing these lines a lot in some demos I'm making so might as well make it a function.

### DIFF
--- a/src/openfermion/utils/_testing_utils.py
+++ b/src/openfermion/utils/_testing_utils.py
@@ -27,8 +27,8 @@ def haar_random_vector(n, seed=None):
     """Generate an n dimensional Haar randomd vector."""
     if seed is not None:
         numpy.random.seed(seed)
-    vector = numpy.random.randn(n).astype(numpy.complex64)
-    vector += 1.j * numpy.random.randn(n).astype(numpy.complex64)
+    vector = numpy.random.randn(n).astype(complex)
+    vector += 1.j * numpy.random.randn(n).astype(complex)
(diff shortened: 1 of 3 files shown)


src/openfermion/utils/_testing_utils.py
```diff
@@ -23,6 +23,16 @@
                              QuadraticHamiltonian)
 
 
+def haar_random_vector(n, seed=None):
+    """Generate an n dimensional Haar randomd vector."""
+    if seed is not None:
+        numpy.random.seed(seed)
+    vector = numpy.random.randn(n).astype(numpy.complex64)
```
kevinsung: Why are you using single-precision? We don't do that anywhere else in the library.

src/openfermion/utils/_testing_utils.py
```diff
@@ -23,6 +23,16 @@
...
+    vector += 1.j * numpy.random.randn(n).astype(numpy.complex64)
+    normalization = numpy.sqrt(vector.dot(numpy.conjugate(vector)))
```
kevinsung: `numpy.linalg.norm(vector)`

babbush:
I don't know. I switched to double precision. However, I don't want to use that norm function because its not very explicit and I'd prefer to use something that I can easily tell what it is doing. "norm" is ambiguous and can mean many things. So I think my way is more readable.

kevinsung:
The word "norm" is ambiguous but `numpy.linalg.norm` is unambiguous. Anyway, I'll approve.

babbush:
It is totally ambiguous. I'd need to look up numpy.linalg.norm to know what it does!

PR #404, opened 2018-07-20
Faster commutators for diagonal Coulomb operators

idk3:
Up to 100x speedup in evaluation of commutators specialized to diagonal Coulomb operators i^ j, i^ i, or i^ j^ i j (with i > j). Actually slightly more general in that it works with arbitrary two-body operators in several cases.

I've tried to comment pretty extensively to explain how and why the logic branches. The commutator evaluation is actually fast enough that the overhead of handling different major cases with different functions (the sheer cost of creating the stack frames) slows overall evaluation down ~15-20%. I think the extra clarity and cleanness outweighs this relatively small disadvantage, but can go the other way if people prefer.


src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,361 @@
...
+    """Compute the commutator of two two-body operators specified by actions.
+
+    Args:
+        diagonal_coulomb_action (tuple): single term of a diagonal Coulomb
+            FermionOperator (i^ j^ i j). Must be in normal-ordered form,
+            i.e. i > j.
+        arbitrary_two_body_action (tuple): arbitrary single term of a two-body
+            FermionOperator, in normal-ordered form, i.e. i^ j^ k l with
+            i > j, k > l.
+        coefficient (complex float): expected coefficient of the commutator.
```
kevinsung: This sentence is confusing. Why is the coefficient `expected`?

idk3:
Added a performance benchmark. The main cases of interest take very long to run (~days) so I've gone with a small demo example (4x4 spinless jellium).

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: This line is trying to convey that the coefficient can flip sign as the terms normal order: it's the coefficient we expect at the outset, but we might end up with its negative.

kevinsung: It's really just the coefficient, though, right? That is, you take the commutator of the two terms, and multiply by this (no negatives). So just call it coefficient.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,361 @@
...
+#   limitations under the License.
+
+"""Faster commutators for two-body operators with diagonal Coulomb terms."""
+
+import warnings
+
+from openfermion import FermionOperator, normal_ordered
+
+
+def commutator_diagonal_coulomb_operators(
```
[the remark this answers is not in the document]
idk3: Done.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,361 @@
...
+                additional.terms[term_a + term_b] = coefficient
+                additional.terms[term_b + term_a] = -coefficient
+                additional = normal_ordered(additional)
+
+                result += additional
+
+    return result
+
+
+def _commutator_two_body_diagonal_with_two_body(
```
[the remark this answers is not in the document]
idk3: Turns out with a little modification, all the methods are the same :P Thanks for forcing me look at this!

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,361 @@
...
+
+    Returns:
+        result: A FermionOperator with the commutator added to it.
+
+    Notes:
+        This function is actually slightly more general than promised, in that
+        it works for *any* two-body operator_b, provided that operator_a has
+        the required form. Similarly, if operator_a is an arbitrary two-body
+        operator and operator_b is a 1-body operator, it will also work.
+        All operators must be normal-ordered regardless.
```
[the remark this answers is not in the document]
idk3: I'll add some more comments in the docs to try to clarify this. My philosophy is "if we can then we should" - raising an error when we could just do the computation anyway (by falling through to slower methods) feels like punishing the user for a mistake. Warning them that they've violated the input promise, and that we're going to use the slower and more general method, is kinder and less disruptive.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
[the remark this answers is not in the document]
idk3: Actually, re: your other comments, I'm renaming it `commutator_ordered_diagonal_coulomb_with_two_body_operator` to properly capture what it does.

examples/performance_benchmarks.py
```diff
@@ -229,12 +283,31 @@ def run_linear_qubit_operator(n_qubits=16, n_terms=10, processes=10):
...
-                 'ratio is %.2f.\n', runtime_sequential, runtime_parallel,
+                 'ParallelQubitOperator took %f seconds with %d processes, '
+                 'and ratio is %.2f.\n', runtime_sequential, runtime_parallel,
                  processes, runtime_sequential / runtime_parallel)
 
     return runtime_sequential, runtime_parallel
 
+
+def run_diagonal_commutator(side_length=4):
+    """Run commutator_diagonal_coulomb_operators benchmark."""
```
babbush: Can you explain a bit more what this benchmark is about?

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,303 @@
...
+    """Compute the commutator of two-body operators provided that both are
+    normal-ordered and that the first only has diagonal Coulomb interactions.
+
+    Args:
+        operator_a: The first FermionOperator argument of the commutator.
+            All terms must be normal-ordered, and furthermore either hopping
+            operators (i^ j) or diagonal Coulomb operators (i^ i or i^ j^ i j).
+        operator_b: The second FermionOperator argument of the commutator.
+            operator_b can be any arbitrary two-body operator.
+        result (optional): The initial FermionOperator to add to.
```
babbush: why is this called "result"?

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,303 @@
...
+                    len(term_a) == 4 and len(term_b) == 2):
+                _commutator_one_body_with_two_body(
+                    term_a, term_b, coefficient, result)
+
+            # Case 3: both terms are one-body operators (both length 2)
+            elif len(term_a) == 2 and len(term_b) == 2:
+                _commutator_one_body_with_one_body(
+                    term_a, term_b, coefficient, result)
+
+            # Final case (case 4): violation of the input promise. Still
```
babbush: I'm a little surprised by the decision to warn the user that their input is out of spec instead of just raise an exception. What's going on with that?

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
```diff
@@ -0,0 +1,303 @@
...
+def _commutator_one_body_with_two_body(action_a, action_b,
+                                       coefficient, result):
+    """Compute commutator of action-specified one- and two-body operators.
+
+    Args:
+        action_a, action_b (tuple): single terms, one one-body and the other
+            two-body, from normal-ordered FermionOperators. It does not matter
+            which is one- or two-body so long as only one of each appears.
+        coefficient (complex float): coefficient of the commutator.
+        result (FermionOperator): the resultant to which to add the commutator.
```
babbush: I still think it is strange there is an input argument named "result". Do you really get a performance advantage from performing the addition within the function?

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: `resultant` maybe? We're (possibly) taking in some other FermionOperator and adding the commutator to it.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: Yeah, see e.g. https://stackoverflow.com/questions/16256913/improving-performance-of-very-large-dictionary-in-python for a bit more info - basically as the underlying dict for result gets bigger and bigger, operations start to slow down a bit. There are some ways around this but they aren't really applicable for us (since we don't have a good idea about the keys ahead of time). And that post is discussing the slowdown for dictionaries with ~100K elements - the commutators here can end up with hundreds of millions of terms, where cutting the time on dictionary operations by a factor 2-3 can make a noticeable difference.

babbush: But how does adding directly to dictionary for the whole system, as opposed to creating a new dictionary that exists just within the scope of this function, help things?

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
babbush: resultant isn't any good either. I think its really weird to have an argument called "result". How about something like "prior_terms".

examples/performance_benchmarks.py
idk3: This just gives a small demo to show the speed difference between `commutator` and the new function. @kevinsung suggested adding one, I think it's nice to see / makes the speedup a bit more tangible. The benchmark splits the terms from a jellium Hamiltonian in two, and computes the commutator of the two parts.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: My philosophy here is that exceptions are painful and should only be used for fatal issues (something where we really *can't* proceed). Here we can still do the computation, and deliver an alert to the user that things won't be as fast because they broke the input promise.

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: I'll change that!

src/openfermion/utils/_commutator_diagonal_coulomb_operator.py
idk3: OK, I see what you're saying. If we compute a Trotter error operator using this function, we're computing a sum of double commutators of terms as they appear in the simulation. Having `prior_terms` means we can add directly to a single FermionOperator instead of having to total a large number of (individually large) FermionOperators. This is where having `prior_terms` reduces the number of operations.

idk3: I ran a small benchmark computing the Trotter error operator for 8x8 spinless jellium. Using the code with the prior result (calling the function with `prior_terms=error_operator`) is a third faster than the identical code using `prior_terms += function()`. So the difference isn't trivial (not just a fraction of a percent ;)) but isn't gigantic either. I'd vote to leave it in, but I'm fine removing it if you think it's much cleaner without.

idk3: Ready for final review (changed to `prior_terms`)! My commit marked this thread outdated but it has the final point for discussion I think - thoughts @babbush @kevinsung ?

babbush: I guess it is fine to leave in. This PR is good to go from my perspective. Just waiting on @kevinsung now.

idk3:
@babbush can you force merge this?

PR #418, opened 2018-07-30
Make SymbolicOperator abstract

### LINKED ISSUE #415: SymbolicOperator should be an abstract base class
See https://docs.python.org/3/library/abc.html.

kevinsung:
Fixes #415 .

### DIFF
--- a/src/openfermion/ops/_boson_operator.py
+++ b/src/openfermion/ops/_boson_operator.py
@@ -11,14 +11,10 @@
 #   limitations under the License.
 
 """BosonOperator stores a sum of products of bosonic ladder operators."""
+
 import numpy
 
-from future.utils import iteritems
-from openfermion.ops import SymbolicOperator
-
-
-class BosonOperatorError(Exception):
-    pass
+from openfermion.ops._symbolic_operator import SymbolicOperator
 
 
 class BosonOperator(SymbolicOperator):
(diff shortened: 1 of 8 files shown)


kevinsung:
@babbush you're going to have to override coveralls on this one. It is complaining about the `pass` statements in the abstract function definitions, which should never get executed.

src/openfermion/ops/_boson_operator.py
```diff
@@ -11,14 +11,8 @@
...
 """BosonOperator stores a sum of products of bosonic ladder operators."""
-import numpy
 
-from future.utils import iteritems
-from openfermion.ops import SymbolicOperator
-
-
-class BosonOperatorError(Exception):
-    pass
+from openfermion.ops._symbolic_operator import SymbolicOperator
```
idk3: Why the shift away from importing from `openfermion.ops`? (No preference, just wondering about the motivation for the change)

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -11,19 +11,19 @@
...
-
+from six import add_metaclass, string_types
 
-class SymbolicOperatorError(Exception):
-    pass
+from openfermion.config import EQ_TOLERANCE
 
 
-class SymbolicOperator(object):
+@add_metaclass(abc.ABCMeta)
```
idk3: Nice!

src/openfermion/ops/_boson_operator.py
kevinsung: Importing directly from `ops` from a file inside `ops` is brittle and can cause circular dependencies. This is the reason we need to put imports out of alphabetical order in `ops/__init__.py`. Doing it my way, we won't need to shuffle around imports in `__init__.py` files.

kevinsung: It's okay to import from `ops` in the test files though. In fact, test files should just import from the top level, `openfermion`, since that is how we should be encouraging users to write their code.

PR #421, opened 2018-08-01
Added iteration to symbolic operator, made unitary rotations function

obriente:
Two small additions that I've found to be of use recently in Openfermion, wanted to share.

The first allows for iteration over the terms in a symbolic operator as an operator of the same type.
The second performs the Pauli rotations discussed in section VII of https://arxiv.org/abs/1807.10050 .

Not sure if either of these exist in some other place in the code, I couldn't find them myself.


src/openfermion/transforms/_unitary_rotations.py
```diff
@@ -0,0 +1,33 @@
...
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+"""Functions for unitary transformations of operators"""
+import numpy
+
+
+def rotate_qubit_by_pauli(qop, pauli, angle):
+    '''
```
babbush: Please add the "args" and "returns" doc strings. I am not sure what type of object "pauli" is.

src/openfermion/transforms/_unitary_rotations.py
```diff
@@ -0,0 +1,33 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
```
babbush: This seems a little bit light to be its own module, unless you plan to build more. Perhaps there is some other place it might fit?

obriente: Perhaps it can be combined with _projection.py ? Both files are used for manipulating QubitOperator objects, although in slightly different manners.

src/openfermion/transforms/_unitary_rotations.py
obriente: Done

src/openfermion/transforms/_unitary_rotations.py
babbush: equality_constraints_projection? I think that's probably not a good place for it. Since this is used for measurements, perhaps I'd feel better if you put it in measurements/_unitary_rotations.py. Does that sound okay?

babbush: Oh okay. Either way works for me. transforms/_projections.py would be okay too.

obriente: Ok, I'll combine this with transforms/_projections.py then.

PR #423, opened 2018-08-04
Add constant property to SymbolicOperator

kevinsung:


kevinsung:
@babbush or @jarrodmcc, need you to override coveralls false positive.

PR #424, opened 2018-08-09
Remove dependence of QuadraticHamiltonian on Givens rotations

### LINKED ISSUE #407: Remove some QuadraticHamiltonian methods
I don't think QuadraticHamiltonian should depend on Givens rotations. I think the only essential method is `diagonalizing_bogoliubov_transform`.

This would be a breaking change.

kevinsung:
Resolves #407 . This removes the method `diagonalizing_circuit` which does work that really should be performed in a circuit compilation library like OpenFermion-Cirq.


babbush:
Actually I think this functionality should remain in OpenFermion somewhere. Basically, anything that can stay in OpenFermion should so that as many people as possible can use it.

babbush:
@kevinsung what is going on with this PR?

kevinsung:
I'll get back to it later.

PR #429, opened 2018-08-14
Diagonalize one-body matrix using eigh and in spatial basis for low rank decomposition

### LINKED ISSUE #427: Basis change matrices from spatial low rank are not unitary
Tested this with LiH. Are they supposed to be unitary?
```
import cirq
import openfermion
from openfermion import low_rank_spatial_two_body_decomposition, prepare_one_body_squared_evolution


def load_molecular_hamiltonian(
        geometry,
        basis,
        multiplicity,
        description,
        n_active_electrons,
        n_active_orbitals):

    molecule = openfermion.MolecularData(
            geometry, basis, multiplicity, description=description)
    molecule.load()

    n_core_orbitals = (molecule.n_electrons - n_active_electrons) // 2
    occupied_indices = list(range(n_core_orbitals))
    active_indices = list(range(n_core_orbitals,
                                n_core_orbitals + n_active_orbitals))

    return molecule.get_molecular_hamiltonian(
            occupied_indices=occupied_indices,
            active_indices=active_indices)


# 4-qubit LiH 2-2 with bond length 1.45
bond_length = 1.45
geometry = [('Li', (0., 0., 0.)), ('H', (0., 0., bond_length))]
hamiltonian = load_molecular_hamiltonian(
        geometry, 'sto-3g', 1, format(bond_length), 2, 2)


# Perform the low rank decomposition of two-body operator.
eigenvalues, one_body_squares, _, one_body_correction = (
    low_rank_spatial_two_body_decomposition(
        hamiltonian.two_body_tensor,
        spin_basis=True))


# Get basis transformation matrices and print whether they are unitary
for j in range(len(eigenvalues)):
    _, basis_change_matrix = (
        prepare_one_body_squared_evolution(one_body_squares[j]))
    print(cirq.is_unitary(basis_change_matrix))
```
Output:
```
False
False
True
False
```

kevinsung:
Fixes #427 .

It's necessary to check whether the matrices are Hermitian (and use eig if they're not) in order to pass the tests of the non-spatial routine.

### DIFF
--- a/src/openfermion/utils/_low_rank.py
+++ b/src/openfermion/utils/_low_rank.py
@@ -19,7 +19,7 @@ import scipy.linalg
 
 from openfermion.config import EQ_TOLERANCE
 from openfermion.ops import FermionOperator, InteractionOperator
-from openfermion.utils import count_qubits, normal_ordered
+from openfermion.utils import count_qubits, is_hermitian, normal_ordered
 
 
 def get_chemist_two_body_coefficients(two_body_operator):
@@ -314,7 +314,7 @@ def low_rank_spatial_two_body_decomposition(two_body_coefficients,
             one_body_correction)
 
 
-def prepare_one_body_squared_evolution(one_body_matrix):
+def prepare_one_body_squared_evolution(one_body_matrix, spin_basis=True):
     """Get Givens angles and DiagonalHamiltonian to simulate squared one-body.
 
     The goal here will be to prepare to simulate evolution under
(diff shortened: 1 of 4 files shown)


src/openfermion/utils/_low_rank.py
```diff
@@ -335,10 +335,27 @@ def prepare_one_body_squared_evolution(one_body_matrix):
         basis_transformation_matrix (ndarray of floats) an N by N array
```
jarrodmcc: Add a doc line for spin_basis if it's not already there for this function.

kevinsung: Done. That staled your review though.

kevinsung:
The status check is stuck ... this has been happening a lot recently.

PR #432, opened 2018-08-14
Wrapper function for loading a molecular Hamiltonian

kevinsung:
This reduces boilerplate for situations in which you just want to get an InteractionOperator.

@jarrodmcc do you think the way of specifying the active space (number of electrons, number of active orbitals) is reasonable?


kevinsung:
We need to override coveralls here again.

kevinsung:
Looks like you're going to have to force it again.

PR #434, opened 2018-08-15
removed non-spatial low rank code

babbush:
* Removes non-spatial low rank code.
* Fixes testing utility that was generating InteractionOperators without 8-fold symmetry.

### DIFF
--- a/src/openfermion/utils/_low_rank.py
+++ b/src/openfermion/utils/_low_rank.py
@@ -22,194 +22,79 @@ from openfermion.ops import FermionOperator, InteractionOperator
 from openfermion.utils import count_qubits, is_hermitian, normal_ordered
 
 
-def get_chemist_two_body_coefficients(two_body_operator):
-    """Convert a FermionOperator or InteractionOperator to low rank tensor.
+def get_chemist_two_body_coefficients(two_body_coefficients, spin_basis=True):
+    """Convert two-body operator coefficients to low rank tensor.
 
-    In the chemistry convention the Hamiltonian should be expressed as
-    :math:`\sum_{pqrs} h_{pqrs} a^\dagger_p a_q a^\dagger_r a_s`
-    Furthermore, we will require that all symmetries are unpacked as
-    :math:`p^\dagger q^\dagger r s =
-    (1/4)(-p^\dagger r q^\dagger s + q^\dagger r p^\dagger s
-          + p^\dagger s q^\dagger r - q^\dagger s p^\dagger r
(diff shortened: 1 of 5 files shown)


src/openfermion/utils/_low_rank.py
```diff
@@ -22,194 +22,79 @@
...
+    return one_body_correction, chemist_two_body_coefficients
 
 
-def low_rank_spatial_two_body_decomposition(two_body_coefficients,
-                                            truncation_threshold=None,
-                                            final_rank=None,
-                                            spin_basis=True):
-    """Convert spatial two-body operator into sum of squared one-body spin ops.
+def low_rank_two_body_decomposition(two_body_coefficients,
+                                    truncation_threshold=None,
```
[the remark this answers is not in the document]
kevinsung: I guess the correct behavior then is that `final_rank` specifies an upper bound on the final rank, not the exact one. Then, having both specified doesn't raise an error.

src/openfermion/utils/_low_rank.py
```diff
@@ -158,17 +158,20 @@ def low_rank_two_body_decomposition(two_body_coefficients,
...
+            max_rank = final_rank
     else:
-        raise ValueError(
-            'Cannot provide both final_rank and truncation_value.')
+        if final_rank is None:
+            max_rank = 1 + numpy.argmax(
+                truncation_errors <= truncation_threshold)
+        else:
+            max_rank = min(1 + numpy.argmax(
+                truncation_errors <= truncation_threshold), final_rank)
```
kevinsung: Let's just enforce that `truncation_threshold` must be a float. I don't see a reason to allow passing `None`.

src/openfermion/utils/_low_rank.py
```diff
@@ -296,21 +156,16 @@ def low_rank_spatial_two_body_decomposition(two_body_coefficients,
...
-        max_rank = final_rank
-    elif final_rank is None:
-        max_rank = 1 + numpy.argmax(truncation_errors <= truncation_threshold)
+    if final_rank is None:
+        max_rank = 1 + numpy.argmax(
+            truncation_errors <= truncation_threshold)
     else:
-        raise ValueError(
-            'Cannot provide both final_rank and truncation_value.')
+        max_rank = final_rank
```
kevinsung: I don't think this is the correct behavior. The truncation threshold should determine which eigenvalues are even considered. Then, `final_rank` determines whether things are further truncated. Really, `final_rank` should be renamed `max_rank`.

PR #439, opened 2018-08-24
Masking + caching for evaluating errors for different Trotter steps

idk3:
Computes Trotter error operators for both the split-operator and fermionic swap network Trotter steps.

The fermionic swap network functionality covers the same use cases as in https://github.com/quantumlib/OpenFermion/blob/master/src/openfermion/utils/_low_depth_trotter_error.py, though the new code is quite a bit faster. That code is much simpler, so I'm not sure we should replace it outright, though we probably could. The new split-operator code uses similar tricks to get a comparable speedup, though evaluating the split-operator Trotter error should be easier in general since there are effectively only two terms in the Trotter ordering.

The highest speedup I've seen so far in testing is ~200x for the 32x32 Hubbard model.

### DIFF
--- a/src/openfermion/utils/_diagonal_coulomb_trotter_error.py
+++ b/src/openfermion/utils/_diagonal_coulomb_trotter_error.py
@@ -1,0 +1,297 @@
+import numpy
+
+from future.utils import iteritems
+
+from openfermion import count_qubits, FermionOperator
+from openfermion.utils._low_depth_trotter_error import (
+    simulation_ordered_grouped_low_depth_terms_with_info)
+from openfermion.utils._commutator_diagonal_coulomb_operator import (
+    commutator_ordered_diagonal_coulomb_with_two_body_operator)
+
+
+def potential_and_kinetic_terms_as_arrays(hamiltonian):
+    potential = FermionOperator.zero()
+    kinetic = FermionOperator.zero()
+
+    for term, coeff in iteritems(hamiltonian.terms):
+        acted = set(term[i][0] for i in range(len(term)))
+        if len(acted) == len(term) / 2:
+            potential += FermionOperator(term, coeff)
+        else:
+            kinetic += FermionOperator(term, coeff)
+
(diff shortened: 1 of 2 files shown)


idk3:
@babbush Coveralls is being very strict - the PR is blocked because one line guarding against a Python 3.4 + numpy 1.10/1.11 interaction (see https://travis-ci.org/quantumlib/OpenFermion/builds/420050409) isn't getting hit. Could you override?

src/openfermion/utils/_diagonal_coulomb_trotter_error.py
```diff
@@ -0,0 +1,290 @@
...
+
+    Args:
+        hamiltonian (FermionOperator): The diagonal Coulomb Hamiltonian to
+                                       compute the Trotter error for.
+
+    Returns:
+        error_operator: The second-order Trotter error operator.
+
+    Notes:
+        Follows Equation 9 of Poulin et al.'s work in "The Trotter Step
```
babbush: Just give the arXiv references instead of listing the authors and name of paper.

idk3: Done.

src/openfermion/utils/_diagonal_coulomb_trotter_error.py
```diff
@@ -0,0 +1,290 @@
...
+from openfermion.utils._commutator_diagonal_coulomb_operator import (
+    commutator_ordered_diagonal_coulomb_with_two_body_operator)
+
+
+def diagonal_coulomb_potential_and_kinetic_terms_as_arrays(hamiltonian):
+    """Give the potential and kinetic terms of a diagonal Coulomb Hamiltonian
+    as arrays.
+
+    Args:
+        hamiltonian (FermionOperator): The diagonal Coulomb Hamiltonian to
```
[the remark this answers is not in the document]
idk3: Modified so it can take both.

PR #441, opened 2018-08-31
Update Docker image

kevinsung:
- Replaces ProjectQ and OpenFermion-ProjectQ with Cirq and OpenFermion-Cirq
- Installs quantumlib packages from PyPI rather than the GitHub repos
- Installs Python 3 instead of Python 2
- Installs Python 3 version of Psi4 instead of Python 2 version
- Updates the README


kevinsung:
@babbush clearly coveralls should be overridden here.

PR #442, opened 2018-09-03
Improve script for testing examples

kevinsung:
Automatically tests all Jupyter notebooks added to the `examples/` folder and removes the need for the testing cell at the top of our current notebooks. Ported from OpenFermion-Cirq. Also re-ran the notebooks.


babbush:
There seem to be some python 2 errors here.

PR #463, opened 2018-09-21
Automatically detect spin symmetry in Bogoliubov transformation matrix

kevinsung:
If the quadratic Hamiltonian is block diagonal with two equal-sized blocks, then it will diagonalize each block separately. Note that in this case, the orbital energies will be sorted in ascending order within each block (rather than being all being sorted together).


kevinsung:
@babbush please override coveralls false positive.

Issue #360, opened 2018-06-07
Adding Bayesian estimator for QPE


obriente:
Over the next few weeks I'll be adding some code to do Bayesian estimation for quantum phase estimation. This is part of a paper in production, along the lines of https://arxiv.org/abs/1508.00869 but with multiple qubits and experimental error.

If anyone wants to help with checking + approving code that would be much appreciated!

Cheers,

Tom.

babbush:
@jarrodmcc and I have spoken with @obriente about this and we're looking forward to the PR!

PR #466, opened 2018-09-26
Made Bayesian and time series estimators

obriente:
Two months after saying that this would be done in a week (#360), I've now written up implementations of two estimators for QPE, namely a Bayesian estimator and a new estimator to appear on the arxiv tonight. Both estimators work very well in the absence of noise and can be flagged to adapt to a depolarizing channel (this works for other error channels, but not as well).

I'll update this pull request with the arxiv link in a few hours, but I figured I'd upload the code now seeing as my tests are passing. It's about 1000 lines of code, if people want I can provide some scripts to show some example uses.

### DIFF
--- a/src/openfermion/measurements/__init__.py
+++ b/src/openfermion/measurements/__init__.py
@@ -17,3 +17,9 @@ from ._equality_constraint_projection import (apply_constraints,
 
 from ._rdm_equality_constraints import (one_body_fermion_constraints,
                                         two_body_fermion_constraints)
+
+from ._qpe_estimators import (ProbabilityDist,
+                              BayesEstimator,
+                              BayesDepolarizingEstimator,
+                              TimeSeriesEstimator,
+                              TimeSeriesMultiRoundEstimator)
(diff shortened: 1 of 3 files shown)


obriente:
I'll fix up the coverage issue in the next few days - there are a few edge cases in the Bayesian estimator that are tricky to trigger in testing.

babbush:
this looks like a great PR and we're happy to have it. However, this PR is somewhat larger than we usually like to review at once. Is it perhaps possible to break it up into smaller PRs? For instance, perhaps you could have one PR be about the Bayesian stuff and the other be about the time series stuff? Reviewing becomes really unruly when the PRs are this big.

obriente:
Can do - the Bayesian and the Time series estimator are completely separate. Do you want them in two separate files, or should I do one now and add the second after?

babbush:
Please make two entirely different PRs

babbush:
It doesn't really matter if its one after the other or not.

PR #467, opened 2018-09-26
Time series estimator

obriente:
Here's the time-series estimator split from #466.


obriente:
Just linted the code following discussions on the other thread. Currently pylint is raising a set of complaints that I'm a) setting class variables outside of the init statement, and b) not using long enough variable names in the functions TimeSeriesMultiRoundEstimator._gen_comb, TimeSeriesMultiRoundEstimator._p_function and TimeSeriesMultiRoundEstimator._update_function. My problems with updating these issues are below, perhaps someone else can look and tell me whether the current status is acceptable or not.

For a), this is slightly annoying to fix, because the way the data is stored in the time-series estimator differs by whether it's single round or multi-round. So I'd probably have to split TimeSeriesMultiRoundEstimator to not be a sub-class of TimeSeriesEstimator (which means there'd be a fair few lines of duplicate code lying around).

For b), I don't really have better variable names than the single letters used in the function: these functions are just implementing Eq.25 of arxiv:1809.09697, and the variables are just all dummy variables.

src/openfermion/measurements/_time_series_estimators.py
```diff
@@ -0,0 +1,460 @@
...
+        eigenvalue U|phi_j>=e^{i*phi_j}|phi_j>.
+    '''
+    def __init__(self,
+                 max_experiment_length,
+                 num_freqs_max,
+                 depolarizing_noise=False,
+                 singular_values_cutoff=1e-12,
+                 store_history=False,
+                 automatic_average=False):
+        '''
```
kevinsung: Use double quotes instead of single quotes.

src/openfermion/measurements/_time_series_estimators.py
```diff
@@ -0,0 +1,460 @@
...
+        self._update_flag = False
+
+        if self._automatic_average is True:
+            self.estimate()
+
+        return True
+
+    def update_mass(self, f_data):
+        '''
+        Takes data to update g(k) in the form  The preferable form to input
```
kevinsung: `in the form` ... ?

src/openfermion/measurements/_time_series_estimators.py
```diff
@@ -0,0 +1,460 @@
...
+    def __init__(self,
+                 max_experiment_length,
+                 num_freqs_max,
+                 depolarizing_noise=False,
+                 singular_values_cutoff=1e-12,
+                 store_history=False,
+                 automatic_average=False):
+        '''
+        Args:
+            max_experiment_length: the largest number (K) of unitary operations
```
kevinsung: Document the type of the arguments in parentheses; for example, `max_experiment_length (int)`. This applies to the rest of the PR too.

src/openfermion/measurements/_time_series_estimators.py
```diff
@@ -0,0 +1,460 @@
...
+import scipy as sp
+
+
+class TimeSeriesEstimator(object):
+    '''
+    Time-series estimator: a QPE estimator based on a time-series
+        analysis of g(k)=sum_jA_je^{i*k*phi_j}, where
+        A_j=|<phi_j|Psi>|^2 is the probabilistic overlap
+        with the starting state |Psi>, and phi_j is the unitary
+        eigenvalue U|phi_j>=e^{i*phi_j}|phi_j>.
```
kevinsung: This docstring needs more detail. It should provide context which seems to be assumed in the rest of the docstrings (i.e., what is "the experiment"?). It should also explain how to use the class, ideally including an example.

src/openfermion/measurements/_time_series_estimators.py
obriente: Thanks for checking through the code. I think I've got all of these on both of the PRs.

src/openfermion/measurements/_time_series_estimators.py
obriente: Good catch, I'd left this hanging and described a few lines below instead (with the definition of f_data). I've just removed this comment now.

src/openfermion/measurements/_time_series_estimators.py
obriente: I think I've got all of these.

src/openfermion/measurements/_time_series_estimators.py
obriente: I've updated this significantly now and included a usage case. Hope it makes sense.

obriente:
I've refactored the code, so this can be closed

PR #468, opened 2018-09-26
Bayesian estimator

obriente:
Here's the Bayesian estimator split from #466.


src/openfermion/measurements/__init__.py
```diff
@@ -17,3 +17,7 @@
 
 from ._rdm_equality_constraints import (one_body_fermion_constraints,
                                         two_body_fermion_constraints)
+
+from ._bayesian_estimators import (ProbabilityDist,
```
babbush: alphabetical imports

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+
+'''
+Classes for estimators for quantum phase estimation. Estimators described
+in arXiv:XXXX.XXXX
```
babbush: you can update this now I guess

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+'''
+
+import numpy as np
```
babbush: As annoying as this is, our style guide does not allow "import _ as _". Please just write out numpy.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+from numpy import cos, sin, pi
+from scipy import sparse, optimize
+import warnings
+from openfermion.config import EQ_TOLERANCE
+
+
+class ProbabilityDist(object):
```
babbush: This seems like a really, really generic name for a class designed to represent a fairly specific type of probability distribution. Can you think of a more specific name?

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+
+        # Store size data
+        self._num_vectors = num_vectors
+        self._num_freqs = num_freqs
+
+        # Store prior information
+        if vector_guess is not None and (
+                num_vectors != vector_guess.shape[1]
+                or 2*num_freqs+1 != vector_guess.shape[0]):
+            raise ValueError('''Prior distribution of phases has shape {},
```
babbush: please add
`Raises:
    ValueError: Prior distribution...`
to the doc strings (in the same style you do for 'Args')

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+                required shape is {}'''.format(
```
babbush: Also, why are there 3 apostrophes? Typically we only do that for main doc strings.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+        # Make sparse matrices
+        cos_matrix = sparse.csr_matrix(
+            (data_cos, (row_ind_cos, col_ind_cos)))
+        sin_matrix = sparse.csr_matrix(
+            (data_sin, (row_ind_sin, col_ind_sin)))
+
+        return cos_matrix, sin_matrix
+
+    def _vector_product(self, vectors, round_data):
+
```
babbush: extra space here. Please run pylint.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+    def _Holevo_centers(self, vectors=None):
+        '''
+        Gets the Holevo average phase from a distribution.
+        Args:
+            vectors: vectors to calculate the average of. If None,
+                defaults to the vectors stored in the class.
+        Returns:
+            centers: the corresponding average angle of the
+                distribution.
+
```
babbush: extra space - please run pylint.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+            # We generate the matrices ourselves in
+            # COO format, and then create a
+            # CSR matrix directly
+
+            cos_matrix, sin_matrix = self._make_matrix(n)
+
+            # Append to list
+            self._matrices.append([cos_matrix, sin_matrix])
+
+    def _make_matrix(self, n):
```
babbush: can you give some more explanation of what this is doing? Perhaps refer to some equations from your paper or something? Its pretty difficult to tell.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,774 @@
...
+    def _single_diff(self, p_vec, a_vec):
+        # The first derivative of a single term in the
+        # likelihood function.
+        return -p_vec / np.dot(p_vec, a_vec[:len(p_vec)])
+
+    def _jacobian_term(self, p_vec, a_vec):
+        # The second derivative of the above function,
+        # evaluated for a single p vector - to evaluate
+        # and store
+        return np.dot(p_vec[:, np.newaxis], p_vec[np.newaxis, :]) /\
```
babbush: this method of breaking lines is definitely going to cause problems with the linter. We really should get automatic linting set up... please just lint the entire PR.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Good point. How does FourierProbabilityDist sound? It's a Fourier representation of a probability distribution on [-pi,pi].

obriente:
Sorry about the linting issue, I've been running flake8 on sublime text and it seems that it lints things slightly differently. I've run everything through pylint now - I seem to have fixed all of the problems (except for the number of arguments/attributes in the classes). Let me reply to some of the other above comments and then commit.

src/openfermion/measurements/_bayesian_estimators.py
obriente: I've added an explicit reference to where this can be found in the paper (and hopefully made the docstring a bit clearer)

src/openfermion/measurements/_bayesian_estimators.py
obriente: Odd, it hasn't done it for me. Would you prefer that I bracket everything?

src/openfermion/measurements/_bayesian_estimators.py
obriente: Pylint didn't catch these spaces suprisingly. I think i've found all of them myself though.

src/openfermion/measurements/_bayesian_estimators.py
obriente: I'm doing this so that I can wrap the line around. Pylint didn't raise any flags here, but if you'd prefer some other format I'm happy to change it.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Done

src/openfermion/measurements/_bayesian_estimators.py
obriente: Done

src/openfermion/measurements/_bayesian_estimators.py
obriente: Done

src/openfermion/measurements/__init__.py
obriente: Done

obriente:
Re: Demos, it's not that hard at all to be honest. The data can just be generated by choosing random eigenvalues and sampling from Eq.10 of the paper. Would you prefer separate notebooks for this and the time series estimator, or can I just combine the two?

babbush:
Ignore the CLA bot

obriente:
Rebased the branch to fix the typo in #473 - seems that this has given an issue with the cla bot. Not sure if this was the easiest way forward, sorry if it's caused more work.

obriente:
I've added a demo to the examples folder, hopefully it makes sense.

kevinsung:
The example notebook contains the error
```
/home/tom/packages/openfermion/src/openfermion/measurements/_bayesian_estimators.py:637: RuntimeWarning: divide by zero encountered in true_divide
  numpy.abs(d_amp / self._amplitude_estimates))
```
Also, update to the latest version of h5py (>= 2.8.0) to get rid of the warning
```
/home/tom/rootenv/lib/python3.6/site-packages/h5py/__init__.py:36: FutureWarning: Conversion of the second argument of issubdtype from `float` to `np.floating` is deprecated. In future, it will be treated as `np.float64 == np.dtype(float).type`.
  from ._conv import register_converters as _register_converters
```

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,796 @@
...
+        self.num_dsets = 0
+
+    def update(self, experiment_data, force_accept=False):
+        '''
+        Performs the expected Bayesian update, and stores result.
+
+        Args:
+            experiment_data: a list of dictionaries, containing
+                the data from each round of the experiment. Each round requires
+                entries: 'num_rotations', 'measurement', and 'final_rotation'.
```
kevinsung: If `experiment_data` is only ever expected to have these three keys, then using a dictionary is overkill. Just use a tuple or collections.namedtuple instead.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,796 @@
...
+        if num_vectors != len(amplitude_guess):
+            raise ValueError('''I need {} amplitudes, but you have
+                given me {}.'''.format(num_vectors, len(amplitude_guess)))
+
+        self._amplitude_guess = amplitude_guess
+        self._amplitude_vars = amplitude_vars
+
+        if any([numpy.abs(self._amplitude_guess[j] -
+                          self._amplitude_guess[j+1]) < EQ_TOLERANCE
+                for j in range(num_vectors - 1)]):
```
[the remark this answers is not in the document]
obriente: Thanks for taking a look at the code Kevin! Change made.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Experiment_data may also have a fourth key ('true_measurement'), and I can imagine adding other keys later on. I figured using a dict makes compatibility/usability easier, if you'd prefer I can change it though.

obriente:
I've fixed the divide by zero error, not entirely sure if this fixes the problem being seen in the travis-ci build? The CI is reporting a weird error when it tries to import numpy (as well as a 2-3 compatibility issue that I've now fixed).

src/openfermion/measurements/_bayesian_estimators.py
kevinsung: Make a class to represent this experiment data concept. Instead of passing dictionaries, pass instances of that class.

obriente: Alright, I've put something together. I've split it into a new file as these can be used for the time-series estimator as well.

src/openfermion/measurements/_data_containers.py
```diff
@@ -0,0 +1,99 @@
...
+                 final_rotation,
+                 measurement,
+                 true_measurement=None):
+        """
+        Args:
+            nrot (int): the number of rotations performed in this round
+            finrot (float): the final rotation performed on the ancilla qubit
+                in this round.
+            msmt (bool): the measurement observed in this round
+            tmsmt (bool): if the ancilla is not reset between rounds, the
```
kevinsung: These should be the actual parameter names (`num_rotations` instead of `nrot`)

src/openfermion/measurements/_data_containers.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class QPERoundData(object):
+    """ Results from a single round in a QPE experiment.
+    Attributes:
+        nrot (int): the number of rotations performed in this round
+        finrot (float): the final rotation performed on the ancilla qubit
+            in this round.
+        msmt (bool): the measurement observed in this round
+        tmsmt (bool): if the ancilla is not reset between rounds, the
```
kevinsung: These should be the actual attribute names (num_rotations instead of nrot)

src/openfermion/measurements/_data_containers.py
```diff
@@ -0,0 +1,99 @@
...
+            'true' measurement (in this case msmt is calculated as the
+            difference between the tmsmt of different rounds).
```
kevinsung: I find the explanation of `true_measurement` to be confusing. Please document here what a QPE experiment is. If you did that somewhere else, you can move it to here.

src/openfermion/measurements/_data_containers.py
```diff
@@ -0,0 +1,99 @@
...
+            list_final_rotation=None,
+            list_measurement=None,
+            list_true_measurement=None):
+        """
+        Initialization can be performed in one of two ways:
+        as stated in the arguments
+
+        Args (accepted in order):
+            rounds: list of RoundData objects or lists of dicts
+                to be converted to RoundData objects.
```
kevinsung: Just limit this to being a list of RoundData objects; don't accept a list of dicts.

src/openfermion/measurements/_data_containers.py
```diff
@@ -0,0 +1,99 @@
...
+
+            list_num_rotations (list of integers),
+            list_final_rotation (list of floats),
+            list_measurement (list of booleans),
+            tmsmt_list(optional) (list of booleans):
+                experimental parameters (see RoundData class for description).
```
kevinsung: Just have one way to initialize: a list of RoundData objects. Don't let it be a list of dicts.

src/openfermion/measurements/_data_containers.py
obriente: Fixed, thanks for spotting.

src/openfermion/measurements/_data_containers.py
obriente: I've updated with a fairly lengthy description, hope it makes sense.

src/openfermion/measurements/_data_containers.py
obriente: Can do.

kevinsung: I also mean to get rid of the arguments `list_num_rotations`, etc. so there is only one way to initialize.

obriente: Sorry about that, fixed now.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,805 @@
...
+            (float) P(angle)
+        """
+        res = vector[0]
+        for j in range(1, self._num_freqs):
+            res += vector[2*j-1] * sin(j*angle)
+            res += vector[2*j] * cos(j*angle)
+        return res / (2*pi)
+
+
+class BayesEstimator(FourierProbabilityDist):
```
kevinsung: I'm not sure about having BayesEstimator be a subclass of FourierProbabilityDist. It seems to be mixing together two quite distinct concepts: the concept of a probability distribution, and the concept of something which maintains a probability distribution internally and uses it to estimate values of things. I think it makes more sense for BayesEstimator to keep an instance of FourierProbabilityDist, which it updates as it gets more information and which it uses to obtain phase estimates. What do you think?

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,805 @@
...
+            for j in range(self._num_vectors):
+                self._fourier_vectors[0, j] = 1
+
+        # Initialize p_vec with some information to break the symmetry
+        self._p_vecs = [numpy.array(self._amplitude_guess)]
+        self._amplitude_estimates = numpy.array(self._amplitude_guess)
+        self._inv_covar_mat = numpy.linalg.inv(
+            numpy.array(self._amplitude_vars))
+
+    def get_real_dist(self, x_vec=numpy.linspace(-pi, pi, 101)):
```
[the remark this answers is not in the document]
obriente: Thanks for finding this; I thought I'd gotten rid of any mutable defaults, bad habit of mine that I'm trying to fix.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,805 @@
...
+                 num_freqs=1000,
+                 max_n=1,
+                 vector_guess=None):
+        """
+        Args:
+            amplitude_guess (numpy array or list):
+                estimates for amplitudes of different
+                eigenstates in the initial state.
+            amplitude_vars (numpy array or list): variance on the amplitude estimates
+            num_vectors (int): number of phases to estimate
```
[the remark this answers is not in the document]
obriente: Done. I'm now making reference in the __init__ docstring to variables defined in the class docstring; is this reasonable or should it be avoided?

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,805 @@
...
+            num_freqs (int): number of frequencies in Fourier representation.
+            max_n (int): maximum number of unitary evolutions before measurement.
+                dictates how many matrices for updates are made and stored.
+            vector_guess (numpy array): a prior estimate of the phases. If none,
+                assumes flat.
```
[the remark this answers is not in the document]
obriente: Done

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,805 @@
...
+            self._fourier_vectors = numpy.zeros((self._num_freqs*2+1,
+                                                 self._num_vectors))
+            for j in range(self._num_vectors):
+                self._fourier_vectors[0, j] = 1
+
+        # Initialize p_vec with some information to break the symmetry
+        self._p_vecs = [numpy.array(self._amplitude_guess)]
+        self._amplitude_estimates = numpy.array(self._amplitude_guess)
+        self._inv_covar_mat = numpy.linalg.inv(
+            numpy.array(self._amplitude_vars))
```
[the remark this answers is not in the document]
obriente: Let me respond to this with your latter question, as it's a good point to bring up.

src/openfermion/measurements/_bayesian_estimators.py
obriente: tl;dr: I agree with both this and the above change, but looking through this I think there's a more broad shuffle of the contents of the two classes needed. I'll implement this now, let me know what you think.

So, to make my response more general (and respond to your above comment at the same time), I think that the interface between these two classes is a really good idea to discuss. These started off as a single class, and I split them into two at some point to try to keep track of everything, but I'm not convinced I got the split correct.

I guess there are a few different parts to this; there's the data stored just for the actual probability distribution ( _fourier_vectors, _amplitude_estimates and _amplitude_vars), there's the functions to do simple arithmetic operations on them (_make_matrices, _vector_product, _make_dist, _holevo_centers, _holevo_variances and their dependencies). Then there's the data structures that store additional data from a series of experiments (_p_vecs), and the functions that update a probability distribution given more experimental data (the update function, but also all the functions from _mlikelihood through to _make_projector).

After writing this it seems like a pretty clear distinction; everything that isn't involved directly with _fourier_vectors, _amplitude_estimates or _amplitude_vars should be chucked out of the ProbabilityDist function, and then this can be indeed treated as a separate object. In which case I might want to go about minimizing the interface between the two (but this is probably a good thing to do).

obriente: I've implemented the changes, but I feel like the dataflow between the two classes is quite large as all the manipulation routines are now in the BayesEstimator class.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -23,9 +23,21 @@
 
 class FourierProbabilityDist(object):
     """
-    Stores a estimate of multiple phases in the form arXiv:1809.09697 App.C.
-    In particular, stores the Fourier representation as a sum of sine and
-    cosine waves (Eq. (C1)), and provides methods for manipulation of these.
+    Stores a multivariant Fourier representation of a function linear
+    in its dependent variables.
```
kevinsung: This class is only meant to represent linear functions? In principle it can represent any periodic function, right?

kevinsung: I think you mean that the representation is linear, and not the function; there should be a better way to word this.

src/openfermion/measurements/_bayesian_estimators.py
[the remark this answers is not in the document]
kevinsung: That's fine.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Yeah, I mean it's linear in the sense that it's a sum of a periodic function in each phi_j. Let me think about how to describe this better.

obriente: I think just 'periodic function' works well; the exact definition is one line below, so I don't need to be so descriptive here.

src/openfermion/measurements/_data_containers.py
obriente: Ooh, I missed this. Fixing now.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,845 @@
...
+                the estimated function.
+            amplitude_vars (numpy array or list): variance on the
+                guess in amplitude_guess.
+            num_vectors (int): number of phases phi_j to estimate
+            num_freqs (int): number of wave components cos(n*phi_j)
+                to store coefficients for (this provides a cut-off to
+                the sensitivity of the function).
+            max_n (int): the maximum n to be used when multiplying
+                the stored function by matrices of the form
+                cos(n*phi_j + beta).
```
kevinsung: This is unclear. What is "the maximum n"?

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,845 @@
...
+            raise ValueError('''Starting amplitudes must be different.''')
+
+        # Store max_n
+        self._max_n = max_n
+
+        # Create matrices for multiplication
+        self._make_matrices()
+        self.reset()
+
+    def reset(self):
```
kevinsung: I thought we were getting rid of this method?

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,845 @@
...
+
+    def get_real_dist(self, x_vec=None):
+        """
+        Generates the real distribution over a sequence of points
+
+        Args:
+            x_vec (numpy array or list): the points at which the distribution is generated.
+                If None, defaults to 101 points between -pi and pi.
+        Returns:
+            y_vecs (list): the value of the distributions at the given x points.
```
kevinsung: Return a numpy array instead and explicitly document that it is two dimensional, and what the rows and columns are.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Sorry, which method?

src/openfermion/measurements/_bayesian_estimators.py
obriente: So here I mean 'assuming the user wants to multiply their distributions by cos(n*phi_j + beta), what is the maximum n that they require'. I guess 'the maximum n to be used when multiplying...' -> 'the maximum n required for multiplying...'

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,845 @@
...
+    It also provides routines for multiplying f by functions of the form
+    cos^2(n*phi_j + beta), and for calculating maximum-likelihood
+    distributions of the A_j variables.
+
+    More details can be found in arXiv:1809.09697, Appendix C.
+    """
+
+    def __init__(self,
+                 amplitude_guess=None,
+                 amplitude_vars=None,
```
kevinsung: Since this just represents a probability distribution, it doesn't make sense to talk about guesses and their variances. You should get rid of the `amplitude_vars` variable here; it doesn't look like you are using it anywhere anyway. `amplitude_guess` should be renamed because it's not a guess; it's just a specification of the coefficients which are part of the description of the probability distribution.

obriente: So the variance is definitely a part of the distribution rather than a part of the estimator, and even though it's not currently used here it could well be later. I agree though that the word 'guess' here might be ill-defined. The reason why 'vector_guess' and 'amplitude_guess' exist is to allow for a reset function. This in turn is because creating these matrices is a fairly expensive procedure, so I'd rather not do that every time I want to perform an estimation.

Perhaps the solution is to pull _matrices out of FourierProbabilityDist? But then BayesEstimator is getting quite large.

src/openfermion/measurements/_bayesian_estimators.py
kevinsung: `reset`

obriente: Oh, I hadn't planned on doing so. The main reason why I want to have the reset option is to prevent having to remake _matrices every time the user does a new estimation, as this is quite costly. Which I guess is just heading towards the idea of 'separate the class and the functions which act on the class', but it doesn't feel completely right to have these functions inside the estimator itself...

Perhaps I should make a third class containing these matrices? Then the Bayesian estimator could work by creating a new FourierProbabilityDist every round, rather than copying and replacing all of the contents every single time; this might be a neater solution.

src/openfermion/measurements/_bayesian_estimators.py
kevinsung: You should pull the matrix generation function out of FourierProbabilityDist because they aren't part of the description of the probability distribution. Don't stick them into BayesEstimator; just have them be stand-alone functions.

My understanding is that the probability distribution is completely defined by the A's and the c's, so I don't see where the variance is coming in. If you mean that the A's aren't known for sure, and we only have guesses for them, then we're actually representing a different object, which is more like a probability distribution over probability distributions.

obriente: Hi Kevin, sorry for taking so long to get back to this.

The probability distribution is a distribution over both the amplitudes (A's) and the angles. For the angles I'm storing a representation of the entire function, but for the amplitudes I'm just storing an estimate of the average (as this is what's needed to update the angles), and data about the experiments (_p_vecs), and an initial estimate (amplitude_guess, amplitude_vars) to update this via maximum likelihood estimation.

I think pulling the matrix generation function out of FourierProbabilityDist is a great idea. I'm also going to pull out all of the estimation functions used in calculating the ML estimation out of BayesEstimator and leave them as separate functions in _probability_distributions.py , as they have a similar function.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,358 @@
...
+            init_amplitude_vars (numpy array or list): variance of the
+                estimated distribution in init_amplitude_guess
+            num_vectors (int): number of phases phi_j to estimate
+            num_freqs (int): number of wave components cos(n*phi_j)
+                to store coefficients for (this provides a cut-off to
+                the sensitivity of the function).
+            max_n (int): the maximum n required by the user in
+                multiplying the stored function functions of the form
+                cos(n*phi_j + beta).
+            vector_guess (numpy array): a prior estimate of the function,
```
kevinsung: What function is this a prior estimate of? The function being represented by this class? That seems circular to me.

obriente: Yes, that's correct. You need a prior for a Bayesian estimator, which you can get from other means (e.g. classical approximations, or taking something generic).

kevinsung: Hmm I thought we agreed to separate the estimator concept from the probability distribution concept. We need to avoid this circular logic in the description of this variable.

obriente: Fair point. I've edited the docstring to remove reference to this and updated the variable names to make more sense, hope this looks better now.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,363 @@
...
+from scipy import sparse
+from numpy import cos, sin, pi
+from openfermion.config import EQ_TOLERANCE
+
+
+class FourierProbabilityDist(object):
+    """
+    Stores a multivariant Fourier representation of a periodic function:
+
+    f(phi_0,phi_1,...) = sum_j A_j f_j
```
kevinsung: I'm sorry, @obriente , but I still do not understand. You're claiming here that this class represents a _function_. Here, you wrote the function as
```
f(phi_0,phi_1,...) = sum_j A_j f_j
```
so `f` is the function being represented by the class.

In the docstring of `__init__`, you refer to the "distribution" of the `A_j`. But for the function `f` defined above, there's no such thing as a "distribution" for the `A_j`; there is just the `A_j`.

So is this class representing the function
```
f(phi_0,phi_1,...) = sum_j A_j f_j
```
or is it representing something else?

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,363 @@
...
+            # at the chosen set of points
+            vector = self._fourier_vectors[:, j]
+            y_vec = [self._make_dist(vector, theta) for theta in x_vec]
+
+            # Store and plot data
+            y_vecs.append(y_vec)
+
+        return numpy.array(y_vecs)
+
+    def _holevo_centers(self, vectors=None):
```
kevinsung: This hidden method isn't used anywhere in the class.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,363 @@
...
+            centers (numpy array): the corresponding average angle of the
+                distribution.
+        """
+        if vectors is None:
+            vectors = self._fourier_vectors
+
+        centers = numpy.angle(vectors[2, :] + 1j*vectors[1, :])
+        return centers
+
+    def _holevo_variances(self, vectors=None, maxvar=4*numpy.pi**2/12):
```
kevinsung: This hidden method isn't used anywhere in the class.

kevinsung:
It seems that FourierProbabilityDist is just being used by BayesianEstimator as a collection of helper functions. It's not really holding its own weight as a class. I would suggest getting rid of the FourierProbabilityDist class and replacing it with functions which take as part of their input the numpy arrays representing the probability distribution.

src/openfermion/measurements/_probability_distributions.py
obriente: Oh, this is a very good point, the docstring is flat wrong. This class should store a representation of a distribution P(A_0,A_1,...,\phi_0,\phi_1,...), and in particular in this class we approximate P=P(A_0,A_1,...)\prod_jP(\phi_j), and store data about P(A_0,A_1,...) and P(phi_j).

Thanks for pointing this out, I feel like this might have been the source of a lot of confusion :(. I'll rewrite the docstring now.

src/openfermion/measurements/_probability_distributions.py
obriente: Very good point, this shouldn't be hidden.

src/openfermion/measurements/_probability_distributions.py
obriente: Same as above.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -21,17 +21,31 @@
...
+    $|\psi\rangle = sum_j a_j |phi_j\rangle$,
 
-    It also provides routines for multiplying f by functions of the form
-    cos^2(n*phi_j + beta), and for calculating maximum-likelihood
-    distributions of the A_j variables.
+    where psi is the quantum state, phi_j are the eigenstates of U,
+    and a_j are the amplitudes.
+
+    This class stores a distribution P(phi_0, phi_1,... , A_0, A_1,...)
+    (where A_j = |a_j|^2) of our knowledge of these variables which may
```
kevinsung: This is confusing. Here, you use `phi_j` as variables, but in the equation above, the `phi_j` label eigenstates, which are not variables.

obriente: I'm labeling eigenstates by their eigenvalues - U|\phi_j>=e^{i\phi_j}|\phi_j>. Let me make explicit mention of this in the docstring.

kevinsung: I see. I would make this clearer by mentioning that the eigenvalues are unknown.

obriente: Fair enough, I've inserted it.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+    $|\psi\rangle = sum_j a_j |phi_j\rangle$,
+
+    where psi is the quantum state, |phi_j\rangle are the eigenstates of U,
+    labeled by their eigenvalues ($U|phi_j\rangle = phi_j|phi_j\rangle$),
+    and a_j are the amplitudes.
+
+    This class stores a distribution P(phi_0, phi_1,... , A_0, A_1,...)
+    (where A_j = |a_j|^2) of our knowledge of these variables which may
+    then be updated by the results of experiments (in particular here,
+    QPE experiments).
```
kevinsung: Since this class is meant to represent the distribution, don't mention anything about the possibility of updating it. If you do want to mention QPE, then say something like "the distribution can _determined_ by the results of experiments". The updating part should be handled solely by the estimator.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+                estimated distribution in amplitude_mean.
+                We only allow None for a single vector (in which case
+                this variable is unused).
+            num_vectors (int): number of phases phi_j to estimate
+            num_freqs (int): number of wave components cos(n*phi_j)
+                to store coefficients for (this provides a cut-off to
+                the sensitivity of the function).
+            fourier_vector (numpy array or None): the function
+                given in terms of the Fourier components c_{j,n}.
+                If None, prior taken as a flat function over [-pi,pi].
```
kevinsung: Please document the expected shapes of all the numpy array arguments.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,378 @@
...
+            amplitude_mean (numpy array or list or None):
+                We take an normal distribution for the amplitudes
+                which requires for each a mean and a variance.
+                We only allow None for a single vector (in which case
+                the amplitude is fixed to 1), as we need this
+                to break the symmetry of the problem.
+            amplitude_var (numpy array or list or None): variance of the
+                estimated distribution in amplitude_mean.
+                We only allow None for a single vector (in which case
+                this variable is unused).
```
kevinsung: Is this a 1-d or 2-d array? If it's 2-d then it's really a covariance matrix, right?

src/openfermion/measurements/_probability_distributions.py
obriente: I've removed the mention of updates from the docstring. The word update appears in a few other places in the code, but I feel where it does it's appropriate.

src/openfermion/measurements/_probability_distributions.py
obriente: 2-d, and yes, this is the covariance matrix. Edited

src/openfermion/measurements/_probability_distributions.py
obriente: Done

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,379 @@
...
+
+    $U|phi_j\rangle = phi_j|phi_j\rangle$,
+
+    and a_j are the (unknown) corresponding amplitudes.
+
+    This class stores a distribution P(phi_0, phi_1,... , A_0, A_1,...)
+    (where A_j = |a_j|^2) of our knowledge of these variables.
+
+    In particular, it approximates this distribution as:
+    P(phi_0, phi_1,... , A_0, A_1,...)=P(A_0, A_1,...)*\prod_jP(\phi_j).
```
kevinsung: Is P(A_0, A_1, ...) a normal distribution? If so then document it here.

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,379 @@
...
+
+    It then represents each phase distribution:
+    P(\phi_j) = sum_n (c_{j,2n} cos(n*phi_j) + c_{j,2n-1} sin(n*phi_j))
+
+    Separately, it stores an estimate of the mean amplitude \bar{A_j},
+    and data that allows for this to be recalculated (rather than a direct
+    representation of the function).
```
kevinsung: Explain what you mean by "data that allows for this to be recalculated".

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,379 @@
...
+    This class stores a distribution P(phi_0, phi_1,... , A_0, A_1,...)
+    (where A_j = |a_j|^2) of our knowledge of these variables.
+
+    In particular, it approximates this distribution as:
+    P(phi_0, phi_1,... , A_0, A_1,...)=P(A_0, A_1,...)*\prod_jP(\phi_j).
+
+    It then represents each phase distribution:
+    P(\phi_j) = sum_n (c_{j,2n} cos(n*phi_j) + c_{j,2n-1} sin(n*phi_j))
+
+    Separately, it stores an estimate of the mean amplitude \bar{A_j},
```
kevinsung: Instead of saying that it stores an "estimate of the mean", you should say that the distribution of the A's is normal and you are storing the mean and covariance of the normal distribution (if I understand correctly).

kevinsung:
Thanks for updating the docstring of FourierProbabilityDist. I think I understand now what it is supposed to represent, and it makes sense to have a class for it.

>I see what you mean here, but I can imagine at some point wanting to replace the FourierProbabilityDist class with another representation of the probability distribution. Which isn't to say we should leave it as is, rather, I think BayesianEstimator needs to be rewritten to treat FourierProbabilityDist as a more abstract object, and FourierProbabilityDist needs to be rewritten to understand its own math.
This means (if I'm not mistaken) I need to pull these update matrices out of both BayesianEstimator (as they're distribution-specific) and FourierProbabilityDist.
Does that make sense? This will take a fair amount of refactoring, not sure if I can get around to it before the end of the year.

I would avoid abstracting the probability distribution concept until we actually have another use case. However, it might be good to abstract the estimator concept. Do you ultimately plan to have BayesianEstimator and TimeSeriesEstimator inherit from the same class?

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,379 @@
...
+                      (a_vec-self._prior_amplitude_mean)[
+                        :, numpy.newaxis])).item()
+
+    def _dinit_mlikelihood(self, a_vec):
+        # The derivative of the above likelihood function
+        return numpy.dot(
+            self._inv_covar_mat,
+            (a_vec-self._prior_amplitude_mean))
+
+    def _jinit_mlikelihood(self):
```
kevinsung: None of these hidden methods are being used in any non-hidden methods in this class. So you should unhide them or get rid of them.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,520 @@
...
+    Class to estimate a QPE experiment in the absence of error.
+    """
+
+    def __init__(self,
+                 prior_amplitude_mean=None,
+                 prior_amplitude_var=None,
+                 num_vectors=1,
+                 num_freqs=1000,
+                 max_rotations=1,
+                 prior_vector=None,
```
kevinsung: Instead of taking the arguments `prior_amplitude_mean`, `prior_amplitude_var`, `num_vectors`, `num_freqs`, and `prior_vector`, just have one argument `prior` which is a FourierProbabilityDist.

src/openfermion/measurements/_bayesian_estimators.py
```diff
@@ -0,0 +1,520 @@
...
+        self._make_projector()
+        self._max_rotations = max_rotations
+        self._matrices = make_fourier_matrices(max_rotations, num_freqs)
+
+        self.reset()
+
+    def reset(self):
+        """
+        Resets estimator to initial state.
+        """
```
kevinsung: Can you explain why one would ever want to reset the estimator to its initial state?

obriente: For testing the estimator performance one needs to iterate over multiple estimations and average the result. But, as you say, it would be possible to just create a new estimator for each iteration. When I've been doing this before this was a pain (because creating BayesEstimator._matrices took so long).  But, as I mentioned above, these matrices shouldn't even be in the estimator, as they're unique to FourierProbabilityDist.

As these matrices are static, I think the solution is to store them separately to both classes in an object that gets referenced by every instance of FourierProbabilityDist. This is a bit abstract for me, so I'll have to look around for the most appropriate pythonic way of doing this. Once I've sorted this out, I think I agree with removing the reset function from BayesEstimator.

src/openfermion/measurements/_bayesian_estimators.py
obriente: Good idea, I'll implement this now.

src/openfermion/measurements/_probability_distributions.py
obriente: Thanks for noticing this, I'll unhide them (they're used by BayesEstimator).

src/openfermion/measurements/_bayesian_estimators.py
obriente: > Thanks for updating the docstring of FourierProbabilityDist. I think I understand now what it is supposed to represent, and it makes sense to have a class for it.
> 
> > I see what you mean here, but I can imagine at some point wanting to replace the FourierProbabilityDist class with another representation of the probability distribution. Which isn't to say we should leave it as is, rather, I think BayesianEstimator needs to be rewritten to treat FourierProbabilityDist as a more abstract object, and FourierProbabilityDist needs to be rewritten to understand its own math.
> > This means (if I'm not mistaken) I need to pull these update matrices out of both BayesianEstimator (as they're distribution-specific) and FourierProbabilityDist.
> > Does that make sense? This will take a fair amount of refactoring, not sure if I can get around to it before the end of the year.
> 
> I would avoid abstracting the probability distribution concept until we actually have another use case. However, it might be good to abstract the estimator concept. Do you ultimately plan to have BayesianEstimator and TimeSeriesEstimator inherit from the same class?

Yes, this would be the ultimate plan. There are a few key differences between the two though, most notably that BayesEstimator can use external data to start with a prior (TimeSeriesEstimator cannot easily), and that TimeSeriesEstimator supports a much more compressed data input (BayesEstimator really needs to process each experiment individually).

src/openfermion/measurements/_probability_distributions.py
obriente: I'm not doing this precisely, but it's honestly what I *should* be doing. Let me take the time to fix this, which might be a few weeks-, because otherwise it's just going to be a source of confusion.

src/openfermion/measurements/_probability_distributions.py
obriente: So this is a bit of a hack, and honestly it's probably just better if I change it (the comment below is how I should be doing this). What I have been doing is running a maximum-likelihood estimator on the prior and a bunch of experimental data to get a good estimate of the mean amplitudes.

The full details of this are in the paper, but they're pretty sloppy and difficult to understand. As mentioned in the next comment, what I should be doing is getting rid of this part, and just updating a Normal distribution, similar to how the prior is passed. I'll get on this, but it'll take some time.

src/openfermion/measurements/_probability_distributions.py
obriente: See below. It should be, and i'll fix it up later.

obriente:
Just to keep track, my todo list is the following now:
 - shift the matrices out of BayesEstimator to a separate object that is recognized by the class FourierProbabilityDist, and shift all of the under the hood arithmetic into FourierProbabilityDist.
 - rewrite the amplitude estimation to really just be updating a normal distribution
 - Make BayesEstimator just take an abstract object ProbabilityDist which it does abstract mathematics on.

(This will take a while to get through, I'll update when it's done)

kevinsung:
@obriente That sounds like a good todo list. No rush. Thanks for bearing with me; I think this is really shaping up!

src/openfermion/measurements/_probability_distributions.py
```diff
@@ -0,0 +1,379 @@
...
+            row_ind_sin += [2*j+2*nrot, 2*j+2*nrot-1]
+            col_ind_sin += [2*j-1, 2*j]
+
+    # Make sparse matrices
+    cos_matrix = sparse.csr_matrix(
+        (data_cos, (row_ind_cos, col_ind_cos)))
+    sin_matrix = sparse.csr_matrix(
+        (data_sin, (row_ind_sin, col_ind_sin)))
+
+    return cos_matrix, sin_matrix
```
kevinsung: I wonder if this method can be made much faster. How large do you think these matrices will get? If they don't get past, say, 1000x1000, then it might be okay to just use dense matrices.

kevinsung:
By the way, if you can find a way to split this PR into smaller pieces, that would help speed up the review process.

src/openfermion/measurements/_probability_distributions.py
obriente: Unfortunately these matrices can be quite large - the estimator accuracy is limited roughly to 2*pi/(matrix size), so I tend to run this up to ~10000 or even 100000. Luckily they're incredibly sparse (max 4 non-zero entries per row).

obriente:
Do you think that would be a good idea at this stage? I'm open to say, making different pull requests for ProbabilityDist and BayesEstimator (and perhaps staging the pull of BayesEstimator after writing a generic estimator class, then TimeSeriesEstimator can be extended from there as well).

kevinsung:
It's okay to leave it in one piece too, for now.

obriente:
Closing PR

PR #476, opened 2018-10-02
Added particle hole symmetry to general Hubbard model

bryano:


src/openfermion/hamiltonians/_general_hubbard.py
```diff
@@ -157,7 +169,9 @@ class FermiHubbardModel:
     def __init__(self, lattice,
                  tunneling_parameters=None,
                  interaction_parameters=None,
-                 potential_parameters=None
+                 potential_parameters=None,
+                 magnetic_field=0,
+                 phs=False
```
kevinsung: Yeah, it's long, but let's keep it as `particle_hole_symmetry`.

src/openfermion/hamiltonians/_general_hubbard.py
```diff
@@ -27,21 +27,30 @@
...
 
-def tunneling_operator(i, j, coefficient):
+def tunneling_operator(i, j, coefficient=1.):
     return (FermionOperator(((i, 1), (j, 0)), coefficient) + 
             FermionOperator(((j, 1), (i, 0)), coefficient.conjugate()))
 
 def number_difference_operator(i, j, coefficient=1.):
     return number_operator(i, coefficient) - number_operator(j, coefficient)
 
+def number_difference_operator(i, j, coefficient=1.):
```
kevinsung: Got duplicated.

src/openfermion/hamiltonians/_general_hubbard.py
```diff
@@ -341,7 +354,20 @@ def potential_terms(self):
                 for spin_index in self.lattice.spin_indices:
                     i = self.lattice.to_spin_orbital_index(
                             site_index, param.dof, spin_index)
-                    terms += number_operator(i, -param.coefficient)
+                    terms += number_operator(i, -param.coefficient, 
+                            particle_hole_symmetry=self.particle_hole_symmetry)
+        return terms
+
+
+    def field_terms(self):
```
kevinsung: Got duplicated.

src/openfermion/hamiltonians/_general_hubbard.py
```diff
@@ -164,7 +170,8 @@ def __init__(self, lattice,
                  tunneling_parameters=None,
                  interaction_parameters=None,
                  potential_parameters=None,
-                 magnetic_field=0
+                 magnetic_field=0,
```
kevinsung: Sorry I didn't catch this before: this should be a float.

src/openfermion/hamiltonians/_general_hubbard.py
```diff
@@ -177,6 +184,8 @@ def __init__(self, lattice,
             potential_parameters (Iterable[Tuple[int, Number]], optional): The
                 potential parameters.
             magnetic_field (Number, optional): The magnetic field. Default is 0.
```
kevinsung: Change this from Number to float

bryano: What about `Real`?

kevinsung: That's not a type ...

bryano: https://docs.python.org/3/library/numbers.html#numbers.Real

kevinsung: Let's just do the normal thing and call it a float.

kevinsung: ... but it's not important enough to block this.

bryano: Sorry for letting this sit. I just changed it, because it doesn't really matter. (It's just in a docstring; if we were actually typechecking, I'd probably want to allow ints, but it seems that typing is a lot more informal in OpenFermion.)

bryano: @kevinsung ready for re-review

PR #486, opened 2018-11-14
Port packaging scripts from Cirq

### LINKED ISSUE #483: Development releases
Currently, openfermion-cirq has to point at this github repository in order to use in-development features of openfermion during development of openfermion-cirq. This causes a bunch of problems w.r.t. packaging, since github dependencies aren't valid dependencies that can be uploaded to pypi or used by setup.py. I recently started making development releases for cirq ([scripts are here](https://github.com/quantumlib/Cirq/tree/master/dev_tools/packaging), example package [here](https://pypi.org/project/cirq/0.4.0.dev4220181112191000/)) and it immediately solved the analogous problem between cirq and openfermion-cirq. You just depend on `cirq~=0.4.0.dev` and it grabs the latest dev release of 0.4.0, and this is a valid version that works with pypi and setup.py.

We should get the same thing working in openfermion.

kevinsung:
Fixes #483


babbush:
Can you explain what this PR does?

kevinsung:
It adds scripts related to packaging and uploading to PyPI. For instance, you can cut a new release with a one-line command. You can also cut a dev release, or upload a release to the Test PyPI and verify that the release works as expected, which is good to do before doing a real release.

dev_tools/packaging/publish-dev-package.sh
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Cirq Developers
```
Strilanc: Change this to the openfermion copyright notice.

dev_tools/packaging/produce-package.sh
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Cirq Developers
```
Strilanc: Change this to the openfermion copyright notice.

dev_tools/packaging/verify-published-package.sh
```diff
@@ -0,0 +1,97 @@
...
+
+    # Check that code runs without dev deps.
+    echo Checking that code executes
+    "${tmp_dir}/${PYTHON_VERSION}/bin/python" -c "import openfermion; print(openfermion.FermionOperator((1, 1)))"
+    "${tmp_dir}/${PYTHON_VERSION}/bin/python" -c "import openfermion; print(openfermion.QubitOperator((1, 'X')))"
+
+    # Run tests.
+    echo Installing pytest requirements
+    if [ "${PYTHON_VERSION}" = "python2" ]; then
+        "${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet pytest mock
```
Strilanc: Are these requirements the same for openfermion as they are for cirq and openfermioncirq? You probably don't need `mock`.

Just unconditionally do `"${tmp_dir}/${PYTHON_VERSION}/bin/pip" install --quiet pytest mock` regardless of version

kevinsung:
@Strilanc I made the changes but staled your review

kevinsung:
... but I need this right now. Anyway, it's pretty clear that I just did what you said.

PR #488, opened 2018-11-29
added diagonal (next-nearest) neighbors to square lattice

bryano:


kevinsung:
Aren't there next-nearest neighbors which are not diagonal? i.e., two steps over to the left/right/up/down

bryano:
Yea, there are those too. I'm not sure what the best terminology is to disambiguate the two. One suggestion is to use "next-nearest neighbor" for the diagonal ones and "next-next-nearest neighbor" for the axis-aligned ones, and explain in the documentation that this is using the Euclidean distance.

The multi-band Hubbard literature I've seen (which is not that much) only uses the diagonal ones here.

kevinsung:
I think the ones you use here should be called "diagonal", and "next-nearest" should mean Manhattan distance 2.

bryano:
Done. Why are the tests taking so long?

kevinsung:
The coveralls check has been hanging lately and I don't know why.

PR #489, opened 2018-12-05
Bump version to 0.9.0.dev

kevinsung:


kevinsung:
I'm making this a dev version so I can release a dev version to test with Openfermion-Cirq before releasing the real thing.

PR #493, opened 2018-12-12
Changes index linearization in _lattice.py to match that in _hubbard.py

bryano:


src/openfermion/utils/_lattice.py
```diff
@@ -187,7 +187,12 @@ def validate_dofs(self, dofs, length=None):
 class HubbardSquareLattice(HubbardLattice):
     r"""A square lattice for a Hubbard model.
     
-    There are two types of edges: 'onsite' and 'neighbor'.
+    Valid edges are:
```
kevinsung: "Valid edge types"

bryano: Done.

PR #496, opened 2018-12-26
Add README link to run Example Notebooks in MyBinder

NHDaly:
The MyBinder website provides free, cloud-based hosting of jupyter notebooks, built from public git repos!! :)

I was able to just enter this repo's URL into the builder on https://mybinder.org/, and it correctly set up the repo based on the Dockerfile and everything seems to work! :) This means with one-click, people can open a browser-based environment and start interacting with the tutorial! :)

Sorry i've never contributed to this Repo before, but i was curious to try this out. The easiest way for me to get started was to plug it in to MyBinder, and so i thought i'd share that back with you!!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,6 +24,7 @@ Hamiltonians. For more information, see our
 
 
 Run the interactive Jupyter Notebooks on MyBinder:
+
 .. image:: https://mybinder.org/badge_logo.svg
     :target: https://mybinder.org/v2/gh/quantumlib/OpenFermion/master?filepath=examples
 
--- a/README.rst
+++ b/README.rst
@@ -22,6 +22,11 @@ Hamiltonians. For more information, see our
 
 .. image:: https://img.shields.io/badge/python-2.7%2C%203.4%2C%203.5%2C%203.6-brightgreen.svg
 
+
+Run the interactive Jupyter Notebooks on MyBinder:
+.. image:: https://mybinder.org/badge_logo.svg
+    :target: https://mybinder.org/v2/gh/quantumlib/OpenFermion/master?filepath=examples
+
 Plugins
 =======
 


NHDaly:
We could also delete the text and just leave the button? Or maybe change the text to directly reference the "tutorial"? Maybe something like `"Open the interactive tutorial notebooks:"`, or `"Click here to run the interactive tutorial notebooks:"`?
And/or maybe it's better to move it to a later section of the readme.

NHDaly:
(Coverage is noisy; this is a README-only change.)

NHDaly:
Also, for posterity, here's the link I've added:
https://mybinder.org/v2/gh/quantumlib/OpenFermion/master?filepath=examples

babbush:
Thanks for the contribution, this looks very cool. And sorry for the delay, many of us were away from OpenFermion for the holidays. Don't worry about the coverage flag, it messes up sometimes.

My main question about this PR concerns security. It seems that we would be relying on your configuration of MyBinder, the security of which we cannot guarantee (please don't take that personally!). Would it instead be possible to walk us through how our organization can configure MyBinder? Perhaps there is another solution that would be acceptable? Looping in @dabacon and @Strilanc who might have suggestions when they are back from vacation.

NHDaly:
> And sorry for the delay, many of us were away from OpenFermion for the holidays.

haha not a problem! Happy holidays; I'm glad you were taking time off, and I hope you enjoyed it! :)

----------

> My main question about this PR concerns security. It seems that we would be relying on your configuration of MyBinder, the security of which we cannot guarantee (please don't take that personally!).

Ah yes, that's a good concern! In this case, it's actually not _my_ configuration of MyBinder. Rather, it's the automatic configuration of MyBinder, built automatically by reading the Docker files from your repo! :) You can read more about it here: https://mybinder.readthedocs.io/en/latest/

> Would it instead be possible to walk us through how our organization can configure MyBinder? 

Hmm so i'm not too sure. (I'm actually not associated with MyBinder at all; i've just used it because it's very handy!)
To set it up, I just went to the main page, https://mybinder.org/, and entered your repo's URL and the path to the notebooks (`examples`). It's a purposefully minimal configuration, because it just reads the config from your Repository. You can change the configuration by changing the files committed in the repository.

That said, if your concern is security of the hosting of the jupyter instances, i don't think you would want to use MyBinder. Of course your organization can always host its own cloud-based Jupyter notebooks, but I think that's non-trivial to set up. From my understanding, the idea behind the MyBinder service is that the NumFocus organization is donating computing for the free hosting of scientific jupyter notebooks in the cloud. :)

NHDaly:
Here's more information on the MyBinder organization: https://jupyterhub-team-compass.readthedocs.io/en/latest/team.html#binder-team

babbush:
Okay, thanks for the clarification. While I'll still wait for advice from my more security saavy but still on vacation colleagues (@dabacon and @Strilanc), I think it is reasonable for us to trust MyBinder, which looks like a reasonable open source project with a BSD license. I think I was confused by what you wrote about using the Dockerfile for our repo. Did you need to point MyBinder at our Dockerfile somehow? Or are you saying it just finds our Dockerfile from searching through the repo? The concern is that if one had to do something like upload the Dockerfile, then one could potentially upload something nefarious instead.

A related question is whether the MyBinder configuration will always automatically pull the latest version of OpenFermion?

NHDaly:
Sorry I didn't do a good job explaining it! :)

I'll answer your questions below, but first here's my attempt  at a better summary explanation:

MyBinder is a service that clones the code and configuration files from a github repo; builds a Docker image that contains all the files in the repo and a jupyter notebook server configured based on any configuration files present in the repo; and then provides free hosted instances of that Docker image to users so that they can play with the code in the given repo in their browser. :) Hopefully that's a better explanation!

---------

Your questions:


> I think I was confused by what you wrote about using the Dockerfile for our repo. Did you need to point MyBinder at our Dockerfile somehow? Or are you saying it just finds our Dockerfile from searching through the repo? The concern is that if one had to do something like upload the Dockerfile, then one could potentially upload something nefarious instead.


No i didn't point it at your Dockerfile, just gave it the repo URL and nothing else! (It's nifty, no?)

Ah, so, I previously thought it was configuring the Notebooks by reading the Dockerfile it found in your repo. But after investigating more carefully, I think it might be reading from the `requirements.txt` file at the top-level of your Notebook.

This is what their docs say about how it configures the notebook server: 

- It can configure the notebook based on any of these types of configuration files:
    https://mybinder.readthedocs.io/en/latest/config_files.html#config-files

    It says: `"If a Dockerfile is present, all other configuration files will be ignored."`, but I'm not sure if it's able to find your Dockerfile in the `docker/` folder, since _this page_ ([#preparing-a-repository-for-binder](https://mybinder.readthedocs.io/en/latest/using.html#preparing-a-repository-for-binder)) says `"Configuration files may be placed in the root of your repository or in a binder/ folder in the repository’s root (i.e. myproject/binder/)."`

Haha sorry i'm not more familiar.

So to answer your question:
 - _No_, i didn't provide any configuration myself.
 - _Yes_, it is reading the config from your repo.
 - And _maybe_ it is reading your `docker/dockerfile`, or maybe it's just reading the `requirements.txt` file.

The first time it builds a repo, the logs say what it's doing, so we could read those to answer the above question. Unfortunately, now it is cached in their server somewhere, so it reuses the already built docker container.

----------

And finally, your next question:

> A related question is whether the MyBinder configuration will always automatically pull the latest version of OpenFermion?

Yes! After any change to the `master` branch (because that's the branch named in the URL I used in this PR), the first time a user clicks the link to open the notebook, MyBinder will rebuild its docker image based on the code+config in this repo, and all subsequent accesses will use this newly generated image.

And so, after the next change to this repo, you could go look at the logs to see which config file it's using to build its notebook image from.

----------

:) I'm sorry i'm not well versed in these topics so I might not be doing a good job explaining!

babbush:
Thanks! This all sounds good to me. I'm going wait until my colleagues with more security knowledge are back next week before merging but I anticipate there won't be a problem.

NHDaly:
Thanks! :)

kevinsung:
My understanding of MyBinder is that the only access it uses (the only access it _can_ use, as it stands) is that which is available to any member of the public. Specifically, it only needs to be able to clone the repo and download the Jupyter notebooks. Therefore, there are no security issues with this PR.

PR #497, opened 2018-12-28
Use pathos package to restart closed worker pools [Fixes #461]

### LINKED ISSUE #461: ParallelLinearQubitOperator can't perform more than one multiplication
I make a ParallelLinearQubitOperator and multiply the same state twice:
```
import openfermion

qubit_op = openfermion.QubitOperator('X0')
linear_qubit_op = openfermion.generate_linear_qubit_operator(
    qubit_op,
    options=openfermion.LinearQubitOperatorOptions(processes=2))
state = [1.0, 0.0]

linear_qubit_op.dot(state)
linear_qubit_op.dot(state)
```
This gives
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-f4067f5432d2> in <module>()
      8 
      9 linear_qubit_op.dot(state)
---> 10 linear_qubit_op.dot(state)

~/environments/ofc/lib/python3.5/site-packages/scipy/sparse/linalg/interface.py in dot(self, x)
    360 
    361             if x.ndim == 1 or x.ndim == 2 and x.shape[1] == 1:
--> 362                 return self.matvec(x)
    363             elif x.ndim == 2:
    364                 return self.matmat(x)

~/environments/ofc/lib/python3.5/site-packages/scipy/sparse/linalg/interface.py in matvec(self, x)
    217             raise ValueError('dimension mismatch')
    218 
--> 219         y = self._matvec(x)
    220 
    221         if isinstance(x, np.matrix):

~/environments/ofc/lib/python3.5/site-packages/openfermion/utils/_linear_qubit_operator.py in _matvec(self, x)
    185         vecs = pool.imap_unordered(apply_operator,
    186                                    [(operator, x)
--> 187                                     for operator in self.linear_operators])
    188         pool.close()
    189         pool.join()

/usr/lib/python3.5/multiprocessing/pool.py in imap_unordered(self, func, iterable, chunksize)
    300         '''
    301         if self._state != RUN:
--> 302             raise ValueError("Pool not running")
    303         if chunksize == 1:
    304             result = IMapUnorderedIterator(self._cache)

ValueError: Pool not running
```
It works fine if I only multiply once.
 
The same error occurs if you try to compute the ground state, like
```
import openfermion

qubit_op = openfermion.QubitOperator('X0 Z5')
linear_qubit_op = openfermion.generate_linear_qubit_operator(
    qubit_op,
    options=openfermion.LinearQubitOperatorOptions(processes=2))

openfermion.get_ground_state(linear_qubit_op)
```

It appears that the Pool "shuts off" after one multiplication. @sheilaliuxl

mapmeld:
```multiprocessing.Pool``` is used only in the file _linear_qubit_operator.py's ```_matvec``` and its tests. This code then runs ```close()``` and ```join()``` to terminate and garbage-collect the stored pool after use, but it is still stored and attempted to be reused in the examples of #461 

I found three options to resolve the issue of a closed pool being reused.

- Removing the storing of ```self.pool``` so we always create a new worker pool
- Changing ```if self.pool is None:``` to add ```or self.pool._state != "RUN":``` - I didn't find another way to check ```pool._state```
- Using pathos.pools.ProcessPool - this adds a dependency, but is otherwise a drop-in replacement. This adds the ability to run ```self.pool.restart()``` on a stored worker pool

I don't know enough about resource management to know why the pool was being reused here, so let me know which of these options makes sense to you

### CODE REVIEW

**Comment by kevinsung** [comment]
For now I would go with solution 1 (making the pool each time). I actually think that this code can be entirely rewritten to be much faster, but I don't have time to look into it right now. So let's just go with solution 1 and avoid the new dependency.

**Comment by mapmeld** [comment]
sounds good - I've changed the solution to reset the pool each time, and added the code from the issue to the test

File: src/openfermion/utils/_linear_qubit_operator_test.py [hunk 1]
```diff
@@ -187,6 +179,14 @@ def test_matvec_compare(self):
...
                                        mat_expected.A))
 
+    def test_closed_workers_not_resued(self):
+        """Issue #461 - no longer attempting to reuse closed worker pool"""
+        qubit_op = QubitOperator('X0 Z5')
+        linear_qubit_op = generate_linear_qubit_operator(
+            qubit_op,
+            options=LinearQubitOperatorOptions(processes=2))
+        get_ground_state(linear_qubit_op)
+
```
kevinsung: Remove this test. Instead, add a test to ParallelLinearQubitOperatorTest that simply constructs a ParallelLinearQubitOperator and applies it twice to the same state (similar to what I did in the issue). No need to mention the issue in the docstring.

Issue #444, opened 2018-09-05
get_interaction_operator should raise an error when the given operator is not Hermitian


kevinsung:
This would improve the type safety of InteractionOperator. `get_quadratic_hamiltonian` and `get_diagonal_coulomb_hamiltonian` already do this.

PR #498, opened 2018-12-31
Test is_hermitian on get_interaction_operator

mapmeld:
Attempted fix for #444 : ```get_interaction_operator``` raises a ```QuadraticHamiltonianError``` if input is not Hermitian

The molecular operator test was not Hermitian, so I've changed its circuit, and used the original code as a new test expected to raise this error.

The tests for bad one-body, bad two-body, and non-molecular are not Hermitian, but their tests expect an ```InteractionOperatorError```. This led me to put ```is_hermitian``` after these checks.

I made an additional change to the docs where a similar ```QuadraticHamiltonianError``` is referred to as a ```TypeError```, though it inherits only from ```Exception```


src/openfermion/transforms/_conversion.py
```diff
@@ -168,6 +169,11 @@ def get_interaction_operator(fermion_operator, n_qubits=None):
             raise InteractionOperatorError('FermionOperator does not map '
                                            'to InteractionOperator.')
 
+    if not is_hermitian(fermion_operator):
+        raise QuadraticHamiltonianError(
+            'FermionOperator does not map '
+            'to QuadraticHamiltonian (not Hermitian).')
```
kevinsung: Running `is_hermitian` on a FermionOperator is very expensive because it requires normal-ordering, which potentially has exponential cost. Instead of using it here, you should check for hermiticity by checking `one_body` and `two_body`. In the case of `one_body`, the check is simply to see that it is a hermitian matrix. For `two_body`, it might be a bit more complicated and the best way to go might involve editing the method to make the tensor more symmetric.

babbush: In general its true that normal ordering can have exponential cost. However, that's only true for N-body fermion operators, and presumably one is not trying to map those to InteractionOperator. Furthermore, we already normal order the fermion_operator in this function on line 131. So, I don't think its necessarily a problem.

mapmeld: Thanks to both of you for explaining this. Is it still costly to check a normal-ordered operator or is it more like sorting an already-sorted list?

Would it make sense to rewrite the code to skip that check here, knowing that the operator is already normal-ordered?

src/openfermion/transforms/_conversion.py
```diff
@@ -113,6 +113,7 @@ def get_interaction_operator(fermion_operator, n_qubits=None):
     Raises:
         TypeError: Input must be a FermionOperator.
         TypeError: FermionOperator does not map to InteractionOperator.
+        QuadraticHamiltonianError: FermionOperator does not map to QuadraticHamiltonian.
```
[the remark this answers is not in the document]
mapmeld: 👍

src/openfermion/transforms/_conversion.py
kevinsung: I still think that the check I suggested would be cheaper and that implementing it would have the added benefit of ensuring that the returned tensors are as symmetric as possible, but I won't insist on it.

So I'm alright with doing the check this way. Please move it earlier, to just after we normal-order the operator. Might as well do all the checks on `fermion_operator` before attempting to construct the InteractionOperator.

kevinsung: >If the FermionOperator is not Hermitian, then I think it is still possible to end up with an InteractionOperator that is Hermitian.

I would consider that a bug because then the routine is doing more than just converting from one type to another; it's altering the operator itself. If we want to allow such behavior then we should be clear about what we're promising and what we're not promising.

babbush: its only altering it if it is non-Hermitian. Which is why adding this check is a good idea, because it makes sure that only Hermitian operators can be converted.

kevinsung: Also, note that my suggested change does not alter the behavior of the method at all; it just makes it so that all the checks are done before attempting to construct the InteractionOperator. If the original FermionOperator is not Hermitian, then the method would construct the InteractionOperator and then still fail that check.

babbush: sorry Kevin, you are actually correct. Indeed, I think this routine is bijective whether or not the input is Hermitian. So if the input is non-Hermitian, the output should be obviously non-Hermitian. So indeed, we should be able to just test Hermiticity on the one_body and two_body. The only trouble is that it is a little bit non-trivial to check Hermiticity on the two-body part. Do we have a canned routine for doing that somewhere?

kevinsung: I'm okay with performing this check this way; I just think it makes more sense to put the check before the loop that constructs the InteractionOperator instead of after the loop. It doesn't really matter, but it's more readable that way and could potentially save some running time.

kevinsung: I'm glad you see what I was talking about, Ryan. No, I haven't thought through the best way to check the two-body term, and I don't think there's a canned routine somewhere. This is related to previous disadvantages of InteractionOperator we've discussed before, namely the fact that there's a bunch of freedom in constructing the tensors and it's often nice to make things as symmetric as possible.

babbush: Well in this case, I suggest that for now we go ahead with this PR as is, because it is complicated to test Hermiticity of the two-body operator.

kevinsung: I'm fine with that after addressing the nit of putting the check before the main loop instead of after.

mapmeld: So the reason that this appears after the main loop is that 3 other tests in ```_conversion_test.py``` are designed to trigger certain errors, but they are not Hermitian and run into this error first. I don't know the material well enough to recommend fixes which trigger their errors but not this one.

With Travis CI running the whole test suite, I see there are 6 tests in other files which also are not Hermitian but make use of InteractionOperator.

Would it make sense for me to move this loop and then seek out advice from someone on how to update the other tests?

src/openfermion/transforms/_conversion.py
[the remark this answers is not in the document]
mapmeld: ok, I understand this better now. I was copying the error message and type from the other ```if not is_hermitian``` in the file, not noting it was specific to that method. I believe the latest update resolves this issue

obriente:
What's the status on this?

obriente:
Closing due to a lack of response

PR #506, opened 2019-02-19
Wedge product code between two tensors

ncrubin:
The two tensors are the moments for a particular sequence of
fermionic ladder operators.

Symmetric vs anitysmmetric could be an easy extension to this code.

### DIFF
--- a/src/openfermion/utils/_wedge_product_test.py
+++ b/src/openfermion/utils/_wedge_product_test.py
@@ -82,7 +82,7 @@ def test_random_two_wedge():
 
 def test_random_two_wedge_two():
     """
-    Test wedge product code with random 4-tensor and a random 2-tensor
+    Test wedge product code with a random 4-tensor and a random 4-tensor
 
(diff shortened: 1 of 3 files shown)


src/openfermion/utils/_wedge_product_test.py
```diff
@@ -0,0 +1,145 @@
...
+                                                     u_perm[2], l_perm[2]]
+
+    true_tensor /= 6 ** 2
+    test_tensor = wedge(n_tensor, m_tensor, (2, 2), (1, 1))
+    assert numpy.allclose(test_tensor, true_tensor)
+
+
+def test_random_two_wedge_two():
+    """
+    Test wedge product code with random 4-tensor and a random 2-tensor
```
mfdgroot: I think you have a tiny typo here from copying the last test.
2-tensor -> 4-tensor

ncrubin: 👍

PR #516, opened 2019-05-15
Tapering off qubits using stabilizer conditions.

xabomon:
This module tapers off qubits given a set of stabilizers.
Additionally, it reduces the number of terms of a Hamiltonian (operator) using stabilizer conditions.

The module allows the user to:
1) Select which qubit to taper off by a given stabilizer.
2) Keep the length of the Pauli string after reducing terms the same as the starting one.
3) Return which qubits have been tapered off.

The module comes with documentation and test module.

Xavier Bonet-Monroig and Mark Steudtner


xabomon:
I have signed the CLA!

msteudtner:
I'm ok with my comments being contributed.

src/openfermion/utils/_qubit_tapering_from_stabilizer.py
```diff
@@ -0,0 +1,406 @@
...
+
+class StabilizerError(Exception):
+    """Stabilizer error class."""
+
+    def __init__(self, message):
+        """Throw custom errors connected to stabilizers."""
+        super().__init__(message)
+
+
+def _check_commuting_stabilizers(stabilizer_list, msg, thres=1e-6):
```
ncrubin: use openfermion.config.EQ_TOLERANCE unless there is a separate reason this threshold should be different.

src/openfermion/utils/_qubit_tapering_from_stabilizer.py
```diff
@@ -0,0 +1,406 @@
...
+                                          fixed_positions)
+
+    if output_fixed_positions:
+        return reduced_operator, fixed_positions
+    else:
+        return reduced_operator
+
+
+def taper_off_qubits(operator, stabilizers,  manual_input=False,
+                     fixed_positions=[], output_tapered_positions=False):
```
ncrubin: Use a non-mutable default type like `None` instead of empty list.

src/openfermion/utils/_qubit_tapering_from_stabilizer.py
```diff
@@ -0,0 +1,406 @@
...
+            term_list)
+
+    return even_newer_terms, fixed_positions
+
+
+def reduce_number_of_terms(operator, stabilizers,
+                           maintain_length=False,
+                           output_fixed_positions=False,
+                           manual_input=False,
+                           fixed_positions=[]):
```
ncrubin: It is unlikely that this method will be referenced abstractly later but for safety you should remove the list optional type in exchange for something that won't be mutable.

src/openfermion/utils/_qubit_tapering_from_stabilizer.py
```diff
@@ -0,0 +1,406 @@
...
+    after some of their products will be the identity.
+    This function checks the list of stabilizers (QubitOperator)
+    and raises an error if the identity is found.
+    """
+    for stab in stabilizer_list:
+        if list(stab.terms.keys())[0] == ():
+            raise StabilizerError(msg)
+
+
+def _fix_single_term(term, position, fixed_op, other_op, stabilizer):
```
ncrubin: These "internal" functions still need argument doc strings.  Later somebody maintaining this will find it easier to read if inputs and outputs are specified.  Ditto on all following internal methods.

ncrubin: In fact, what is the motivation for these being "_" named?  Presumably these helper functions are likely to be used in further analysis for this line of research.

src/openfermion/transforms/__init__.py
```diff
@@ -43,3 +43,6 @@
                              weyl_polynomial_quantization)
 from ._remove_symmetry_qubits import (symmetry_conserving_bravyi_kitaev,
                                       edit_hamiltonian_for_spin)
+
+from ._qubit_tapering_from_stabilizer import (reduce_number_of_terms,
```
ncrubin: move import into alphabetical order.

ncrubin:
Hey @xabomon thanks for the PR.  Looks great.  Can you merge with master so we get a fresh CI build and we can check out why this is failing. I suspect something with the init registration.

src/openfermion/transforms/__init__.py
xabomon: Hi Nicholas,

Thanks for the review, I'm happy to see that most of the things are in agreement with the code standards.

The __init__ is in fact wrong. We originally put these code in transformations, but at the end decided to move it to utils, hence the transforms init is wrong, and the one in utils must be updated.
This will be done for the merge.

Best,
Xavi

ncrubin:
@xabomon looks like on of your tests is still attempting to import `reduce_number_of_terms` from transforms and not utils.

xabomon:
Hi @ncrubin,

Sorry for this, I just fixed the import in the test module.

Best,
Xavi

xabomon:
Hi @ncrubin ,

After yesterday's commit I saw in the CI that one of the test was failing in the python2.7 version because I was defining it as an empty list, after changing the mutable variables to None.

I updated the test module that includes the new possible case, and change the old one. This should now have no problem in the CI.

Let me know if anything else needs to be updated.

Best,
Xavi

ncrubin:
I believe the rules for implementing `super()` are different in python2.7.  I can look this weekend to resolve this if you don't solve by then!

xabomon:
Hi Nick,

Don’t worry about it! I can’t look at it during the weekend, but I will do it next week.

Thanks,
Xavi

ncrubin:
@xabomon bump on this?  it would be great to get this in!

xabomon:
Hi Nick,

I'm really sorry for not being active on this lately.

I manage to narrow down that the test fail in python 2.7. I am suspicious that the fail is due to our customize error messages but I have seen code within OpenFermion with customize errors.

Currently I'm building an environment with 2.7 to see if I can reproduce the issue outside the test module.

Please, let me know if you have heard about similar issues before and how to address them.

Best,
Xavi

xabomon:
Hi @ncrubin ,
We have fixed the CI issue but now the coverage is failing.

As far as I understand, the coveralls counts how many new lines are used in the test module. We have tested all possible issue in the main functions, this implies that internal functions are indirectly tested. Do we have to also cover internal functions?

Thanks for the assistance.

Xavi

xabomon:
I finally managed to fulfill CI and coverage.

Hopefully, we can now merge this PR.

Best,
Xavi

PR #518, opened 2019-05-16
Add link to Quantum Computing Stack Exchange so questions may be dire…

vtomole:
…cted there instead of the issue tracker.


babbush:
I'm okay with this but since OpenFermion core does not explicitly depend on cirq, perhaps we should not mention cirq here.

PR #523, opened 2019-05-23
Adds a first draft of tools to work with number preserving sparse operators

Wuggins:
These are the tools (and the tests for them) that I wrote to create scipy sparse operators directly in some number (and possibly spin / excitation rank) subspace. Feedback about what to do to get them ready to pull into OpenFermion would be great.

Thanks,

Bill


src/openfermion/transforms/_conversion_test.py
```diff
@@ -507,3 +513,374 @@ def test_get_fermion_operator_majorana_operator():
...
+        sparse_ham = get_number_preserving_sparse_operator(
+            hamiltonian_fop,
+            self.molecule.n_qubits,
+            self.molecule.n_electrons,
+            spin_preserving=False,
+            excitation_level=2)
+
+        eig_val, eig_vec = scipy.sparse.linalg.eigsh(sparse_ham, k=1, which='SA')
+
+        assert numpy.abs(eig_val[0] - self.molecule.cisd_energy) < 1E-9
```
kevinsung: Add a test for the Hubbard model.

src/openfermion/transforms/_conversion_test.py
```diff
@@ -507,3 +513,374 @@ def test_get_fermion_operator_majorana_operator():
...
+        one_body_regular_sparse_op = get_sparse_operator(one_body_fop)
+
+        make_hf_fop = FermionOperator(((3, 1), (2, 1), (1, 1), (0, 1)))
+        make_hf_sparse_op = get_sparse_operator(make_hf_fop, n_qubits=12)
+
+        hf_state = numpy.zeros((2**12))
+        hf_state[0] = 1.0
+        hf_state = make_hf_sparse_op.dot(hf_state)
+
+        regular_sparse_hf_energy = (hf_state @ (one_body_regular_sparse_op.dot(hf_state))).real
```
kevinsung: We need to support Python 2, which doesn't have the `@` operator.

src/openfermion/transforms/_conversion_test.py
```diff
@@ -507,3 +513,374 @@ def test_get_fermion_operator_majorana_operator():
...
+            space_size = state_array.shape[0]
+            if do_print:
+                print("Dimension of Hilbert space for sparse operator = %d" % space_size)
+
+            reference = numpy.zeros((space_size))
+            reference[0] = 1.0
+
+            return reference
+
+        self.get_reference_state = get_reference_state
```
kevinsung: Why not just define this function as a method of the class?

src/openfermion/transforms/_conversion_test.py
```diff
@@ -507,3 +513,374 @@ def test_get_fermion_operator_majorana_operator():
...
+            spin_preserving=True)
+
+        reference = self.get_reference_state(
+            self.molecule.n_qubits,
+            self.molecule.n_electrons,
+            spin_preserving=True)
+
+        sparse_hf_energy = reference @ (one_body_sparse_op.dot(reference))
+
+        assert numpy.linalg.norm(sparse_hf_energy - regular_sparse_hf_energy) < 1E-9
```
kevinsung: Add a test where you construct the sparse operator in two different ways: using your new method, and using `get_sparse_operator` combined with `jw_number_restrict_operator` and `jw_sz_restrict_operator`. Verify that the entire spectrums match.

ncrubin:
@Wuggins This would be great to have but no immediate pressing need.  Let me know what you'd like to do with this PR.  I am happy to keep it open for the time being if you want to do more development.

src/openfermion/transforms/_conversion_test.py
Wuggins: You know, this isn't really doing anything anymore except just making a vector [1.0, 0.0, ... , 0.0] so I just put that code in the tests where I used it and removed this entirely.

src/openfermion/transforms/_conversion_test.py
Wuggins: This is a really great test!

Wuggins:
Hey @kevinsung, are there other changes I should make?

kevinsung:
You missed a few of my comments; perhaps you need to load the "hidden conversations" to see them?

Wuggins:
@kevinsung. Thank you, that was a silly mistake to make! I fixed those remaining things and have pushed the changes again.

src/openfermion/transforms/_conversion.py
```diff
@@ -556,3 +557,331 @@ def get_boson_operator(operator, hbar=1.):
...
+
+    state_array = numpy.asarray(list(_iterate_basis_(
+        reference_determinant, excitation_level, spin_preserving)))
+    # Create a 1d array with each determinant encoded as an integer for sorting purposes.
+    int_state_array = state_array.dot(1 << numpy.arange(state_array.shape[1])[::-1])
+    sorting_indices = numpy.argsort(int_state_array)
+
+    space_size = state_array.shape[0]
+
+    fermion_op = normal_ordered(fermion_op)
```
kevinsung: Is the normal-ordering necessary? If not we should avoid it because it could be expensive.

src/openfermion/transforms/_conversion_test.py
```diff
@@ -507,3 +517,490 @@ def test_get_fermion_operator_majorana_operator():
...
+    def test_hf_energy(self):
+        hamiltonian_fop = get_fermion_operator(self.molecular_hamiltonian)
+
+        sparse_ham = get_number_preserving_sparse_operator(
+            hamiltonian_fop,
+            self.molecule.n_qubits,
+            self.molecule.n_electrons,
+            spin_preserving=True)
+
+        space_size = sparse_ham.shape[0]
```
kevinsung: Add a test that the resulting space size is what you'd expect (i.e. much smaller than a non-number-preserving operator).

src/openfermion/transforms/_conversion.py
```diff
@@ -556,3 +557,331 @@ def get_boson_operator(operator, hbar=1.):
                         "supported for get_boson_operator.")
 
     return boson_operator
+
+
+def get_number_preserving_sparse_operator(
+        fermion_op,
+        num_qubits,
+        num_electrons,
+        spin_preserving=True,
```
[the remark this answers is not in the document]
Wuggins: Done

src/openfermion/transforms/_conversion.py
Wuggins: It dramatically simplifies the logic of other pieces and is necessary for the current implementation. With normal ordered terms we can just look at the presence or absence of certain operators and immediately infer whether or not a specific determinant has a non-zero matrix element.

Wuggins: I would love to take on speeding up normal ordering in OpenFermion as a pet project though. I agree that it could be quite painful even for an otherwise reasonable operator.

src/openfermion/transforms/_conversion_test.py
Wuggins: Done. I specifically check that space_size is what should be expected (and this number is smaller than the naive size of course).

PR #530, opened 2019-07-23
Update other_projects.md

bsenjean:
Added the OpenFermion-Dirac interface and our recent article that used it.


bsenjean:
I signed it !

PR #535, opened 2019-08-31
Adapt RDM to spin-orbital representation

rugantio:
This PR aims at solving the problem presented in issue:
https://github.com/quantumlib/OpenFermion-PySCF/issues/49

At the moment, the RDMs are processed incorrectly by ``` get_molecular_rdm ```, lacking the split of spin-orbitals needed for a proper quantum computing mapping. A new function: ``` get_molecular_rdm_spin ``` aims to accomplish this task. 

I am not really sure with the correctness of the function, since I don't know the convention used to store the spin-orbitals RDM in the hdf5 files, but it's a good starting point to solve this issue and doesn't break previous code. 

Here's a possible way to check the calculations (useful for further unit tests):
```
import numpy as np
from openfermion.hamiltonians import MolecularData
from openfermionpyscf import run_pyscf

# compute molecule
molecule = MolecularData([('H', (0,0,0)),('H', (0,0,0.7414))],'sto-3g', 1)
molecule = run_pyscf(molecule,
                     run_scf=1,
                     run_fci=1)
h2_rdm_spin = molecule.get_molecular_rdm_spin(use_fci=True) 

## load stored RDM for comparison
THIS_DIRECTORY = './Downloads/OpenFermion'
filename = os.path.join(THIS_DIRECTORY, 'src','openfermion','data','H2_sto-3g_singlet_0.7414.hdf5')
molecule = MolecularData(filename=filename)
molecule.load()
h2_rdm_old = molecule.get_molecular_rdm(use_fci=True)

# check calculations
np.allclose(h2_rdm.one_body_tensor,h2_rdm_old.one_body_tensor,atol=1e-6,rtol=0)
np.allclose(h2_rdm.one_body_tensor,h2_rdm_old.one_body_tensor,atol=1e-6,rtol=0)
```

### DIFF
--- a/src/openfermion/hamiltonians/_molecular_data.py
+++ b/src/openfermion/hamiltonians/_molecular_data.py
@@ -933,6 +933,61 @@ class MolecularData(object):
         # Cast to InteractionRDM class.
         rdm = InteractionRDM(one_rdm, two_rdm)
         return rdm
+    
+    def get_molecular_rdm_spin(self, use_fci=False):
+        """Method to return 1-RDM and 2-RDMs from CISD or FCI, adapted for
+           spin-orbitals.
+
+        Args:
+            use_fci: Boolean indicating whether to use RDM from FCI
+                calculation.
+
+        Returns:
+            rdm: An instance of the MolecularRDM class.
+
+        Raises:
+            MisissingCalculationError: If the CI calculation has not been
+                performed.
+        """
+        # Make sure requested RDM has been computed and load.
+        if use_fci:
+            if self.fci_energy is None:
+                raise MissingCalculationError(
+                    'Missing FCI RDM in {}'.format(self.filename) +
+                    'Run FCI calculation before loading FCI RDMs.')
+            else:
+                one_rdm = self.fci_one_rdm
+                two_rdm = self.fci_two_rdm
+        else:
+            if self.cisd_energy is None:
+                raise MissingCalculationError(
+                    'Missing CISD RDM in {}'.format(self.filename) +
+                    'Run CISD calculation before loading CISD RDMs.')
+            else:
+                one_rdm = self.cisd_one_rdm
+                two_rdm = self.cisd_two_rdm
+
+        one_rdm = numpy.kron(one_rdm/2,numpy.eye(2))
+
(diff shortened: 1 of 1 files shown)


rugantio:
@googlebot I signed it!

rugantio:
Hi @ncrubin thank for addressing the PR. I tested my function also on LiH, since it is conveniently stored as hdf5. I noticed that the RDMs stored are a bit different from the ones that I obtained with the new function ```get_molecular_rdm_spin```, despite producing the same final state energy. I also checked with other molecules (BeH2, H2O) and it seems consistent with the results. Nonetheless, could please check that this function is correct? Here's the code I used as unit test:
```
geometry = [('Li', (0., 0., 0.)), ('H', (0., 0., 1.45))]
molecule = MolecularData(geometry,'sto-3g',1)
molecule = run_pyscf(molecule,
                     run_scf=1,
                     run_fci=1)
#molecule = MolecularData(geometry, 'sto-3g', description="1.45")
#molecule.load()

molecular_hamiltonian = molecule.get_molecular_hamiltonian()
rdm = molecule.get_molecular_rdm_spin(use_fci=True)
#rdm = molecule.get_molecular_rdm(use_fci=True)


# Test energy of RDM.
rdm_energy = molecular_hamiltonian.constant
rdm_energy += np.sum(rdm.one_body_tensor *
                     molecular_hamiltonian.one_body_tensor)
rdm_energy += np.sum(rdm.two_body_tensor *
                     molecular_hamiltonian.two_body_tensor)
print('RDM energy: ',rdm_energy)
print('FCI energy: ',molecule.fci_energy)
np.isclose(rdm_energy,molecule.fci_energy)
```

ncrubin:
Hi @rugantio ,  I believe this code will only work for singlet states.  I think if you try this for an openshell molecule the code will return the wrong energy.  

I believe the root cause is that pyscf is returning only the spin-summed RDMs.  I believe the correct fix to be finding the appropriate pyscf functions that will return the non spin-summed RDMs.

ncrubin:
@rugantio I am closing this PR for now since we haven't had motion on it in awhile.   I can reopen if need be.

PR #538, opened 2019-09-06
Enhancement of Project onto sector function.

xabomon:
Hi all,

After some discussion with the user @rugantio I decided to improve the existing functions project_onto_sector and projection_error.

The main enhancement consists on checking that the sectors are passed as 0 or 1 (which is clearly documented) instead of 1 or -1.
Additionally, I added some trivial assertions and update the test such that the assertions are tested.

Please, let me know if anything else is required.

PS: A previous PR didn't have the commits, sorry.

Best,
Xavi


ncrubin:
Thanks @xabomon!

PR #540, opened 2019-10-11
Added qubit partition functions

obriente:
An implementation of the partial state tomography methods for qubits that we introduced in https://arxiv.org/pdf/1908.05628.pdf .

### DIFF
--- a/src/openfermion/measurements/_qubit_partitioning.py
+++ b/src/openfermion/measurements/_qubit_partitioning.py
@@ -20,18 +20,25 @@ except ImportError:
     from itertools import izip_longest as zip_longest
 
 
-def binary_partition_iterator(qubit_list, num_partitions=None):
+def binary_partition_iterator(qubit_list, num_iterations=None):
     '''Generator for a list of 2-partitions of N qubits
     such that all pairs of qubits are split in at least one partition,
-    following ArXiv:1908.05628
+    This follows a variation on ArXiv:1908.0562 - instead of
+    explicitly partitioning the list based on the binary indices of
+    the qubits, we repeatedly divide the list in two and then
+    zip it back together.
 
     Args:
         qubit_list(list): list of qubits to be partitioned
+        num_iterations(int or None): number of iterations to perform.
+            If None, will be set to ceil(log2(len(qubit_list)))
 
     Returns:
         partition(iterator of lists of lists): the required partitioning
     '''
-    if num_partitions is not None and num_partitions == 0:
+
+    # Some edge cases
+    if num_iterations is not None and num_iterations == 0:
         return
     num_qubits = len(qubit_list)
     if num_qubits < 2:
@@ -40,22 +47,29 @@ def binary_partition_iterator(qubit_list, num_partitions=None):
         yield [[qubit_list[0]], [qubit_list[1]]]
         return
 
-    num_partitions = num_partitions or\
+    num_iterations = num_iterations or\
         int(numpy.ceil(numpy.log2(num_qubits)))
 
-    partition = [qubit_list[:num_qubits//2], qubit_list[num_qubits//2:]]
+    # Calculate the point where we need to split the list each time.
+    half_point = int(numpy.ceil(num_qubits/2))
 
-    for j in range(num_partitions):
+    # Repeat the division and zip steps as many times
+    # as required.
+    for j in range(num_iterations):
+        # Divide the qubit list in two and return it
+        partition = [qubit_list[:half_point],
+                     qubit_list[half_point:]]
         yield partition
-        merged_partition = list(
-            chain(*zip_longest(partition[1], partition[0])))
-        if merged_partition[-1] is None:
-            del merged_partition[-1]
-        partition = [merged_partition[:num_qubits//2],
-                     merged_partition[num_qubits//2:]]
+        # Zip the partition together to remake the qubit list.
+        qubit_list = list(chain(*zip_longest(partition[0], partition[1])))
+        # If len(qubit_list) is odd, the end of the list will be 'None'
+        # which we delete.
+        if qubit_list[-1] is None:
+            del qubit_list[-1]
+        
 
 
-def partition_iterator(qubit_list, partition_size, num_partitions=None):
+def partition_iterator(qubit_list, partition_size, num_iterations=None):
     '''Generator for a list of k-partitions of N qubits such that
     all sets of k qubits are perfectly split in at least one
     partition, following ArXiv:1908.05628
@@ -63,20 +77,23 @@ def partition_iterator(qubit_list, partition_size, num_partitions=None):
     Args:
         qubit_list(list): list of qubits to be partitioned
         partition_size(int): the number of partitions.
+        num_iterations(int or None): the number of iterations in the
+            outer iterator. If None, set to ceil(log2(len(qubit_list)))
 
     Returns:
         partition(iterator of lists of lists): the required partitioning
     '''
-    if num_partitions is not None and num_partitions == 0:
+
+    # Some edge cases
+    if num_iterations is not None and num_iterations == 0:
         return
     if partition_size == 1:
         yield [qubit_list]
         return
     elif partition_size == 2:
-        for p in binary_partition_iterator(qubit_list, num_partitions):
+        for p in binary_partition_iterator(qubit_list, num_iterations):
             yield p
         return
-
     num_qubits = len(qubit_list)
     if partition_size == num_qubits:
         yield [[q] for q in qubit_list]
@@ -84,28 +101,36 @@ def partition_iterator(qubit_list, partition_size, num_partitions=None):
     elif partition_size > num_qubits:
         raise ValueError('I cant k-partition less than k qubits')
 
-    num_partitions = num_partitions or\
+
+    num_iterations = num_iterations or\
         int(numpy.ceil(numpy.log2(num_qubits)))
 
+    # First iterate over the outer binary partition
     outer_iterator = binary_partition_iterator(
-        qubit_list, num_partitions=num_partitions)
-
+        qubit_list, num_iterations=num_iterations)
     for p1, p2 in outer_iterator:
-        num_partitions -= 1
 
+        # Each new partition needs to be subdivided fewer times
+        # to prevent an additional k! factor in the scaling.
+        num_iterations -= 1
+
+        # Iterate over all possibilities of subdividing the first
+        # partition into l partitions and the second partition into
+        # k - l partitions.
         for inner_partition_size in range(1, partition_size):
             if inner_partition_size > len(p1) or\
                     partition_size - inner_partition_size > len(p2):
                 continue
 
+            # subdivide the first partition
             inner_iterator1 = partition_iterator(
-                p1, inner_partition_size, num_partitions)
-
+                p1, inner_partition_size, num_iterations)
             for ips1 in inner_iterator1:
+
+                # subdivide the second partition
                 inner_iterator2 = partition_iterator(
                     p2, partition_size-inner_partition_size,
-                    num_partitions)
-
+                    num_iterations)
                 for ips2 in inner_iterator2:
                     yield ips1 + ips2
 
--- a/src/openfermion/measurements/_qubit_partitioning.py
+++ b/src/openfermion/measurements/_qubit_partitioning.py
@@ -13,7 +13,11 @@
 
 """ Code to generate Pauli strings for measurement of local operators"""
 import numpy
-from itertools import chain, zip_longest
+from itertools import chain
+try:
+    from itertools import zip_longest
+except ImportError:
+    from itertools import izip_longest as zip_longest
 
 
 def binary_partition_iterator(qubit_list, num_partitions=None):
--- a/src/openfermion/measurements/_qubit_partitioning.py
+++ b/src/openfermion/measurements/_qubit_partitioning.py
@@ -22,7 +22,7 @@ except ImportError:
 
 
 def binary_partition_iterator(qubit_list, num_iterations=None):
-    '''Generator for a list of 2-partitions of N qubits
+    """Generator for a list of 2-partitions of N qubits
     such that all pairs of qubits are split in at least one partition,
     This follows a variation on ArXiv:1908.0562 - instead of
     explicitly partitioning the list based on the binary indices of
@@ -36,7 +36,7 @@ def binary_partition_iterator(qubit_list, num_iterations=None):
 
     Returns:
         partition(iterator of lists of lists): the required partitioning
-    '''
+    """
 
     # Some edge cases
     if num_iterations is not None and num_iterations == 0:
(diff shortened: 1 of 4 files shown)


src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,141 @@
...
+try:
+    from itertools import zip_longest
+except ImportError:
+    from itertools import izip_longest as zip_longest
+
+
+def binary_partition_iterator(qubit_list, num_partitions=None):
+    '''Generator for a list of 2-partitions of N qubits
+    such that all pairs of qubits are split in at least one partition,
+    following ArXiv:1908.05628
```
kevinsung: The code in this function does not follow the paper as far as I can tell; the scheme in the paper is based on using the binary representations of the qubit indices. Is there a reason you don't follow the paper? It's difficult for me to verify the correctness of the algorithm you give here.

obriente: Sorry for not detailing the difference - I coded it this way because the output is more balanced, and because this zipping method seems faster/more pythonic than trying to deal with binary division in python.

The code does follow the paper in the case where len(qubit_list) = 2^n. By cutting the list in two and zipping it back together we in effect shuffle the indices.

For example, consider the list [1, 2, 3, 4, 5, 6, 7, 8]. Both the code and the paper first iterate to generate [[1, 2, 3, 4], [5, 6, 7, 8]].
Then, in the paper, we should take the second index, i.e. [[1, 2] + [5, 6], [3, 4] + [7, 8]].
In the code, we zip the list together to generate [1, 5, 2, 6, 3, 7, 4, 8], and then split this again in 2 to obtain [[1, 5, 2, 6], [3, 7, 4, 8]]. As we don't care about order in the lists, the two sets are equivalent.

When len(qubit_list) != 2^n, the code and the paper differ, but the result of the code is better as it is more balanced. For example, consider the list [1, 2, 3, 4, 5] - the paper would want to split this initially into [[1, 2, 3, 4], [5]], while the code would split this into [[1, 2, 3], [4, 5]]. The former would be less preferable in an experiment as we wind up making more repeat measurements of the form XX/YY/ZZ on pairs of qubits.

The new code also has the nice effect for k>2 that the partitions are ordered 'correctly' for the sub-partitioning in Fig.1(b) of the paper. For example, if I look again at [1,2,3,4,5,6,7,8], after partitioning into [[1, 2] + [5, 6], [3, 4] + [7, 8]], the scheme in the paper requires me to partition [1, 2, 5, 6] into [1] + [5], [2] + [6]. In the code, as our list is ordered already as [1, 5, 2, 6], we can repeat the direct division into half.

I've updated the tests to check that the code behaves appropriately for k=3 and N=1 to 15 (previously it only checked for k=3, N=8) - this is the 'test_partition_three' function in the test suite. This checks that for each index [i, j, k] we find one partition that exactly divides them, as required. I've also added some comments for clarity. Hope this helps.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # Calculate the point where we need to split the list each time.
+    half_point = int(numpy.ceil(num_qubits/2))
+
+    # Repeat the division and zip steps as many times
+    # as required.
+    for j in range(num_iterations):
+        # Divide the qubit list in two and return it
+        partition = [qubit_list[:half_point],
+                     qubit_list[half_point:]]
```
kevinsung: Yield a tuple instead of a list.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def partition_iterator(qubit_list, partition_size, num_iterations=None):
+    '''Generator for a list of k-partitions of N qubits such that
+    all sets of k qubits are perfectly split in at least one
+    partition, following ArXiv:1908.05628
+
+    Args:
+        qubit_list(list): list of qubits to be partitioned
+        partition_size(int): the number of partitions.
```
kevinsung: "the number of sets in a partition"

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+import numpy
+from itertools import chain
+try:
+    from itertools import zip_longest
+except ImportError:
+    from itertools import izip_longest as zip_longest
+
+
+def binary_partition_iterator(qubit_list, num_iterations=None):
+    '''Generator for a list of 2-partitions of N qubits
```
kevinsung: Use double quotes instead of single quotes.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+    Returns:
+        partition(iterator of lists of lists): the required partitioning
+    '''
+
+    # Some edge cases
+    if num_iterations is not None and num_iterations == 0:
+        return
+    if partition_size == 1:
+        yield [qubit_list]
```
kevinsung: Yield a tuple.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        partition_size(int): the number of partitions.
+        num_iterations(int or None): the number of iterations in the
+            outer iterator. If None, set to ceil(log2(len(qubit_list)))
+
+    Returns:
+        partition(iterator of lists of lists): the required partitioning
+    '''
+
+    # Some edge cases
+    if num_iterations is not None and num_iterations == 0:
```
kevinsung: Just
```
if num_iterations == 0:
```

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        return
+    num_qubits = len(qubit_list)
+    if num_qubits < 2:
+        raise ValueError('Need at least 2 qubits to partition')
+    if num_qubits == 2:
+        yield [[qubit_list[0]], [qubit_list[1]]]
+        return
+
+    num_iterations = num_iterations or\
+        int(numpy.ceil(numpy.log2(num_qubits)))
```
kevinsung: Checking for the None case looks better than doing a line continuation:
```
if num_iterations is None:
    num_iterations = ...
```

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+    the qubits, we repeatedly divide the list in two and then
+    zip it back together.
+
+    Args:
+        qubit_list(list): list of qubits to be partitioned
+        num_iterations(int or None): number of iterations to perform.
+            If None, will be set to ceil(log2(len(qubit_list)))
+
+    Returns:
+        partition(iterator of lists of lists): the required partitioning
```
kevinsung: I think it's slightly better to yield tuples of lists instead of lists of lists.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+    if partition_size == 1:
+        yield [qubit_list]
+        return
+    elif partition_size == 2:
+        for p in binary_partition_iterator(qubit_list, num_iterations):
+            yield p
+        return
+    num_qubits = len(qubit_list)
+    if partition_size == num_qubits:
+        yield [[q] for q in qubit_list]
```
kevinsung:
```
yield tuple([q] for q in qubit_list)
```

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        return
+    elif partition_size > num_qubits:
+        raise ValueError('I cant k-partition less than k qubits')
+
+
+    num_iterations = num_iterations or\
+        int(numpy.ceil(numpy.log2(num_qubits)))
```
kevinsung: Checking for the None case looks better than doing a line continuation:
```
if num_iterations is None:
    num_iterations = ...
```

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+    # First iterate over the outer binary partition
+    outer_iterator = binary_partition_iterator(
+        qubit_list, num_iterations=num_iterations)
+    for p1, p2 in outer_iterator:
```
kevinsung: I think you chose the letter p for partition, but these are really sets in a partition. So I suggest renaming to `set1` and `set2`.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+        # Each new partition needs to be subdivided fewer times
+        # to prevent an additional k! factor in the scaling.
+        num_iterations -= 1
+
+        # Iterate over all possibilities of subdividing the first
+        # partition into l partitions and the second partition into
+        # k - l partitions.
```
kevinsung: I think you're mixing up the word "partition" with "set in partition".  Perhaps you mean
```
        # Iterate over all possibilities of partitioning the first
        # set into l parts and the second set into k - l parts.
```

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        for inner_partition_size in range(1, partition_size):
+            if inner_partition_size > len(p1) or\
+                    partition_size - inner_partition_size > len(p2):
+                continue
+
+            # subdivide the first partition
+            inner_iterator1 = partition_iterator(
+                p1, inner_partition_size, num_iterations)
+            for ips1 in inner_iterator1:
```
kevinsung: Better variable name for `ips1`.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+
+                # subdivide the second partition
+                inner_iterator2 = partition_iterator(
+                    p2, partition_size-inner_partition_size,
+                    num_iterations)
+                for ips2 in inner_iterator2:
```
kevinsung: Better variable name for `ips2`.

src/openfermion/measurements/_qubit_partitioning_test.py
```diff
@@ -0,0 +1,181 @@
...
+    def test_partition_three(self):
+        for num_qubits in range(1, 16):
+            qubit_list = list(range(num_qubits))
+            for i in range(num_qubits):
+                for j in range(i+1, num_qubits):
+                    for k in range(j+1, num_qubits):
+                        print('Testing {}, {}, {}'.format(i, j, k))
+                        pi = partition_iterator(qubit_list, 3)
+                        count = 0
+                        for p1, p2, p3 in pi:
```
kevinsung: Check that p1, p2, and p3 are a valid partition, i.e., that their disjoint union is the same set as qubit_list.

src/openfermion/measurements/_qubit_partitioning_test.py
```diff
@@ -0,0 +1,181 @@
...
+                            print('Partition obtained: ', p1, p2, p3)
+                            if max([sum([1 for x in p if x in [i, j, k]])
+                                    for p in [p1, p2, p3]]) == 1:
```
kevinsung: The brackets `[` are unnecessary in the arguments to `max` and `sum`.

src/openfermion/measurements/_qubit_partitioning_test.py
obriente: Good idea, thanks for the suggestion

src/openfermion/measurements/_qubit_partitioning_test.py
obriente: Thanks, I didn't know that

src/openfermion/measurements/_qubit_partitioning.py
obriente: Yes, indeed, thanks for resolving the confusion

src/openfermion/measurements/_qubit_partitioning.py
obriente: Good idea, done

src/openfermion/measurements/_qubit_partitioning.py
obriente: I've changed it to 'inner_partition1'

src/openfermion/measurements/_qubit_partitioning.py
obriente: I've changed it to 'inner_partition2'

src/openfermion/measurements/_qubit_partitioning.py
obriente: Changed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed, thanks

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        qubit_list(list): list of qubits to be partitioned
+        num_iterations(int or None): number of iterations to perform.
+            If None, will be set to ceil(log2(len(qubit_list)))
+
+    Returns:
+        partition(iterator of lists of lists): the required partitioning
+    '''
+
+    # Some edge cases
+    if num_iterations is not None and num_iterations == 0:
```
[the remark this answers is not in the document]
obriente: Thanks, fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Sounds good, I've done this throughout the three functions in the file and added some checks in the test suite to ensure that they are indeed returning tuples.

src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -0,0 +1,167 @@
...
+        return
+    num_qubits = len(qubit_list)
+    if num_qubits < 2:
+        raise ValueError('Need at least 2 qubits to partition')
+    if num_qubits == 2:
+        yield [[qubit_list[0]], [qubit_list[1]]]
```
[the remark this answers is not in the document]
obriente: Done

kevinsung:
Hi @obriente , I think you missed a few of my comments. Perhaps you need to expand the "hidden conversations" to see them?

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

src/openfermion/measurements/_qubit_partitioning.py
obriente: Fixed

obriente:
Indeed I did - sorry about that. I think I've got all of them now.

PR #541, opened 2019-11-13
Hamiltonian separation by stabilizers and QSE tools

xabomon:
Hi all,

This PR contains two new modules added to measurements.

1. Hamiltonian split by stabilizers
- A Hamiltonian is split into subsets of Pauli operators that are affected when a stabilizer is measured detecting an error.
- This module allows one to find the subsets of Pauli operators unaffected after detecting an error using the stabilizers. Hopefully, this will allow to reduce the number of measurements to be performed while using symmetry verification/error detection.

2. Quantum subspace expansion
- I added a module that calculates QSE, including a function that returns a list of additional operators to measure to be able to perform the expansion.
- The test module of these functions requires estimated expectation values, hence I used calculations of my own to generate such estimated Hamiltonians. If a better way of doing the test exists I'll be happy to change it.

PS: I will not be able to work on improvements of the PR in the upcoming weeks, but as soon as I can I will make all the necessary changes. Sorry about it.
Best,
Xavi

### DIFF
--- a/src/openfermion/measurements/__init__.py
+++ b/src/openfermion/measurements/__init__.py
@@ -15,6 +15,11 @@ from ._equality_constraint_projection import (apply_constraints,
                                               linearize_term,
                                               unlinearize_term)
 
+from ._hamiltonian_stabilizers_sets import (get_hamiltonian_subsets)
+
+from ._quantum_subspace_expansion import (calculate_qse_spectrum,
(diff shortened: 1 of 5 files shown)


kevinsung:
Please split this PR into two separate PRs, each introducing just one of the modules.

PR #544, opened 2019-11-29
Add functionality to retrieve specific types of structure info from PubChem

hay-k:
Currently, the function **geometry_from_pubchem** in **utils/_pubchem.py/** constructs the geometry of a molecule based on 3D structure info if available, otherwise the 2D structure info is used. If both 3D and 2D info are not available in PubChem, None is returned.

In this pull request I have implemented a new argument for this functiion: structure. With this argument users of the function will be able to request _only 3D_ or _only 2D_ information. The default behavior of the function is unchanged, i.e. if the 'structure' argument is not specified, the function will behave as before this pull request, for backwards compatibility.

The reason for the implementation of this argument is, that for a bunch of molecules, the 2D structure information has nothing to do with the actual geometric sizes of the molecule. And, when 3D is not available (e.g for H2), the default behavior will return incorrect geometry based on the non-accurate 2D structure info. It is sometimes better to return nothing, than return incorrect data. With this new option users will be able to force the function to use e.g. only 3D structure info. 

PubChem support confirms, that the 2D structure info is not guaranteed to reflect actual information about molecules. Here is a snipped from their reply:

> PubChem 2-D structures have coordinates generated by a layout algorithm.  They do not reflect anything other than that to help provide a depiction that is a suitable distance from each other.  Yes, they could be normalized in some way (e.g., 0 and 1 or -0.5 and 0.5) but they are arbitrary, in a sense.
[Re_ [Pubchem-help] 2D coordinates of compounds_a.pdf](https://github.com/quantumlib/OpenFermion/files/3904024/Re_.Pubchem-help.2D.coordinates.of.compounds_a.pdf)

My full email conversation with PubChem support is attached!

### DIFF
--- a/src/openfermion/utils/_pubchem.py
+++ b/src/openfermion/utils/_pubchem.py
@@ -11,13 +11,20 @@
 #   limitations under the License.
 
 
-def geometry_from_pubchem(name):
+def geometry_from_pubchem(name, structure=None):
     """Function to extract geometry using the molecule's name from the PubChem
-    database.
+    database. The 'structure' argument can be used to specify which structure
+    info to use to extract the geometry. If structure=None, the geometry will
+    be constructed based on 3D info, if available, otherwise on 2D (to keep
+    backwards compatibility with the times when the argument 'structure'
+    was not implemented).
 
     Args:
         name: a string giving the molecule's name as required by the PubChem
             database.
+        structure: a string '2d' or '3d', to specify a specific structure
+            information to be retrieved from pubchem. The default is None.
+            Recommended value is '3d'.
 
     Returns:
         geometry: a list of tuples giving the coordinates of each atom with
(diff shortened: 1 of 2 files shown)


hay-k:
@googlebot I signed it!

src/openfermion/utils/_pubchem.py
```diff
@@ -11,43 +11,47 @@
...
     """
     import pubchempy
 
-    pubchempy_2d_molecule = pubchempy.get_compounds(name, 'name',
-                                                    record_type='2d')
-
-    # Check if 2-D geometry is available. If not then no geometry is.
-    if not pubchempy_2d_molecule:
-        print('Unable to find molecule in the PubChem database.')
+    if structure:
```
ncrubin: explicitly check if None since multiple invalid inputs will work here.

src/openfermion/utils/_pubchem.py
```diff
@@ -11,43 +11,47 @@
...
+        pubchempy_molecule = pubchempy.get_compounds(name, 'name',
+                                                     record_type=structure)
+    else:
+        # Ideally get the 3-D geometry if available.
+        pubchempy_molecule = pubchempy.get_compounds(name, 'name',
+                                                        record_type='3d')
+
+        # If the 3-D geometry isn't available, get the 2-D geometry instead.
+        if not pubchempy_molecule:
```
ncrubin: Check if None.

src/openfermion/utils/_pubchem.py
```diff
@@ -11,43 +11,47 @@
...
+            pubchempy_molecule = pubchempy.get_compounds(name, 'name',
+                                                        record_type='2d')
+
+    if not pubchempy_molecule:
```
ncrubin: ditto above.

ncrubin:
@hay-k Thanks for the submission and clarifying this!  This is really helpful to get cleared up and make sure people are using pubchem in a way that is reasonable.  I've only requested minor changes to make.

src/openfermion/utils/_pubchem.py
hay-k: Hi @ncrubin 
Shall we raise a ValuError whenever the structure argument is not '2d', '3d' or None? Or shall we just print a message and return None (like it was done in the case of "Unable to find ...")?

src/openfermion/utils/_pubchem.py
hay-k: The line "_if not pubchempy_molecule:_" actually should not check only the case of None, but should check if **pubchempy_molecule** is None or empty (an empty list).

ncrubin: Ah sorry about that.  This is fine as is.  Maybe a type hint here would be helpful (at least for me in the future!).  I didn't look up that get_compound returns a list.

src/openfermion/utils/_pubchem.py
hay-k: ditto below.

src/openfermion/utils/_pubchem.py
hay-k: I went with ValueError approach.

hay-k:
Hi @ncrubin, I have commited the changes.

src/openfermion/utils/_pubchem.py
```diff
@@ -11,43 +11,50 @@
...
+        # Ideally get the 3-D geometry if available.
+        pubchempy_molecule = pubchempy.get_compounds(name, 'name',
+                                                        record_type='3d')
+
+        # If the 3-D geometry isn't available, get the 2-D geometry instead.
+        if not pubchempy_molecule:
+            pubchempy_molecule = pubchempy.get_compounds(name, 'name',
+                                                        record_type='2d')
+    else:
+        raise ValueError('Incorrect value for the argument structure=%s' % structure)
```
ncrubin: It seems this line is not covered by a test.  You can construct this test by mocking a get_compounds method from pubchempy which returns something other than ['2d', '3d', None] and then checking the ValueError with pytest.raises.

ncrubin: On second thought...I'll just merge this now.  Coveralls doesn't seem to mind.   I'll add a test to this later.

PR #545, opened 2019-12-03
Hamiltonian split by stabilizers

xabomon:
Hi all,
I split the PR that I submitted a couple of weeks ago.

Here I add a module that allows one to find the subsets of Pauli operators of a Hamiltonian that have support on a set of stabilizers.

1. A Hamiltonian is split into subsets of Pauli operators that are affected when a stabilizer is measured detecting an error.
2. This module allows one to find the subsets of Pauli operators unaffected after detecting an error using the stabilizers. Hopefully, this will allow to reduce the number of measurements to be performed while using symmetry verification/error detection.

The code comes fully documented and tested as requested by OpenFermion guidelines.

Best,
Xavi


src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+        if qp == ():
+            overlap = False
+        if qp[0] in qbts_in_stab:
+            overlap = True
+    return overlap
+
+
+def _check_missing_paulis(hamiltonian, subsets):
+    """
+    Auxiliar function of Hamiltonian subset splitting.
```
kevinsung: "Auxiliary"

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+
+"""Tools to split the Hamiltonian into subsets based on the support of
+its stabilizers."""
+
+import numpy
+import copy
+from openfermion.ops import QubitOperator
+
+
+def _check_stabilizer_overlap(pauli_op, stabilizer):
```
kevinsung: This function doesn't really have anything to do with stabilizers. So rename it to `_has_overlapping_indices(op1, op2)`, and remove references to stabilizers in the docstring.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+    stabilizer_list = list(copy.deepcopy(stabilizers))
+    ham = copy.deepcopy(hamiltonian)
+
+    hamiltonian_subsets = []
+    for stab in stabilizer_list:
+        # Initialize subset
+        ham_subset = QubitOperator()
+        for pauli in ham:
+            if _check_stabilizer_overlap(pauli, stab) is False:
+                ham_subset += pauli
```
kevinsung: The way you wrote this, a Pauli will be added to multiple subsets if it is compatible with them all. This seems surprising; if this is intentional then it should be clearly documented. For applications, do you need the Pauli to appear in every subset that it is compatible with, rather than just one of them?

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+        TypeError: Input stabilizer_list must be QubitOperator or list.
+    """
+    if not isinstance(hamiltonian, QubitOperator):
+        raise TypeError('Input terms must be QubitOperator.')
+    if not isinstance(stabilizers, (QubitOperator, list,
+                                    tuple, numpy.ndarray)):
+        raise TypeError('Input stabilizers must be QubitOperator or list.')
+
+    stabilizer_list = list(copy.deepcopy(stabilizers))
+    ham = copy.deepcopy(hamiltonian)
```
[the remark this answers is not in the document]
xabomon: Hi Kevin,

Thanks for reviewing the PR, sorry that it took a while for me to update it.

Indeed, these objects are never modified, and hence there is no need to deepcopy them.

I'm going to remove the additional variable name as it is also redundant, but I'm happy to put it back.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+its stabilizers."""
+
+import numpy
+import copy
+from openfermion.ops import QubitOperator
+
+
+def _check_stabilizer_overlap(pauli_op, stabilizer):
+    """
+    Auxiliar function of get Hamiltonian subsets.
```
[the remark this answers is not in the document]
xabomon: Corrected! Sorry about this.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,134 @@
...
+    # Find qubits involved
+    qbts_in_stab = [qbt for (qbt, p) in list(stabilizer.terms.keys())[0]]
+
+    overlap = False
+    for qp in list(pauli_op.terms.keys())[0]:
+        if qp == ():
+            overlap = False
+        if qp[0] in qbts_in_stab:
+            overlap = True
+    return overlap
```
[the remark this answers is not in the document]
xabomon: Thanks for this piece of code, very simple and does the job. Changed!

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
xabomon: Corrected! Sorry about this.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
xabomon: Indeed, the functions only aims to check if two Pauli operators have common qubits.

Updated name and docstring.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
xabomon: Thanks for the comment. This is a tricky bit because it is a matter of what the user expects from it.

My original idea was to split the Hamiltonian into subsets of Pauli operators that do not share qubits with a stabilizer (or symmetry) such that if an error is detected by measuring the stabilizer, some information can still be recovered. This allows one to extract information of a circuit run, although an error has occurred, as opposed to a naive error detecting scheme where the run will be disregarded all together.

If you have an opinion about it, please let me know. I will update the documentation accordingly.

Best,
Xavi

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,126 @@
...
+        ham_subset = QubitOperator()
+        for pauli in hamiltonian:
+            if _has_overlapping_indices(pauli, stab) is False:
+                ham_subset += pauli
+
+        # Adds the subset to the list.
+        hamiltonian_subsets.append(ham_subset)
+
+    # Check if there are missing strings
+    rest_paulis = _check_missing_paulis(hamiltonian, hamiltonian_subsets)
```
kevinsung: Instead of executing this function to construct the QubitOperator with the remaining Pauli strings, construct it in the following more efficient way: At the beginning of this function, make a copy of the original Hamiltonian. Then, whenever you add a Pauli string to a subset, delete it from this copy. At the end, this copy will be the thing you want. Also, rename `rest_paulis` to `remaining_paulis`.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,126 @@
...
+    if not isinstance(stabilizers, (QubitOperator, list,
+                                    tuple, numpy.ndarray)):
+        raise TypeError('Input stabilizers must be QubitOperator or list.')
+
+    hamiltonian_subsets = []
+    for stab in list(stabilizers):
+        # Initialize subset
+        ham_subset = QubitOperator()
+        for pauli in hamiltonian:
+            if _has_overlapping_indices(pauli, stab) is False:
```
kevinsung: It's usually clearer to have a positive conditional rather than a negative one. Change the function `_has_overlapping_indices` to `_has_disjoint_indices` (which returns the opposite value) and use that instead. Of course, update the documentation of that function too.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
xabomon: Hi Kevin,
Thanks for taking the time to look into this.

Indeed, it was very dumb to check remaining operators in a different function. Thanks for you help there.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
xabomon: Changed to a positive conditional.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,93 @@
...
+    Auxiliary function of get Hamiltonian subsets.
+
+    Check if two QubitOperators have any qubit in common.
+
+    Args:
+        op1(QubitOperator): Single Pauli string.
+        op2(QubitOperator): Single Pauli string.
+
+    Returns:
+        overlap (Boolean): True if the operator and stabilizer overlap.
```
kevinsung: This needs to be updated. "True if the operators have disjoint indices, False otherwise."

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,93 @@
...
+import numpy
+import copy
+from openfermion.ops import QubitOperator
+
+
+def _has_disjoint_indices(op1, op2):
+    """
+    Auxiliary function of get Hamiltonian subsets.
+
+    Check if two QubitOperators have any qubit in common.
```
kevinsung: This needs to be updated. I think a one-liner "Check if two operators have disjoint indices." is sufficient. Put it on the same line as the triple quotes.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,93 @@
...
+                                    tuple, numpy.ndarray)):
+        raise TypeError('Input stabilizers must be QubitOperator or list.')
+
+    hamiltonian_subsets = []
+    remaining_paulis = copy.deepcopy(hamiltonian)
+    for stab in list(stabilizers):
+        # Initialize subset
+        ham_subset = QubitOperator()
+        for pauli in hamiltonian:
+            if _has_disjoint_indices(pauli, stab) is True:
```
kevinsung: `is True` is unnecessary; remove it.

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,93 @@
...
+                ham_subset += pauli
+                # Try if a Pauli string is still in the remaining set
+                # otherwise continue.
+                try:
+                    remaining_paulis.terms.pop(list(pauli.terms.keys())[0])
+                except:
+                    continue
```
kevinsung: Why is the try-except necessary? Isn't the Pauli here guaranteed to be inside `remaining_paulis'?

src/openfermion/measurements/_hamiltonian_stabilizers_sets.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns:
+        overlap (Boolean): True if the operator and stabilizer overlap.
+    """
+    indices_in_op1 = {index for (index, _) in list(op1.terms.keys())[0]}
+    indices_in_op2 = {index for (index, _) in list(op2.terms.keys())[0]}
+
+    return indices_in_op1.isdisjoint(indices_in_op2)
+
+
+def get_hamiltonian_subsets(hamiltonian, stabilizers):
```
ncrubin, line 40: These needs a more descriptive name

ncrubin (commented):
@xabomon what's the status of this?

xabomon:
Decided to close PR in last OF sync. Functional will be marginally useful.

Issue #457, opened 2018-09-13
Let SymbolicOperator coeff be sympy expressions


joesink:
It would be useful to add sympy symbolic expression compatibility to the FermionicOperator/other operators to allow for cases when the amplitude on the operators is unknown.

babbush:
This is something we've discussed in the past. Indeed, it would be very interesting and would be very useful. It would make OpenFermion a tool for deriving expressions in very general contexts.

It would be great to have more discussion of what the implementation might look like. I think we'd need to implement in such a fashion that the functionality is entirely optional. Otherwise having symbolic objects everywhere would probably kill performance for some important applications.

obriente:
I think I'll need this in my course, as it will allow for better integration with cirq. I'd like to show UCC by saying 'we write the cluster operator in OpenFermion with free parameters and then exponentiate and Trotterize', rather than having to introduce a nice package and then having to do something that feels quite natural by hand. So I'd like to take this on.

I propose beginning by removing the type-checking on SymbolicOperator's coefficient, and then writing some tests to check that algebraic operations and conversion between the different types of SymbolicOperators works as expected (and patching when necessary). This has the advantage that it doesn't reduce performance for anyone who isn't using sympy. It has the disadvantage that we're letting more errors slip, but I think initiating an operator with an inappropriate coefficient is something that will be easily traceable --- please let me know if you think otherwise.

I think that we can't expect to support sympy in every piece of code in OpenFermion --- I would suggest that anything that goes beyond the standard algebraic operations go unsupported (at least for now), and we be explicit about this.

Thoughts?

bryano:
[OFC#353](https://github.com/quantumlib/OpenFermion-Cirq/pull/353) is relevant; it's intended to be a first step towards 'writing the cluster operator with free parameters and then exponentiating and trotterizing'. Once that and #549 are in, I can submit a PR with my `CoupledCluster` class in OFC.

babbush:
@ncrubin was thinking about doing something related recently. I worry that this could be come very unwieldy if we're not careful with it. Tom can you explain the use case a bit more so we get a better picture of what you're hoping to do?

ncrubin:
[...]

PR #549, opened 2020-01-17
Adding capability for SymbolicOperator coefficients to be sympy.Expr

obriente:
Response to issue #457 .

I've been fairly conservative - sympy.Expr don't work with numpy.ndarray, so I've not tried to add any functionality to PolynomialTensor or similar classes (I think anything we do here will impact performance). I've also not inserted any simplification of sympy expressions for similar reasons. In practice, this appears to leave some number of easily-cancelling terms (e.g. following JW or BK), but this can easily be fixed by the user by hand. E.g:

`for key in operator.terms:
`    if isinstance(operator.terms[key], sympy.Expr):
`        operator.terms[key] = sympy.simplify(operator.terms[key])

(Do we want to perhaps implement this as a function in SymbolicOperator - operator.simplify might work, though it is very close to operator._simplify --- or maybe operator.simplify_sympy?)

I've also not implemented saving or loading of operators with sympy - currently saving raises an exception instead. This could perhaps be implemented by stripping whitespace from the coefficient, storing it as a string, and re-inserting it with sympy's sympify function, but this will then require that we accept strings as input when creating a SymbolicOperator. This seems to have been explicitly forbidden before, so I thought I should discuss before blindly pushing (or at least make a second PR).


obriente:
Not sure what's going on here - the build seems to have failed to load something? And some of the uncovered lines appear to be Python 2 specific; is there some way of re-running the CI?

vtomole:
@obriente You should be able to rerun by signing into Travis. https://travis-ci.org/quantumlib/OpenFermion/jobs/638409285?utm_medium=notification. I've filed https://github.com/quantumlib/OpenFermion/issues/550 for the other check.

bryano:
@obriente @kevinsung Any update on this? It would be nice to have in.

obriente:
Trying to remember where this was at - I've just merged master in, lets see what the new tests do.

From memory the PR was mostly ready, however it's difficult to check that sympy behaves well with the functionality we have. Ideally we'd run through all the functions using SymbolicOperators and check that they work with sympy, I guess, and I've not done this here.

obriente:
Alright, seems like everything is working. @kevinsung  @ncrubin , what do you think we need in addition to this?

In particular, I think there's a fairly reasonable line to be drawn between supporting sympy in SymbolicOperators but not in PolynomialTensors. I'm also pretty sure that most of the functions we have that use SymbolicOperators should work reasonably with sympy, but some might not. I'm happy to go through and test everything properly - perhaps @bryano can help, but this will take me a bit of time.

src/openfermion/transforms/_conversion.py
```diff
@@ -40,6 +41,21 @@
                                qubit_operator_sparse)
 
 
+def _check_no_sympy(operator):
```
bryano, line 44: I'm concerned that it's too wasteful to iterate through the terms twice like this. Might be better to catch the errors inline.

src/openfermion/utils/_operator_utils.py
```diff
@@ -605,6 +608,11 @@ def save_operator(operator, file_name=None, data_directory=None,
     else:
         raise TypeError('Operator of invalid type.')
 
+    for term in operator.terms:
```
bryano, line 611: See above.

obriente:
Kevin suggested to check JW and BK and make sure that they work - I'll work on this over the next week.

obriente:
@kevinsung - I've added tests for the JW and BK transforms that make sure everything works, including any linear algebra on coefficients. Think it's enough?

bryano:
I found one bug: the `SymbolicOperator.compress` method doesn't work with symbolic expressions.
For example, the following throws an `AttributeError`:

```openfermion.FermionOperator('', sympy.Symbol('x') + sympy.Symbol('y')).compress()```

**Comment by obriente** [comment]
Good catch, thanks!
I've pushed a fix.

**Comment by obriente** [comment]
Found two other functions that I missed in SymbolicOperator - get_many_body_order() and long_string_init().

The first one was quite easy to change. However, allowing non-understood terms in a string to parse through sympy is something I'm a bit worried about - @bryano , can you think of a way of doing this safely?

File: src/openfermion/ops/_symbolic_operator.py [hunk 3]
```diff
@@ -163,8 +167,12 @@ def _long_string_init(self, long_string, coefficient):
                     else:
                         coef = float(coef_string)
                 except ValueError:
-                    raise ValueError(
+                    try:
+                        coef = sympy.sympify(coef_string)
+                    except:
+                        raise ValueError(
                             'Invalid coefficient {}.'.format(coef_string))
+            print(coef, coefficient)
```
kevinsung, line 175: Leftover debug statement.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -400,8 +410,14 @@ def __iadd__(self, addend):
             for term in addend.terms:
                 self.terms[term] = (self.terms.get(term, 0.0) +
                                     addend.terms[term])
-                if abs(self.terms[term]) < EQ_TOLERANCE:
-                    del self.terms[term]
+                if isinstance(self.terms[term], sympy.Expr):
+                    if (sympy.simplify(
+                            abs(self.terms[term]) < EQ_TOLERANCE)) == True:
```
kevinsung, line 415: Nit: extra parentheses.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -436,8 +452,12 @@ def __isub__(self, subtrahend):
             for term in subtrahend.terms:
                 self.terms[term] = (self.terms.get(term, 0.0) -
                                     subtrahend.terms[term])
-                if abs(self.terms[term]) < EQ_TOLERANCE:
-                    del self.terms[term]
+                if isinstance(self.terms[term], sympy.Expr):
+                    if sympy.simplify(self.terms[term] < EQ_TOLERANCE) == True:
```
kevinsung, line 456: Missing `abs`.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -568,16 +588,32 @@ def __eq__(self, other):
...
         # terms only in one (compare to 0.0 so only abs_tol)
         for term in set(self.terms).symmetric_difference(set(other.terms)):
             if term in self.terms:
-                if not abs(self.terms[term]) <= EQ_TOLERANCE:
-                    return False
-            elif not abs(other.terms[term]) <= EQ_TOLERANCE:
-                return False
+                if isinstance(self.terms[term], sympy.Expr):
+                    if sympy.simplify(
+                            abs(self.terms[term]) <= EQ_TOLERANCE) != True:
```
kevinsung, line 604: The code for checking whether a sympy Expr is small is repeated many times. Factor it out into a helper function.

src/openfermion/ops/_symbolic_operator.py
```diff
@@ -163,8 +167,12 @@ def _long_string_init(self, long_string, coefficient):
                     else:
                         coef = float(coef_string)
                 except ValueError:
-                    raise ValueError(
+                    try:
+                        coef = sympy.sympify(coef_string)
```
kevinsung, line 171: I don't see this functionality tested in any of the tests. I'm actually a bit wary of allowing this, since `sympify` uses `eval` and hence this could execute arbitrary Python code. So unless there is a really good reason to support this use case, I say we don't support it.

src/openfermion/ops/_symbolic_operator.py
obriente: Fixed

src/openfermion/ops/_symbolic_operator.py
obriente: Oh my, thanks for spotting!

src/openfermion/ops/_symbolic_operator.py
obriente: So the main place this would be useful is with saving and loading a SymbolicOperator, as it allows storage via plaintext. Marshal doesn't seem to work with sympy - when we use Marshal we currently cast everything to complex (which definitely doesn't work), but if I lift this requirement then I see a 'ValueError: unmarshallable object' when I try to save a sympy expression, so I think it's not supported.

Looking around, it seems sympy offers support for pickle, which might be an alternative, but apparently pickling has similar code vulnerabilities (so does marshal, it seems), so I'm not sure what to do. I'm happy to just drop support for saving operators containing sympy because there's probably no safe way to load them, or we can maybe throw in a default-off flag for the user, or perhaps something else?

src/openfermion/ops/_symbolic_operator.py
obriente: Fair point, I'll stop being lazy

src/openfermion/ops/_symbolic_operator.py
kevinsung: I'd rather not introduce this vulnerability into the loading of our objects, so I say we drop this feature. A better solution is to implement a more structured serialization format, like the JSON serialization in Cirq. Of course, that is a separate matter.

obriente: SGTM, I'll drop it.

obriente: Dropped

PR #555, opened 2020-02-20
Use shutil.move instead of os.rename

gyu-don:
This modification avoids OSError when current directory and
openfermion/data directory are not on the same filesystem.

Following code is minimal sample to reproduce the problem.
If OpenFermion installed directory and current directory are on the same filesystem,
this code works without any errors.
However, if they're not on the same filesystem, error occurs.

```py
from openfermion import MolecularData
m = MolecularData([['Li', [0, 0, 0]], ['H', [0, 0, 0.7414]]], 'sto-3g', 1, 0, 'test')
m.save()
```

Error:

```
Traceback (most recent call last):
  File "bad.py", line 3, in <module>
    m.save()
  File "/usr/local/lib/python3.6/dist-packages/openfermion/hamiltonians/_molecular_data.py", line 642, in save
    "{}.hdf5".format(self.filename))
OSError: [Errno 18] Invalid cross-device link: '71c31fbb-737f-45ac-959c-11d762f9783c.hdf5' -> '/usr/local/lib/python3.6/dist-packages/openfermion/data/H1-Li1_sto-3g_singlet_test.hdf5'
```

This is described in Python's documentations.

>  The operation may fail on some Unix flavors if src and dst are on different filesystems. 

Cited from: https://docs.python.org/3/library/os.html#os.rename

And, it can avoid by using shutil.move.
> If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied to dst using copy_function and then removed.

Cited from: https://docs.python.org/3/library/shutil.html#shutil.move

### DIFF
--- a/src/openfermion/hamiltonians/_molecular_data.py
+++ b/src/openfermion/hamiltonians/_molecular_data.py
@@ -9,6 +9,7 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+
 """Class and functions to store quantum chemistry data."""
 
 import h5py
@@ -19,6 +20,8 @@ import uuid
 
 from openfermion.config import *
 from openfermion.ops import InteractionOperator, InteractionRDM
+
+
 """NOTE ON PQRS CONVENTION:
   The data structures which hold fermionic operators / integrals /
   coefficients assume a particular convention which depends on how integrals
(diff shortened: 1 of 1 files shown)


kevinsung:
You're getting caught by the newly introduced incremental formatting check. You can either manually fix the one line, or you can run
```
yapf --style=google --in-place _molecular_data.py
```

kevinsung:
Hmm it's still complaining. Are you using yapf version 0.27.0? That's the one that is used by our continuous integration check so it should work.

gyu-don:
I'm using yapf version 0.29.0.
I reverted and correct by hand.

Issue #416, opened 2018-07-28
Set up pylint as part of continuous integration


kevinsung:
This was agreed upon at #319 . We should use the same configuration as Cirq.

fdmalone:
This is done

PR #559, opened 2020-02-26
Remove trailing whitepaces from .py files

vtomole:
Part of https://github.com/quantumlib/OpenFermion/issues/416


vtomole:
@kevinsung I haven't added or removed code. All I did was format. Is this one of those false positives you were talking about?

Issue #420, opened 2018-07-31
Move to pytest-cov


kevinsung:
The incremental coverage scripts used by Cirq and OpenFermion-Cirq are nice because you can use comments to indicate that certain lines don't need to be covered. I think we should move to that system and stop using coveralls. In the past we have had to override the coveralls status check many times.

babbush:
That's fine. I feel like this is a bit redundant with #414. Also, I would deem all of this very low priority.

PR #563, opened 2020-02-27
Remove unused variables

vtomole:


vtomole:
@kevinsung How do we address the coverage check here?

src/openfermion/measurements/_equality_constraint_projection_test.py
```diff
@@ -55,14 +54,14 @@ def setUp(self):
 
     def test_linearize_term(self):
         past_terms = set()
-        for term, coefficient in self.fermion_hamiltonian.terms.items():
+        for term, __ in self.fermion_hamiltonian.terms.items():
```
kevinsung: Can we configure your tool to use just one underscore `_` instead of two?

kevinsung:
Let's just override the coverage check in this case. The longer term solution is #420 .

src/openfermion/measurements/_equality_constraint_projection_test.py
vtomole: Done

vtomole:
@kevinsung Ready for merging.

.travis.yml
```diff
@@ -58,5 +58,5 @@ install:
 script: sphinx-build -nW -b html -d docs/_build/doctrees/ docs/ docs/_build/html
 script: export OMP_NUM_THREADS=1 && pytest src/openfermion --cov src/openfermion
 
-after_success:
-  - coveralls
+# after_success:
+#   - coveralls
```
kevinsung: Oh I didn't mean to actually disable the check. I meant that we could merge this PR in spite of the failing check.

vtomole: I didn't know that was possible. I thought you had to disable, merge then re-enable. I've reverted this.

PR #570, opened 2020-03-06
Add mypy

vtomole:


src/openfermion/measurements/_qubit_partitioning.py
```diff
@@ -21,7 +21,7 @@
     from itertools import izip_longest as zip_longest
 
 
-def binary_partition_iterator(qubit_list, num_iterations=None):
+def binary_partition_iterator(qubit_list: list, num_iterations: int = None):
```
viathor: This should be `List` (capital L) and will require `from typing import List`. However, this is moot if we want to continue to support python 2, since then we'd have to put all type annotations in comments.

kevinsung: We've decided to drop support for Python 2.

ncrubin:
@vtomole What's the status of this w.r.t the move to github actions?

vtomole:
@ncrubin Github actions opened: https://github.com/quantumlib/OpenFermion/pull/585. Need to make build pass and we'll be good to go.

PR #576, opened 2020-04-01
Fix file reference in example notebook

### LINKED ISSUE #571: Correction in example  notebbok file `openfermion_tutorial.ipynb`
OpernFermion example juypter notebook consist of wrong python module name, we need to correct it
From,
ops/_fermion_operators.py.
To,
ops/_fermion_operator.py.

kevinsung:
Fixes #571


kevinsung:
@vtomole The whole paragraph is contained in one line in the file. It is annoying that it doesn't show the one word in the line that was changed.

PR #586, opened 2020-04-17
Support dense density matrices in expectation()

mpharrigan:


src/openfermion/utils/_sparse_tools.py
```diff
@@ -744,32 +744,36 @@ def expectation(operator, state):
...
         if isinstance(operator, scipy.sparse.linalg.LinearOperator):
             raise ValueError('Taking the expectation of a LinearOperator with '
                              'a density matrix is not supported.')
         product = state * operator
-        expectation = numpy.sum(product.diagonal())
+        return numpy.sum(product.diagonal())
 
-    elif isinstance(state, numpy.ndarray):
+    if isinstance(state, numpy.ndarray):
         # Handle state vector.
```
kevinsung, line 756: Should remove this comment now that it includes a density matrix case.

mpharrigan: done

mpharrigan:
ptal

src/openfermion/utils/_sparse_tools.py
```diff
@@ -744,32 +744,34 @@ def expectation(operator, state):
...
+            return numpy.dot(numpy.conjugate(state), operator * state)
+        if len(state.shape) == 2:
+            if state.shape[1] == 1:
+                # Column vector
+                return numpy.dot(numpy.conjugate(state.T),
+                                 operator * state).item()
+            if state.shape[0] == state.shape[1]:
+                # Density matrix
+                product = state * operator
+                return numpy.sum(product.diagonal()).item()
```
kevinsung, line 767: You can just use `numpy.trace`

mpharrigan: aha! you would think that. it does not work on sparse matrices

kevinsung: Oh

kevinsung: But multiplying a sparse matrix with a numpy array gives a numpy array. So here, `product` should be a numpy array, right?

ncrubin: looks like this is correct.  state is a sparse matrix and operator is a sparse matrix.

kevinsung: `state` is a Numpy array (see the if statement in line 755).

ncrubin:
@kevinsung what's the status of this?

mpharrigan:
The status is I forgot about it. This is sortof low on my priority list right now

obriente:
Closing for now (feel free to reopen if you can find the time).

PR #594, opened 2020-05-07
AQA tutorials

obriente:
@karlunho requested that I put the tutorials from the course I ran at Leiden University into openfermion. Currently these have cirq (and some openfermioncirq) dependencies as well, not entirely sure how to rearrange around this, but we figured it was better to just throw them in a PR and then debate. @babbush @kevinsung , thoughts?


obriente:
Looks like everything is failing because of the calls to cirq. Suggestions?

vtomole:
@obriente Why not make this pull request against openfermion-cirq?

babbush:
@vtomole I think this is part of the push we discussed at last OpenFermion meeting to get more of the core documentation into OpenFermion as opposed to the plugins. Not sure the best way to accomplish that though.

obriente:
Yeah, what @babbush said --- I'm personally happy either way.

vtomole:
@obriente @babbush  Okay this is what we can do. Add a new requirements file separate from Openfermions' requirements. Like how we do for `contrib` in Cirq: https://github.com/quantumlib/Cirq/blob/master/cirq/contrib/contrib-requirements.txt. We can have a `tutorials-requirements.txt` that contains the cirq and openfermion-cirq dependencies in the `examples` directory and ignore testing this directory on all of our checks. Then we add a new Github workflow that just runs the example directory.

In Cirq we do

`check/pytest --ignore=cirq/contrib` for the tests where we don't want to run contrib.

obriente:
Seems that there's already a requirements.txt file in /examples - should I run with that? Or should this be separate?

vtomole:
@obriente Seems like we don't even use that file currently so i think it's game. We may want to rename it and remove all the other stuff in it.

obriente:
Alright, I'll rename and remove the other stuff; if people complain than we can change back.

obriente:
This has been added; not sure how to update the workflows though

vtomole:
@obriente Add a step to https://github.com/quantumlib/OpenFermion/blob/master/.github/workflows/ci.yml.

obriente:
@xabomon @StefanoPolla - you might want to review as well. I think the current failure is just a matplotlib issue.

obriente:
Closing this as it's not going to get used.

PR #600, opened 2020-05-28
Erpa

ncrubin:
Implements the ERPA equations.  Generates the non-symmetric eigenvalue problem and uses geev from scipy.lapack to solve the matrix pencil.

Here's the erpa equations with a 2-RDM generated from a Hartree-Fock 1-RDM.  We recover the results from the RPA implementation in pyscf. 

Screen Shot 2020-05-27 at 10 12 53 PM


ncrubin:
@bsenjean may be interested in this.

PR #609, opened 2020-06-29
Added build_api_docs.py script

MichaelBroughton:
Puts `api_docs` build script similar to the ones in Cirq, copied from TFQ. @lamberta I've tested this locally and it should be good to go to be iterated on.

### DIFF
--- a/dev_tools/docs/build_api_docs.py
+++ b/dev_tools/docs/build_api_docs.py
@@ -1,0 +1,63 @@
+# Copyright 2020 The OpenFermion Authors. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


dev_tools/docs/build_api_docs.py
```diff
@@ -0,0 +1,63 @@
...
+from tensorflow_docs.api_generator import doc_controls
+from tensorflow_docs.api_generator import generate_lib
+from tensorflow_docs.api_generator import public_api
+
+import openfermion as of
+
+flags.DEFINE_string("output_dir", "/tmp/of_api", "Where to output the docs")
+
+flags.DEFINE_string("code_url_prefix",
+                    ("https://github.com/quantumlib/openfermion/tree/master/src"
```
lamberta, line 30: not sure it makes a difference, but `docs_to_devsite --preview` is case-sensitive for the repo name ("OpenFermion")

MichaelBroughton: Updated the link to be: https://github.com/quantumlib/OpenFermion/tree/master/src/openfermion . The url uses camel case whereas `src/openfermion` is all lower case.

PR #621, opened 2020-07-23
Givens rotations always insert

obriente:
Adding a flag that allows the user to force all Givens rotations (including trivial rotations) to be inserted. This can be important when processing data from multiple NISQ circuits, as otherwise the changing circuit length can make the signal hard to resolve.


ncrubin:
@obriente FYI we are holding PRs until this the reorg PR is in. This will make updates easier.  hopefully, the reorg PR will be in today.  Sorry for the short delay

obriente:
Oh, good point - lemme close this and wait for the reorg.

PR #622, opened 2020-07-24
Givens rotation always insert

obriente:
Adds option to keep trivial gates in Givens rotation circuit - useful in NISQ for circuit stability.


obriente:
Whoops, just noticed that I'd left a print statement in the reorg pull request. I've removed it here.

PR #625, opened 2020-07-27
QubitOperator to PauliSum translator

xabomon:
This PR contains the functions to translate from QubitOperator to PauliSum.

From the re-organization of the repository it was not clear to me where this functionality should go. I created a new folder `openfermion.circuits.translator` for it, but feel free to suggest a more suitable place.

Previously, this PR was approved in https://github.com/quantumlib/OpenFermion-Cirq/pull/396


xabomon:
ATM Cirq is not found when running pytest, should I do anything from my side about this?

ncrubin:
@xabomon testing setup happens by installing requirements.txt.  cirq isn't in there yet and your PR doesn't add it.  I think adding a cirq dependency to the requirements.txt will solve this issue.

xabomon:
Hi @ncrubin ,

correct me if I'm wrong: adding Cirq to the requirements.txt file implies that Cirq is then needed to install OF.

I recall some discussion in the OF sync if making OF dependent of Cirq was a good idea. Unfortunately I had to miss a couple of these meetings and I'm not sure what's the status of this. Unless someone disagrees I will add Cirq to requirements.

Best,

ncrubin:
@xabomon Adding cirq to requirements is fine.  We are moving towards having OF depend on Cirq.

obriente:
Thanks for putting this PR together Xavi.

I'm not sure the extra folder is needed, as I can't think of anything else that would go in there. Perhaps either openfermion.circuits, or openfermion.transforms.opconversions - @ncrubin , any suggestions?

xabomon:
Hi Tom,

I'm totally open to suggestions on where to place this module.

obriente:
Yeah, this is pretty much my thought process as well. I think opconversions is probably the most appropriate place to put it to be honest, if we say 'everything that depends on cirq should go in circuits', why not ensure that scipy.sparse.linalg is only found in linalg? Shift it there, and I'll do a once-over.

src/openfermion/transforms/opconversions/qubitoperator_to_paulisum.py
```diff
@@ -0,0 +1,61 @@
...
+
+def qubit_operator_to_pauli_sum(operator: QubitOperator,
+                                qubits: Optional[Sequence[cirq.Qid]] = None
+                               ) -> cirq.PauliSum:
+    """
+    Convert QubitOperator to a sum of PauliString.
+
+    Args:
+        operator (QubitOperator): operator to convert.
+        qubits (List): Optional list of qubit names.
```
obriente, line 43: Add a mention of what this defaults to when left empty.

PR #638, opened 2020-08-04
Cleanup of remaining wildcard imports

viathor:
Follow-up to #613. As before, made numpy import explicit as well.


mpharrigan:
Cool. This is more sanitary

PR #641, opened 2020-08-09
Remove wildcard imports

### LINKED ISSUE #630: Make imports explicit in __init__.py
With the current state of the docs (and given that the new filestructure might still need some edits), finding whether or not a function already exists can be difficult (esp. if the function is in an odd location). An easy solution to this is to make imports explicit right down to the root directory, as then one can ctrl-f in the __init__ file.

Also, being explicit prevents having multiple functions with the same name being defined and overwriting each other (... or at least means that it becomes easier to catch).

@ncrubin you had some qualms with this, wanted to raise an issue to start the debate before just dropping a PR.

ncrubin:
Explicit imports throughout the library.

Resolves #630


src/openfermion/__init__.py
```diff
@@ -17,16 +17,142 @@
...
-from openfermion.testing import *
-from openfermion.third_party import *
-from openfermion.transforms import *
-from openfermion.utils import *
+from openfermion.chem import (make_atom, make_atomic_lattice, make_atomic_ring,
+                              angstroms_to_bohr, bohr_to_angstroms,
+                              MolecularData, name_molecule, geometry_from_file,
+                              load_molecular_hamiltonian, periodic_table,
+                              periodic_hash_table, periodic_polarization,
+                              geometry_from_pubchem, make_reduced_hamiltonian)
```
viathor, line 25: If you add a comma after the last item and reformat then the formatter should list one symbol per line which is more visually appealing (and - and this is totally minor - allows you to count exported symbols by looking at line numbers). See this example [`__init__.py`](https://github.com/quantumlib/Cirq/blob/master/cirq/__init__.py) in cirq for example of what this looks like.

(Applies throughout.)

ncrubin: Thanks for the tip on the formatter!

PR #643, opened 2020-08-09
Add option to launch tutorials in Colab

### LINKED ISSUE #628: Colab instead of Binder?
I've been thinking about a unified way of running notebooks easily in Cirq as well as OpenFermion. I saw the binder link - initially I liked it but then I opened it and it takes way too long to bring up the container (5+ minutes). 

What do you think about using Colab? 

https://colab.research.google.com/github/quantumlib/OpenFermion

It is much faster - but you'll need a Google account as soon as you want to run anything. 
Also you'll need to add in all notebooks the `!pip install openfermion` command at the beginning with clear instructions that it's only for colab.

ncrubin:
Updated readme and all notebooks now install openfermion from
first cell if in the colab env.

Resolves #628


ncrubin:
bump @balopat

README.rst
```diff
@@ -20,10 +20,10 @@ Hamiltonians. For more information, see our
     :target: https://badge.fury.io/py/openfermion
 
 
-Run the interactive Jupyter Notebooks on MyBinder:
+Run the interactive Jupyter Notebooks in Colab_ or MyBinder_:
```
kevinsung, line 23: Why the trailing underscores _? Also, the links don't work for me.

balopat:
Awesome!

PR #647, opened 2020-08-12
OpenFermion-Cirq merge: Tutorials

kevinsung:
This PR adds the tutorials from OpenFermion-Cirq, except for the deprecated Tutorial 4 on variational algorithms.


kevinsung:
The naming scheme "tutorial_#" doesn't make sense anymore because there are additional tutorials in OpenFermion that don't follow the numbering scheme. These 3 tutorials are related so they are numbered separately.

PR #652, opened 2020-08-20
Symm cons keywords

philipp-q:
This PR addresses an issue, that appears when using the symmetry_conserving_bravyi_kitaev transformation on operators acting only on a subset of qubits of the whole Hamiltonian (e.g. when one wants to determine the entries of the 2-RDM). This may not be very useful on certain systems (e.g. 2 fermions, 4 orbitals), as the transformation does not yield any benefit, but allows broader application.

Fixed by simple addition of 2 keywords in src/openfermion/transforms/opconversions/remove_symmetry_qubits.py + added a test with a dummy operator.

(A similar issue appears with the BKSF transformation, but would require deeper changes)


philipp-q:
@googlebot I signed it!

googlebot <notifications@github.com> schrieb am Do., 20. Aug. 2020, 11:23:

PR #664, opened 2020-09-18
Add coefficients to SymbolicOperator/PolynomialTensor classes as constant terms

obriente:
This PR adds the functionality to add coefficients to the SymbolicOperator/PolynomialTensor class - in particular by adding them to the constant term.

It also implements __radd__ and __rsub__ for both classes.

This allows to use the sum command to sum a list of SymbolicOperators/PolynomialTensors, which makes the construction of some operators a bit more pythonic. I think this usage of adding a coefficient is also unambiguous, though perhaps people might disagree?


obriente:
Closing PR to fix commit history.

PR #671, opened 2020-10-14
Update tutorial list

ncrubin:
When OpenFermion-Cirq was merged we gained those tutorials.

The list in docs is now updated and reordered.


viathor (approved):
LGTM. I recall there were four of them, but I see only three have been copied over.

ncrubin:
@viathor Thanks for the LGTM.  The fourth tutorial was on the `VariationalStudy` object and infrastructure.  The plan was to deprecate that functionality and thus we won't be moving that notebook.

PR #673, opened 2020-10-21
Added functions to init file

obriente:
Finally fixed my commit history!

I've gone through and either exposed or made explicitly hidden all (I think) functions in the repo. Some of the names were a bit vague (e.g. 'two_body'), LMK if you think something should be exposed that isn't or vice versa.


ncrubin:
@obriente Even resolving the merge conflict it still seems like there is an issue with importing.  Maybe cutting this down to something less than 21 files.  Maybe just bumping this back to when it was passing tests.  Then we can add the extra files in a following PR.

ncrubin:
@googlebot I fixed it.

ncrubin:
@googlebot I signed it.

ncrubin:
@googlebot I consent.

ncrubin:
@obriente bump. I went in and cleaned up some of the imports that were broken from the move of a method.  Let me know if everything else looks in place.

src/openfermion/transforms/opconversions/commutator_diagonal_coulomb_operator.py
```diff
@@ -14,8 +14,9 @@
 import warnings
 
 from openfermion.ops.operators import FermionOperator
-import openfermion.transforms.opconversions.term_reordering\
-    as term_reordering
+from openfermion.transforms.opconversions import term_reordering
+# import openfermion.transforms.opconversions.term_reordering\
+#     as term_reordering
```
obriente, line 19: Not sure why this is commented out, maybe we can just delete it? (No idea if this was from me or you @ncrubin ).

obriente:
Thanks for sorting that out, looks like it must have been a pain to fix.

Everything looks good to me, I guess we just need one extra review now.

src/openfermion/transforms/opconversions/commutator_diagonal_coulomb_operator.py
ncrubin: That was me.  I was sorting out what was causing the issue and I changed this to a from import.  I'll fix it.

ncrubin:
No need for extra review.  thanks for kicking this off and sorting out the extra imports that users were not finding.  This is an impactful PR.

src/openfermion/transforms/opconversions/commutator_diagonal_coulomb_operator.py
obriente: Ah ok. Did that make a difference? I feel a bit lost in this damn mess of imports we have.

obriente:
Ah ok - I can't merge right now without an extra review, maybe you have more privileges than I do here. But everything LGTM

src/openfermion/transforms/opconversions/commutator_diagonal_coulomb_operator.py
ncrubin: The problem wasn't the importing structure.  A file that got moved `double_commuting...two_body` was being incorrectly imported (wrong location).  Thus tests were failing.  The error message was different from when you get circular imports...though sounded similar.  

All that I did to fix was to import from the correct location.

obriente: Hah, very sorry about that then. For some reason tests were passing on my computer... Or at least I thought they were. (I mean, this isn't the kind of error that should pass on any machine)

PR #681, opened 2020-10-27
Anti-symmetrized integrals, coulomb and exchange integrals

Emieeel:
Added functionality to extract anti-symmetrized integrals, coulomb and exchange integrals.


Emieeel:
@googlebot I signed it!

src/openfermion/chem/molecular_data.py
```diff
@@ -173,6 +173,45 @@ def geometry_from_file(file_name):
...
+
+    Returns:
+    antisymints : Numpy array of anti-symmetrized integrals
+        <ij||kl> = <ij|kl> - <ij|lk> (physicist ordering).
+    """
+    symints = numpy.copy(two_body_integrals.transpose(0,1,3,2), order='C')
+    antisymints = symints - two_body_integrals
+    return antisymints
+
+def J_matr(two_body_integrals):
```
ncrubin, line 190: change to lower case for naming consistency.  Do the same for below.

src/openfermion/chem/molecular_data.py
```diff
@@ -173,6 +173,45 @@ def geometry_from_file(file_name):
                 geometry += [(atom, coordinates)]
     return geometry
 
+def antisymint(two_body_integrals):
+    """
+    Args:
+    two_body_integrals : Numpy array of two-electron integrals with OpenFermion
```
ncrubin, line 179: elaborate that this should be spin-orbital integrals.  not spatial.

ncrubin (commented):
Thanks for the code!   A couple of things,

The code should have full tests coverage.  I suggest using the J and K matrices for a simple computational basis state energy calculation.  

We also want to be careful with spin and spatial labels.  In a lot of quantum chem codes the language is pretty loose because the audience is all in-the-know.  But here we will want to lean towards being too descriptive.

src/openfermion/chem/molecular_data.py
Emieeel: They are now in spatial orbital basis, but I can put them in spin-orbital basis if that is better. I suggest then a separate function of the molecular_data class dedicated to going from spatial to spin-orbital basis (that we can also use in the get_molecular_hamiltonian function).

Emieeel:
Ah, so, maybe there is some confusion here.
Chemists often mean with the coulomb/exchange matrix: 
$$
J_{\mu,\nu} = \sum_{\lambda, \sigma} (\mu \nu|\lambda \sigma) D_{\lambda,\sigma}\\
K_{\mu,\nu} =  \sum_{\lambda, \sigma} (\mu \lambda|\nu \sigma) D_{\lambda,\sigma}
$$
in the AO basis, where D_{\lambda,\sigma} is the 1-RDM.
I implemented instead
$$
J_{p,q} = (pp|qq)
K_{p,q} = (pq|qp)
$$
in the MO basis.

Do you want to return the first or the second option? In standard quantum chemistry codes as psi4 and pyscf, usually the first case is returned when calling a function as get_jk().

ncrubin:
I think as a first pass we'll want spin-orbital antisymmeterized 2-electron integrals.  1) we can formulate the Hamiltonian this way (with the 1/4 coeff out front) 2) we can use it to calculate overlaps of the 2-body operator and computational basis states that differ by 2 or fewer occupations.   this also let's us calculate the fock operator (when it is not supplied by a computation) as `einsum('ijjk', antisymm_spin_orb_tei[:, :occ, :occ, :])`.  So I think having spin-orbital antisymm 2 electron integrals will be nice to have.

Emieeel:
The newest commit contains the following changes:
- The anti-symmetrized integrals in the spin-orbital basis (and so documented)
- A separate function to convert spatial orbitals to spin-orbitals, that is also employed in get_molecular_hamiltonian
- Tests to calculate the HF ground state energy with these anti-symmetrized integrals in spin-orbital basis and with coulomb/exchange integrals in the spatial orbital basis

ncrubin:
@Emieeel Thanks for adding the tests.  It looks like formatting and linting are failing.  You can use the `./check/pylint` script and `./check/format-incremental` scripts to help pass this check.  `./check/format-incremental --apply` will reformat your code according to our style guide.

Emieeel (commented):
Hmm... it seems the format test still fails while I applied the ./check/format-incremental

ncrubin:
@googlebot I consent.

ncrubin:
Hi @Emieeel you have to run the ` ./check/format-incremental --apply` for the script to automatically reformat the code.  As for the coverage I checked the lines not covered and it seems that one was a residual if/else in the test and the others were code paths (MissingCalculations) that were not touched.  I added coverage for those lines and resubmitted to your branch.  It looks like everything is passing now.  

Thank you for the commits.  Feel free to open a PR editing the readme with your name and affiliations.  You are now officially an OpenFermion developer!  Welcome!

PR #683, opened 2020-10-29
Verified phase estimation - estimators

obriente:
Splitting a relatively large PR into some staging (this is stage 1)

This PR implements the data processing required for verified phase estimation. It shifts some of the basic lstsq curve-fitting functions from measurements to linalg, and it creates an estimator that analyses VPE data to extract phase function estimates from a set of cirq.TrialResults (circuits to generate those results to come in the next stage), and expectation values from a phase function.

@ncrubin , could you take a look at this one?


ncrubin (approved):
LGTM

PR #686, opened 2020-10-30
Package testing data with OpenFermion

ncrubin:
Data was moved out of the package with the expectation that users
would using this data if they were working on a development branch.

Thus far we have found it useful to have this data available when
developing chemistry programs and techniques.

Therefore the correct place for this data is in testing.

We also explore other options like including this as package_data.

Generally, package_data wasn't originally designed with this particular
usecase in mind.

Overall, though this adds a bit of data to the sdist it makes testing
and using OpenFermion as a development platform easier.


ncrubin:
well that's confusing....I wonder why it can't see that file...

PR #693, opened 2020-12-12
Type hints hamiltonians

subzjee:
I added type hints to the function parameters and for the function return for some files in Hamiltonians.
I also noticed that the parity parameter is expected to take a value of -1 or 1. It doesn't check whether the actual value is either one of those. It will now throw a ValueError for other values.

### DIFF
--- a/src/openfermion/.vscode/settings.json
+++ b/src/openfermion/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "python.linting.pylintEnabled": false,
+    "python.linting.pycodestyleEnabled": true,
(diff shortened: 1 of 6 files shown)


subzjee:
The Format and Lint check won't pass because of the long type hints that define geometry.
Such as:
`geometry: Optional[List[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]]] = None`
I tried splitting them up into two lines but that's not allowed in Python.

ncrubin (approved):
LGTM!  Thanks

PR #705, opened 2021-01-20
Valdemoro 3-RDM reconstruction

ncrubin:
Beware! Cumulant is ZERO!


NnktYoshioka:
Hi Nicholas, I am interested in generating recursively higher order cumulants using lower orders.
Do you have any plans to extend your program to general k-RDM cumulants?

ncrubin:
I don't have plans right now to extend the code to other RDM approximations. This code assumes cumulant is zero.  Higher RDMs can be implemented recursively.  It would be a great addition!

PR #712, opened 2021-02-26
Added IsingOperator to count_qubits.

mstechly:
I've noticed that `count_qubits` in `operator_utils` doesn't handle `IsingOperator` and couldn't find a good reason why not, so I added it.

### DIFF
--- a/src/openfermion/utils/operator_utils.py
+++ b/src/openfermion/utils/operator_utils.py
@@ -21,7 +21,7 @@ from scipy.sparse import spmatrix
 from openfermion.config import DATA_DIRECTORY, EQ_TOLERANCE
 from openfermion.ops.operators import (BosonOperator, FermionOperator,
                                        MajoranaOperator, QuadOperator,
-                                       QubitOperator)
+                                       QubitOperator, IsingOperator)
 from openfermion.ops.representations import (PolynomialTensor,
                                              DiagonalCoulombHamiltonian,
                                              InteractionOperator,
@@ -183,6 +183,15 @@ def count_qubits(operator):
     # Handle PolynomialTensor
     elif isinstance(operator, PolynomialTensor):
         return operator.n_qubits
+
+    # Handle IsingOperator
+    elif isinstance(operator, IsingOperator):
+        num_qubits = 0
+        for term in operator.terms:
+            if term:
+                if term[-1][0] + 1 > num_qubits:
+                    num_qubits = term[-1][0] + 1
+        return num_qubits
 
     # Raise for other classes.
     else:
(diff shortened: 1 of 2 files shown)


src/openfermion/utils/operator_utils.py
```diff
@@ -184,6 +184,15 @@ def count_qubits(operator):
     elif isinstance(operator, PolynomialTensor):
         return operator.n_qubits
 
+    # Handle IsingOperator
+    elif isinstance(operator, IsingOperator):
+        num_qubits = 0
+        for term in operator.terms:
+            if term:
```
ncrubin, line 191: Does this check for the term not being identity?  If so can you make it explicit?

ncrubin:
@mstechly Thank you.  Just a minor question and then we're good to go!

src/openfermion/utils/operator_utils.py
mstechly: It doesn't, but any other case in this function doesn't do that either, so if we would be checking this only for Ising, that would be inconsistent.

I've checked and right now the tests for checking if `n_qubits` returns `5` are passing for all the following cases:

```
self.qubit_operator = QubitOperator("[] + [X1] + [Y2] + [Y3] + [Z4]")
self.qubit_operator = QubitOperator("[X0] + [X1] + [Y2] + [Y3] + [Z4]")
self.ising_operator = IsingOperator("[Z0] + [Z1] + [Z2] + [Z3] + [Z4]")
self.ising_operator = IsingOperator("[] + [Z1] + [Z2] + [Z3] + [Z4]")
```

ncrubin: Huh...that seems broken..the first and last examples should return 4 no?

mstechly: I guess it depends how you want to count qubits.
If you want to act on the 4th qubit, you technically need 5 qubits, right?
We had some issues with how "empty" qubits are handled in various frameworks, @max-radin might have more context.

ncrubin: Certainly in OpenFermion we don't minimize the qubit could.  If you generate a raising operator matrix on the fifth mode you will have a 2**5 matrix.  Cirq is more clever and will only work in the specified space of qubits.

ncrubin (dismissed):

src/openfermion/utils/operator_utils.py
mstechly: Makes sense!
Branch updated, thanks :)

mstechly: I did it in a non-cla compliant manner -.-
Will do it properly later.

mstechly: Ok, should be ok now!

ncrubin (approved):

PR #714, opened 2021-03-06
Update cirq version and citation

ncrubin:
Citation now references published paper.  Cirq version is now pinned to new release.


mpharrigan:
cc @balopat @rmlarose we really need to disentangle the requirements.txt file from install_requires. This causes issues for packages that depend both on cirq and openfermion. Is it actually broken if you use new openfermion with cirq 0.9?

PR #724, opened 2021-04-27
Add FQE docs to devsite

rmlarose:
Self-explanatory from title.

cc @lamberta if you have time for a quick review. The goal is to include the [FQE docs](https://github.com/quantumlib/OpenFermion-FQE/tree/master/docs) on https://quantumai.google/openfermion under a new "FQE" tab. I followed how this is done for ReCirq and Cirq - see https://github.com/quantumlib/OpenFermion-FQE/pull/81 for more details.


lamberta:
To pull docs from the [OpenFermion-FQE](https://github.com/quantumlib/OpenFermion-FQE) repo you'll need to set up separate docs infra for that project. It is much easier to keep the source of those docs in this repo and use the existing nightly jobs.

docs/_book.yaml
```diff
@@ -33,6 +33,10 @@ upper_tabs:
       - title: "Constructing Trotter Steps With Low-Rank Decomposition"
         path: /openfermion/tutorials/circuits_3_arbitrary_basis_trotter
 
+    - name: "FQE"
```
lamberta, line 36: We generally limit the amount of lower tabs to keep navigation consistent throughout the site. Also, "FQE" isn't very clear title for navigation. I think it would be better to move this a separate section in the existing Guide/Tutorials tab.

ncrubin: Could we have it as a separate top tab if it read "OpenFermion-FQE".   Similar to how Cirq has an "Experiments" tab that houses recirq.  My next question is where do the references for this simulator go?  In the "References" tab of OpenFermion?  That doesn't seem right.

rmlarose:
> To pull docs from the [OpenFermion-FQE](https://github.com/quantumlib/OpenFermion-FQE) repo you'll need to set up separate docs infra for that project. It is much easier to keep the source of those docs in this repo and use the existing nightly jobs.

Ahah, thanks for this 🙏

@ncrubin would you prefer to

1. Migrate fqe/docs into this repo as suggested?
2. Set up separate docs infra for FQE?
3. Other?

(1) is easy and I can do this (today). (2) I don't know how to do at the moment.

rmlarose:
Pulled in the FQE notebooks as we discussed, PTAL.

Note for posterity: I did not write 1800+ lines here, just moved the notebooks from https://github.com/quantumlib/openfermion-fqe.

docs/_book.yaml
```diff
@@ -33,6 +33,18 @@ upper_tabs:
       - title: "Constructing Trotter Steps With Low-Rank Decomposition"
         path: /openfermion/tutorials/circuits_3_arbitrary_basis_trotter
 
+      - heading: "FQE Guide & Tutorials"
```
lamberta, line 36: I think the plan was to add sections to existing lower-tabs, yes? Similar to https://www.tensorflow.org/responsible_ai/guide

The RAI `_book.yaml` isn't available in an OSS repo but I can ping you the internal link, if you'd like.

docs/fqe/guide/introduction.ipynb
```diff
@@ -0,0 +1,504 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<table class=\"tfo-notebook-buttons\" align=\"left\">\n",
+    "  <td>\n",
+    "    <a target=\"_blank\" href=\"https://quantumai.google/openfermion/fqe/guide/introduction\"><img src=\"https://quantumai.google/site-assets/images/buttons/quantumai_logo_1x.png\" />View on QuantumAI</a>\n",
+    "  </td>\n",
+    "  <td>\n",
+    "    <a target=\"_blank\" href=\"https://colab.research.google.com/github/quantumlib/OpenFermion-FQE/blob/master/docs/guide/introduction.ipynb\"><img src=\"https://quantumai.google/site-assets/images/buttons/colab_logo_1x.png\" />Run in Google Colab</a>\n",
```
lamberta, line 45: For the notebooks you'll need to update the button paths to point to the new repo location.
You can try `nblint --fix --arg=repo:quantumlib/OpenFermion [notebook.ipynb]` ([doc](https://github.com/tensorflow/docs/tree/master/tools/tensorflow_docs/tools#nblint))

lamberta (changes requested):

docs/_book.yaml
rmlarose: Talked with Nick and the conclusion was to use the current structure (with a different heading name).

docs/fqe/guide/introduction.ipynb
rmlarose: Good catch, thanks!

lamberta (approved):

PR #725, opened 2021-04-30
Added functionality to compute one-norm of qubit/majorana Hamiltonian from molecular integrals

Emieeel:
Added a script in functionals to compute the 1-norm lambda = sum_i |h_i| of a qubit/majorana Hamiltonian H = sum_i h_i P_i directly from the molecular integrals (either input the integrals directly or input a MolecularData class). This is significantly faster than doing an explicit Jordan-wigner/bravyi kitaev transformation, which can become very costly for large systems. See https://arxiv.org/abs/2103.14753 for more information on its usefulness and a derivation for the 1-norm in terms of the integrals.

### DIFF
--- a/src/openfermion/functionals/get_one_norm.py
+++ b/src/openfermion/functionals/get_one_norm.py
@@ -1,0 +1,127 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""
+Function to calculate the 1-Norm of a molecular Hamiltonian after
+fermion-to-qubit transformation from an InteractionOperator
+"""
+
+import numpy as np
+
+from openfermion import MolecularData
+
+def get_one_norm(mol_or_int, return_constant=True):
+    r"""
+    Returns the 1-Norm of the Hamiltonian after a fermion-to-qubit
+    transformation given nuclear constant, one-body (2D np.array)
+    and two-body (4D np.array) integrals.
+
+    Parameters
+    ----------
+
+    mol_or_int : Tuple of (constant, one_body_integrals, two_body_integrals)
+    constant : Nuclear repulsion or adjustment to constant shift in Hamiltonian
+            from integrating out core orbitals
+    one_body_integrals : An array of the one-electron integrals having
+                shape of (n_orb, n_orb).
+    two_body_integrals : An array of the two-electron integrals having
+                shape of (n_orb, n_orb, n_orb, n_orb).
+
+    -----OR----
+
+    mol_or_int : MolecularData class object
+
+    -----------
+
+    return_constant (optional) : If False, do not return the constant term in
+                in the (majorana/qubit) Hamiltonian
+
+    Returns
+    -------
+    one_norm : 1-Norm of the Qubit Hamiltonian
+    """
+    if isinstance(mol_or_int, MolecularData):
+        return _get_one_norm_mol(mol_or_int, return_constant)
+    else:
+        if return_constant:
+            constant, one_body_integrals, two_body_integrals = mol_or_int
+            return _get_one_norm(constant,
+                                 one_body_integrals,
+                                 two_body_integrals)
+        else:
+            _, one_body_integrals, two_body_integrals = mol_or_int
+            return _get_one_norm_woconst(one_body_integrals,
+                                         two_body_integrals)
+
+
+def _get_one_norm_mol(molecule, return_constant):
+    if return_constant:
+        return _get_one_norm(molecule.nuclear_repulsion,
+                             molecule.one_body_integrals,
+                             molecule.two_body_integrals)
+    else:
+        return _get_one_norm_woconst(molecule.one_body_integrals,
+                                     molecule.two_body_integrals)
+
+def _get_one_norm(constant, one_body_integrals, two_body_integrals):
+    n_orb = one_body_integrals.shape[0]
+
+    htilde = constant
+    for p in range(n_orb):
+        htilde += one_body_integrals[p,p]
+        for q in range(n_orb):
+            htilde += ((1/2 * two_body_integrals[p,q,q,p]) -
+                       (1/4 * two_body_integrals[p,q,p,q]))
+
+    htildepq = np.zeros(one_body_integrals.shape)
+    for p in range(n_orb):
+        for q in range(n_orb):
+            htildepq[p,q] = one_body_integrals[p,q]
+            for r in range(n_orb):
+                htildepq[p,q] += ((two_body_integrals[p,r,r,q]) -
+                                  (1/2 * two_body_integrals[p,r,q,r]))
+
+    one_norm = abs(htilde) + np.sum(np.absolute(htildepq))
+
+    for p in range(n_orb):
+        for q in range(n_orb):
+            for r in range(n_orb):
+                for s in range(n_orb):
+                    if p>q and r>s:
+                        one_norm += 1/2  * abs(two_body_integrals[p,q,r,s] -
+                                               two_body_integrals[p,q,s,r])
+                    one_norm += 1/4 * abs(two_body_integrals[p,q,r,s])
+
+    return one_norm
+
+
+def _get_one_norm_woconst(one_body_integrals, two_body_integrals):
+    n_orb = one_body_integrals.shape[0]
+
+    htildepq = np.zeros(one_body_integrals.shape)
+    for p in range(n_orb):
+        for q in range(n_orb):
+            htildepq[p,q] = one_body_integrals[p,q]
+            for r in range(n_orb):
+                htildepq[p,q] += ((two_body_integrals[p,r,r,q]) -
+                                  (1/2 * two_body_integrals[p,r,q,r]))
+
+    one_norm = np.sum(np.absolute(htildepq))
+
+    for p in range(n_orb):
+        for q in range(n_orb):
+            for r in range(n_orb):
+                for s in range(n_orb):
+                    if p>q and r>s:
+                        one_norm += 1/2  * abs(two_body_integrals[p,q,r,s] -
+                                               two_body_integrals[p,q,s,r])
+                    one_norm += 1/4 * abs(two_body_integrals[p,q,r,s])
+    return one_norm
--- a/src/openfermion/functionals/get_one_norm_test.py
+++ b/src/openfermion/functionals/get_one_norm_test.py
@@ -1,0 +1,49 @@
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Tests for get_one_norm."""
+
+import os
+import unittest
+from openfermion import get_one_norm, MolecularData, jordan_wigner
+from openfermion.config import DATA_DIRECTORY
+
+class test_get_one_norm(unittest.TestCase):
+
+    def setUp(self):
+        self.filename = os.path.join(DATA_DIRECTORY,
+                                     "H1-Li1_sto-3g_singlet_1.45.hdf5")
+        self.molecule = MolecularData(filename=self.filename)
+        self.molecular_hamiltonian = self.molecule.get_molecular_hamiltonian()
+        self.qubit_hamiltonian = jordan_wigner(self.molecular_hamiltonian)
+
+    def test_one_norm_from_molecule(self):
+        self.assertAlmostEqual(self.qubit_hamiltonian.induced_norm(),
+                               get_one_norm(self.molecule))
+
+    def test_one_norm_from_ints(self):
+        ints = (self.molecule.nuclear_repulsion,
+                self.molecule.one_body_integrals,
+                self.molecule.two_body_integrals)
+        self.assertAlmostEqual(self.qubit_hamiltonian.induced_norm(),
+                               get_one_norm(ints))
+
+    def test_one_norm_woconst(self):
+        one_norm_woconst = (self.qubit_hamiltonian.induced_norm() -
+                            abs(self.qubit_hamiltonian.constant))
+        ints = (self.molecule.nuclear_repulsion,
+                self.molecule.one_body_integrals,
+                self.molecule.two_body_integrals)
+        self.assertAlmostEqual(one_norm_woconst,
+                               get_one_norm(self.molecule,
+                                            return_constant=False))
+        self.assertAlmostEqual(one_norm_woconst,
+                               get_one_norm(ints,return_constant=False))
--- a/src/openfermion/__init__.py
+++ b/src/openfermion/__init__.py
@@ -34,7 +34,7 @@ from openfermion.chem import (
     make_reduced_hamiltonian,
 )
 
-from openfermion.functionals import contextuality
+from openfermion.functionals import (contextuality, get_one_norm)
 
 from openfermion.hamiltonians import (
     FermiHubbardModel,
--- a/src/openfermion/functionals/__init__.py
+++ b/src/openfermion/functionals/__init__.py
@@ -11,3 +11,5 @@
 #   limitations under the License.
 
 from .contextuality import is_contextual
+
+from .get_one_norm import get_one_norm


src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,127 @@
...
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""
+Function to calculate the 1-Norm of a molecular Hamiltonian after
+fermion-to-qubit transformation from an InteractionOperator
+"""
+
```
obriente, line 16: Maybe put a reference to the paper here?

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,127 @@
...
+import numpy as np
+
+from openfermion import MolecularData
+
+def get_one_norm(mol_or_int, return_constant=True):
+    r"""
+    Returns the 1-Norm of the Hamiltonian after a fermion-to-qubit
```
obriente, line 24: Also put a paper reference here :)

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,127 @@
...
+
+    htilde = constant
+    for p in range(n_orb):
+        htilde += one_body_integrals[p,p]
+        for q in range(n_orb):
+            htilde += ((1/2 * two_body_integrals[p,q,q,p]) -
+                       (1/4 * two_body_integrals[p,q,p,q]))
+
+    htildepq = np.zeros(one_body_integrals.shape)
+    for p in range(n_orb):
```
obriente, line 85: @ncrubin will know this better than I, but for the purposes of speed could this possibly be improved by using np.einsum?

src/openfermion/functionals/get_one_norm_test.py
```diff
@@ -0,0 +1,49 @@
...
+#   See the License for the specific language governing permissions and
+#   limitations under the License.
+"""Tests for get_one_norm."""
+
+import os
+import unittest
+from openfermion import get_one_norm, MolecularData, jordan_wigner
+from openfermion.config import DATA_DIRECTORY
+
+class test_get_one_norm(unittest.TestCase):
```
obriente, line 20: Nit: I think we're moving slightly away from unittest and towards pytest? I have no personal preference here, just wanting to flag.

src/openfermion/functionals/get_one_norm.py
Emieeel: Good one :)

src/openfermion/functionals/get_one_norm_test.py
Emieeel: I'm fine with everything :) I just mimicked some other test scripts

src/openfermion/functionals/get_one_norm.py
Emieeel: Because of the restriction on the sums I didn't know how to employ einsum here. There could be a way to do that, it would certainly make it faster... In the calculations of the paper I actually used numba http://numba.pydata.org/ to speed it up, which helped a lot, but I didn't include it here

src/openfermion/functionals/get_one_norm_test.py
ncrubin: Yes please move over to pytest

obriente:
This looks good to me. @ncrubin , do you have any issues before I merge? I left a question in the code about whether this can be sped up using einsum, but this sounds like it might be a lot of work and Emiel has this working on a hundred spin orbitals so I think it's good enough for now?

ncrubin:
Does antisymmeterization not work to remove the if statements for p>q etc?

src/openfermion/functionals/get_one_norm.py
ncrubin: also it seems like there is no spin-function information.

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,130 @@
...
+                for s in range(n_orb):
+                    if p > q and r > s:
+                        one_norm += 1 / 2 * abs(two_body_integrals[p, q, r, s] -
+                                                two_body_integrals[p, q, s, r])
+                    one_norm += 1 / 4 * abs(two_body_integrals[p, q, r, s])
+
+    return one_norm
+
+
+def _get_one_norm_woconst(one_body_integrals, two_body_integrals):
```
ncrubin, line 109: still needs a doc string

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,130 @@
...
+    n_orb = one_body_integrals.shape[0]
+
+    htildepq = np.zeros(one_body_integrals.shape)
+    for p in range(n_orb):
```
ncrubin, line 113: n_orbs or spin-orbitals?

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,130 @@
...
+        for q in range(n_orb):
+            htildepq[p, q] = one_body_integrals[p, q]
+            for r in range(n_orb):
+                htildepq[p, q] += ((two_body_integrals[p, r, r, q]) -
+                                   (1 / 2 * two_body_integrals[p, r, q, r]))
+
+    one_norm = np.sum(np.absolute(htildepq))
+
+    for p in range(n_orb):
```
ncrubin, line 122: use itertools product or convert to contraction over antisymmeterized integrals.

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,130 @@
...
+fermion-to-qubit transformation from a MolecularData class.
+See https://arxiv.org/abs/2103.14753
+"""
+
+import numpy as np
+
+from openfermion import MolecularData
+
+
+def get_one_norm(mol_or_int, return_constant=True):
```
ncrubin, line 23: Be explicit with inputs.  Use default None and allow the user to specify.  This partially documents the function.

ncrubin: It will also give you freedom to calculate norms for 1-body ops only.  or 2-body only.  Is there a difficulty extending this to arbitrary interactions?

ncrubin (changes requested):

src/openfermion/functionals/get_one_norm.py
Emieeel: This is all in spatial orbital basis

src/openfermion/functionals/get_one_norm.py
Emieeel: Hmmm, this is only valid in the first place for two-body tensors which have real orbital symmetries. I'm not sure how to extend this, as an n-body interaction will not be a molecular system. There is probably a way to write an arbitrary n-body Fermionic Hamiltonian in Majorana representation in terms of the original coefficients though? (I just wouldn't know how)

src/openfermion/functionals/get_one_norm.py
Emieeel: Converted to contraction over antisymmetrized integrals. Thank you for the suggestion!

ncrubin:
I think once @obriente pulls back in @Emieeel to answer 1 or 2 more questions we are good to go.  I would also like the documents to say that this is explicitly for Hamiltonians generated from RHF and ROHF. I don't think this L1 norm is applicable to Hamiltonians built from UHF orbitals

Emieeel:
> I think once @obriente pulls back in @Emieeel to answer 1 or 2 more questions we are good to go. I would also like the documents to say that this is explicitly for Hamiltonians generated from RHF and ROHF. I don't think this L1 norm is applicable to Hamiltonians built from UHF orbitals

Thats true, my understanding was that openfermion only handled restricted molecular Hamiltonians, but maybe I'm wrong?

Emieeel:
I don't understand -- why does the coverage check fail?

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,142 @@
...
+                in the (majorana/qubit) Hamiltonian
+
+    Returns
+    -------
+    one_norm : 1-Norm of the qubit Hamiltonian
+    """
+    if isinstance(mol_or_int, MolecularData):
+        return _get_one_norm_mol(mol_or_int, no_constant=no_constant)
+    else:
+        if no_constant is not None:
```
obriente, line 60: I don't understand this tree - is the idea that no_constant should be either True, False, or None? Also, do we plan the default 'None' to return the constant, or not? I'd recommend setting the default to False or True, and then having a single if statement.

(Also, it seems one of your issues is that you're not covering all three branches of this in your test case).

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,142 @@
...
+                                     two_body_integrals)
+        else:
+            constant, one_body_integrals, two_body_integrals = mol_or_int
+            return _get_one_norm(constant, one_body_integrals,
+                                 two_body_integrals)
+
+
+def _get_one_norm_mol(molecule, no_constant=None):
+    """Compute one_norm for a MolecularData class"""
+    if no_constant is not None:
```
obriente, line 77: Same issue here as above.

obriente:
Seems that there are two of the branches of your handling of no_constant that you don't enter. I've left comments at the relevant places - I think the structure of the code there could be simplified a bit as well.

Thanks for getting onto this!

obriente:
(For the record, if you click on the details of the coverage check, and scroll to the bottom it will tell you the lines in question. There's a bunch of other text, but all the important stuff is at the bottom.)

src/openfermion/functionals/get_one_norm.py
Emieeel: Hmm that is what it was before. I thought this is what Nick meant with "Use default None and allow the user to specify", but I'm probably wrong.

obriente: Heh, fair enough. @ncrubin , can you comment here? I think that a 'None' default is less explanatory as it doesn't tell the user whether the constant will be by default returned or not. Is there any issue with using True/False as a default?

obriente: Hi @Emieeel , just chatted with @ncrubin about this, and I think that the original comment was misunderstood. The point is that instead of taking a tuple as input with all the different data, you should split it up into the different parts, i.e.

def get_one_norm_int(constant, one_body_integrals, two_body_integrals, no_constant=True),

and make a second function 'get_one_norm_mol' which takes a molecular Hamiltonian.

Sorry for the delay in sorting this out!

obriente: Could also do a 'get_one_norm' and 'get_one_norm_mol' as a convenience function.

Emieeel: I'm sorry it took so long, I forgot a bit about it; Is it fine now? I defined two functions: get_one_norm_int and get_one_norm_mol, with no_constant=True.

obriente (approved):
LGTM. I notice that Nick has two outstanding comments regarding docstrings - as the functions are private and the public functions are clearly documented then I think this is ok. Would suggest merging as-is to prevent this dragging on any longer.

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,135 @@
...
+Function to calculate the 1-Norm of a molecular Hamiltonian
+(in the restricted formalism) after fermion-to-qubit transformation
+from a MolecularData class. See https://arxiv.org/abs/2103.14753
+for more information on the 1-norm.
+"""
+
+import numpy as np
+
+
+def get_one_norm_mol(molecule, no_constant=True):
```
ncrubin, line 22: do not overload the inputs like this.  I understand you are dispatching off the boolean input but just have two different functions.  This burdens the user with trying to remember what are valid inputs.

ncrubin (commented):
The get_one_norm_mol needs to still be changed. Overloading the input like that should not be done.  just make a separate function.

src/openfermion/functionals/get_one_norm.py
obriente: I'm not sure that this is being overloaded anymore? @Emieeel has split this into different functions, and the 'molecule' input should be identical for the two.

However, I think the docs need to be updated, as currently you are referring to 'constant', 'one_body_integrals' and 'two_body_integrals' which are no longer taken as input here.

Might also want to reference the two different functions in the docstring - saying something like 'Returns the 1-norm of a Hamiltonian contained in a MolecularData class (to calculate directly from one and two-body integrals refer to the 'get_one_norm_int' function).

Also, if I understand correctly, the '_get_one_norm' and '_get_one_norm_woconst' function literally differ by whether or not the constant is added, right? If so, you could get rid of the _get_one_norm function, and have a workflow like:

```
one_norm = _get_one_norm_woconst(molecule.one_body_integrals, molecule.two_body_integrals)
if add_constant:
    one_norm += abs(molecule.nuclear_repulsion)
return one_norm
```

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,135 @@
...
+        return _get_one_norm_woconst(one_body_integrals, two_body_integrals)
+    else:
+        return _get_one_norm(constant, one_body_integrals, two_body_integrals)
+
+
+def _get_one_norm(constant, one_body_integrals, two_body_integrals):
+    """Compute 1-norm given molecular integrals"""
+    n_orb = one_body_integrals.shape[0]
+
+    htilde = constant
```
obriente, line 70: This should be htilde = abs(constant), right? Otherwise we can wind up with negative 1 norms...

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,135 @@
...
+    if no_constant:
+        return _get_one_norm_woconst(molecule.one_body_integrals,
+                                     molecule.two_body_integrals)
+    else:
+        return _get_one_norm(molecule.nuclear_repulsion,
+                             molecule.one_body_integrals,
+                             molecule.two_body_integrals)
+
+
+def get_one_norm_int(constant,
```
obriente: Setting no_constant == False simply declares that we should add the constant to the result, no? This seems simple enough for the user to do here (as opposed to in get_one_norm_mol where the user has to know how to access the constant in the MolecularData structure). I'd suggest removing this function and exposing '_get_one_norm_woconst' to the user.

src/openfermion/functionals/get_one_norm.py
obriente: Given my above comments, I think this function should be exposed to the user, in which case the docstring from 'get_one_norm_int' can be copied down with minor adjustments.

obriente: Also, in quantum computing 1-norms should almost always be reported without the constant term. I'd suggest calling this function 'get_one_norm', and making it clear in the docs that the constant term is not added. ( @ncrubin , let me know if you have worries about this.)

obriente:
Damn, just realised I was commenting on the wrong branch. Let me change to the new one.

obriente:
@Emieeel , thanks for the patience. I left one comment which I think is a genuine bug, please ignore the others.

@ncrubin , please confirm that this now fits your expectations.

src/openfermion/functionals/get_one_norm.py
Emieeel: No it is actually correct like this. The reason is that you have in principle two (or three if you consider an active space) contributions to the constant term in majorana operators: The nuclear repulsion (always positive), the mean field energy of the frozen orbitals if you consider a CAS (always negative) and the ((1 / 2 * g_[ppqq]) - (1 / 4 *g_[pqpq])) term from rearranging the majorana operators. It is important to actually take the absolute value after counting them up together as they can cancel eachother. I take the absolute value of htilde on the line below, so the 1-norm can never be negative.

obriente: Ah ok, thanks for the clarification!

babbush: no I think this goes back to the comment I made in the "issue" about this. It depends on how you are defining and accessing the LCU. There is a way to represent the LCU such that what @Emieeel is saying is true. And it is a clever trick that he's pointed out there. But under the simplest way to simulate the LCU, you don't get that subtraction from using the Majorana operators.

src/openfermion/functionals/get_one_norm.py
ncrubin: So can you add this as a doc string?

ncrubin:
This will not be merged unless the commit authors actually document the functions.  Spatial integrals is still not referenced.

src/openfermion/functionals/get_one_norm.py
```diff
@@ -0,0 +1,117 @@
...
+    Returns
+    -------
+    one_norm : 1-Norm of the qubit Hamiltonian
+    """
+    return get_one_norm_int(molecule.nuclear_repulsion,
+                            molecule.one_body_integrals,
+                            molecule.two_body_integrals)
+
+
+def get_one_norm_mol_woconst(molecule):
```
ncrubin, line 44: Okay this is the very last thing and then this is done.  It seems that now that we aren't dispatching these functions need to be properly documented.  You  You can just reference `get_one_norm_int` or specify that the parameters molecule: molecularData class.

src/openfermion/functionals/get_one_norm.py
ncrubin: Last thing. See above.  Just document input with "parameters" like the other functions since this is now a public function.

ncrubin (commented):
@Emieeel Thanks for the changes.  There are only two remaining things that are small.  Please document the functions that you have now made public.  This can either reference another function or you can duplicate to docstring.  

Lastly, please add your name and institutions to the contributor list for the package!

PabloAMC:
Hi @Emieeel, awesome and super useful job! By the way, would there be a possibility to make this also return capital Lambda (the maximum value of the amplitudes in the Jordan-Wigner operator) and Gamma (the number of non-zero terms)? I think it is pretty straightforward and it is used in some old phase estimation methods. Doing it myself would be easy but a bit dirty if I intend to publish the code. Thanks again!

src/openfermion/functionals/get_one_norm.py
Emieeel: I think its fixed now. Thank you for being precise Nick, it was a learning experience for me :)

Emieeel:
Hi @PabloAMC, I didn't see you you're comment up to now unfortunately. I think it is definitely possible to return both Lambda and Gamma from spatial integrals with a similar code. The easiest way would just be during the summation to check every term whether its bigger than the current maximum, and keep count how many terms there are (if they are non-zero).

PabloAMC:
Hi @Emieeel, I have made a small Jupiter notebook collab to try to compute lambda using the function you made and also from the corresponding Jordan-Wigner operator, but I think I don't get the same result (roughly double in the former vs the latter). I've also computed Gamma (the number of non-zero terms) and I think I got it wrong too. Capital Lambda (not shown) is right.
The link is [this one](https://colab.research.google.com/drive/1Rp6CP7i-ybAe5LQt3z5LDyuWLs9a11c6?usp=sharing).
Do you know what I may not be getting right for each of these two parameters?
Thanks!

Emieeel:
@PabloAMC I had a look at your notebook, and I have a few corrections. First of all -- and I guess this is why it is important that the function was properly documented -- you're working with spin-orbital integrals, while the input of the 1-norm code is spatial integrals. Then you'll see you get the right 1-norm. Then, computing Lambda is quite easy, as its just the maximum of everything we sum up to the one_norm. 

Computing Gamma is a bit more tedious. If you look at https://doi.org/10.1103/PhysRevResearch.3.033127 Eq. (C13), the Hamiltonian is expressed in terms of Majorana operators. Taking into account that we sum over spin, we have 2*# of 1bdy terms. For the 2bdy terms, the antisymmetric integrals are summed with p>r, s>q and a sum over spin, so we have 1/4 * 2 * # of antisym terms. In trying to compute Gamma here, I found out that there is actually another symmetry for the very last term that doesn't impact the 1-norm at all, but is important for Gamma. That is: switching all together pq \sigma <-> rs \tau (It doesn't matter for the 1-norm as g_{pqrs} = g_{rspq}). This results in a factor of 1/2 for the Gamma, so we have another 1/2 * 2 * # 2bdy terms. Also, don't forget to put a tolerance to set numerical errors to zero (OpenFermion has a standard tolerance of 1e-8 I believe).

I put my corrected code in [this notebook](https://colab.research.google.com/drive/1fR67cZ5lDQVCkA6iHPWPmYmhbJTjwwuG?usp=sharing), where I compute the 1-norm, Lambda and Gamma for O2.

PabloAMC:
Quick question on the above:

> that is: switching all together pq \sigma <-> rs \tau (It doesn't matter for the 1-norm as g_{pqrs} = g_{rspq}). This results in a factor of 1/2 for the Gamma, so we have another 1/2 * 2 * # 2bdy terms.

Why are we multiplying by 2? I thought we had already taken into account the sum over spin

Emieeel:
We still have the sum over sigma != tau in the last term

PabloAMC:
Ah, of course, you're totally right

PabloAMC:
Final question, if I may: lets say I have defined a `molecule_geometry` and a `grid` and run
```
spinless = False

V_dual = dual_basis_potential(grid = grid, spinless = spinless)
U_dual = dual_basis_external_potential(grid = grid, geometry = molecule_geometry, spinless = spinless)
T_primal = plane_wave_kinetic(grid, spinless = spinless)

lambda_value_V = V_dual.induced_norm() - V_dual.constant
lambda_value_U = U_dual.induced_norm() - U_dual.constant
lambda_value_T = T_primal.induced_norm() - T_primal.constant
```
However, if I run
```
JW_op = jordan_wigner(V_dual)
l = abs(np.array(list(JW_op.terms.values())))
lambda_V = np.sum(np.abs(l[1:]))

JW_op = jordan_wigner(U_dual)
l = abs(np.array(list(JW_op.terms.values())))
lambda_U = 2*np.sum(np.abs(l[:]))

JW_op = jordan_wigner(T_primal)
l = abs(np.array(list(JW_op.terms.values())))
lambda_T = 2*np.sum(np.abs(l[1:]))
```
I get different values. In the case of the T term and U terms I get them right, but in V I don't no matter how I do it.
If you want to take a look I've done an example [here](https://colab.research.google.com/drive/1B0th93EfeaR5hXJKsiulJzKIhPWjnYEr?usp=sharing).
Thanks!

Emieeel:
i'm not sure what you're calculating with the grid here, but it seems that the V_dual, U_dual and T_primal have a different induced norm before and after a jordan-wigner transformation. This is not weird as indeed some coefficients can cancel each other. If you calculate JW_op.induced_norm() you get the 1-norm of the QubitOperator. Also, the constant terms are zero here so watch out with l[1:] -- convert it to l[:] if the constant is zero.

PabloAMC:
You're right, we should carry out some calculations similar to the ones you did in the article to match the Jordan-Wigner operator one-norm. However, what is clear is that
```
lambda_value_V = (V_dual.induced_norm() - V_dual.constant)/2
lambda_value_U = (U_dual.induced_norm() - U_dual.constant)/2
lambda_value_T = (T_primal.induced_norm() - T_primal.constant)/2
```
should be an upper bound to the actual JW one-norm, isn't it? I could use
```
JW_op.induced_norm()
```
but the problem is that the bottleneck is in computing `JW_op`

Emieeel:
It is an upper bound, yes. If you express analytically the form of the V_dual, U_dual and T_primal in terms of (unique) Majorana operators, you can find out what the 1-norm is without actually doing the JW transform.

PabloAMC:
I think I know the reason why both `U_dual` and `T_primal` give correct results, because they are one_body. Anyway, I think this is what I was looking for?
```
maj = openfermion.transforms.get_majorana_operator(V_dual)
val = maj.terms.values()
l = np.abs(np.array(list(val)))
one_norm = sum(l[1:])
```
and I get the same as if I do
```
JW_op = jordan_wigner(V_dual)
l_V = abs(np.array(list(JW_op.terms.values())))
lambda_V = np.sum(np.abs(l_V[1:]))
```
but much quicker. So thanks a lot @Emieeel !

PR #735, opened 2021-07-13
Loosen Cirq version pinning

mpharrigan:
Use "compatible release" spec. This is important because
we use requirements.txt to set install_requires


balopat (approved):
LGTM to have flexibility on the patch level

PR #740, opened 2021-08-07
Adding mention of OpenFermion-FQE to the frontpage

babbush:
I noticed that we aren't advertising OpenFermion-FQE here. But we should be right? Feel free to edit aspects of the description.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -87,10 +87,10 @@ Follow the links below to learn more!
 High performance simulators
 ------------------------------------------
 * `OpenFermion-FQE <https://github.com/quantumlib/OpenFermion-FQE>`__ is
-  a high performance emulator of quantum circuit specified by a sequence of
-  fermion operators, which can exploit fermionic symmetries such as spin and
-  particle number. It is built as plugin because C++ backends for
-  OpenFermion-FQE must be compiled.
+  a high performance emulator of fermionic quantum evolutions specified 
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -84,7 +87,15 @@ Specifically, plugins are used to simulate and compile quantum circuits and to p
...
 Follow the links below to learn more!
 
-Circuit compilation and simulation plugins
+High performance simulators
+------------------------------------------
+* `OpenFermion-FQE <https://github.com/quantumlib/OpenFermion-FQE>`__ is
+  a high performance emulator of fermionic quantum evolutions specified 
+  by a sequence of fermion operators, which can exploit fermionic 
+  symmetries such as spin and particle number. It is built as a 
+  plugin because C++ backends for OpenFermion-FQE must be compiled.
```
ncrubin, line 96: I'd leave off the C++ part.  1) fqe will fall back on already faster than cirq python librarires 2) it is built on top of C-extensions.

babbush: so in that case is there a good reason why we don't merge into OpenFermion core?

babbush: in any event, I removed mention of that. I also decided to delete the link to the projects page since I don't feel like we've kept up with it at all and at this point it is not even close to a representative sample of the many projects using OpenFermion. What do you think?

ncrubin (approved):
LGTM

PR #753, opened 2021-10-26
Implement Bravyi-Kitaev for InteractionOperator #752

### LINKED ISSUE #752: Implement Bravyi-Kitaev Transform for InteractionOperator

jamesquantum:
Fixes #752 

Extends the implementation of the Bravyi-Kitaev method as described by Seeley, Richard, Love. 2012 (https://arxiv.org/pdf/1208.5986.pdf) to InteractionOperators. Table II of arXiv:1208.5986 [quant-ph] is implemented by introducing the seeley_richard_love (SRL) method which can also accept complex operators. This method is believed to reduce the overheads in simplification required when evaluating products of creation/annihilation operators as Pauli strings in the Bravyi-Kitaev qubit basis.

image

The previous approach to converting InteractionOperator to QubitOperator would be to:
1. Convert to a FermionOperator (negligible)
2. Use the bravyi_kitaev method
In total this is between 1.2 - 2.2 times slower than using SRL.

image

For large molecules 20% is considerable runtime. When running on an HPC node, transforming trifluoroacetic acid (TFA, 70 qubits) using SRL saved 3.7 minutes.

The SRL approach will not work for general FermionOperator objects. To speedup these, convert to an InteractionOperator before running the transformation.

Here is  simplified explanation of the speedup:
- Consider the simpler case of a Jordan-Wigner transformation. Transforming an arbitrary one-body interaction term (eg. i = N, j = 3) requires taking the product of two fermion operators and writing it as a Pauli string:
image
Between any two operators, many trailing Z gates will lead to cancellations that are trivial when written in this form but must still be computed and evaluated by a general approach.

- By applying each of the 5 cases arising from second-quantisation the Jordan-Wigner transformation can be written as a set of general Pauli string:
image Reference: https://docs.microsoft.com/en-us/quantum/user-guide/libraries/chemistry/concepts/jordan-wigner

- The simplification for Bravyi-Kitaev follows a similar logic, albeit more involved. For each of the 5 equations above, the RHS now evaluates to an expression that is determined by each of the 10 cases illustrated in Table II (p,q -> i, j) above and given here in Python as the SRL method.

This pull request also:
- Applies consistency between  _parity_set and _update_set definitions in arXiv:1208.5986 [quant-ph] (Note: needed to update fenwick_tree_test.py to handle minor change in parity set function imported from bravyi_kitaev.py):
image  image

- Adds additional class to bravyi_kitaev_test.py to maintain full coverage over new code:
Screenshot 2021-10-26 at 17 35 36

### DIFF
--- a/src/openfermion/transforms/opconversions/bravyi_kitaev.py
+++ b/src/openfermion/transforms/opconversions/bravyi_kitaev.py
@@ -132,14 +132,12 @@ def _transform_majorana_operator(majorana_index, n_qubits):
     parity_set = _parity_set(q)
 
     if b:
-        return QubitOperator([(q, 'Y')] + [(i, 'X')
-                                           for i in update_set - {q}] +
+        return QubitOperator([(q, 'Y')] + [(i, 'X') for i in update_set - {q}] +
                              [(i, 'Z')
                               for i in (parity_set ^ occupation_set) - {q}])
     else:
-        return QubitOperator([(i, 'X')
-                              for i in update_set] + [(i, 'Z')
-                                                      for i in parity_set])
+        return QubitOperator([(i, 'X') for i in update_set] +
+                             [(i, 'Z') for i in parity_set])
 
 
 def _transform_operator_term(term, coefficient, n_qubits):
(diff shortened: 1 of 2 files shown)


jjgoings (dismissed):
LGTM!

jjgoings:
@googlebot I fixed it.

jamesphysics:
@googlebot I consent

jamesphysics:
@googlebot I consent.

jjgoings:
@googlebot I consent.

jjgoings (approved):
LGTM

PR #755, opened 2021-11-03
Revamp CI: matrix, multi-Cirq

mpharrigan:
- Use CI "matrix" to reduce duplication in testing infrastructure
- stop pinning cirq in `install_requires`. Lower bound only. More often than not, we are artificially restricting dependers from using the latest Cirq than preventing real problems. Testing against pre-release should prevent surprises.
- Test 1) lowest supported version of things 2) latest version of things 3) vs. cirq pre-release
- Clarify and bump supported python version: latest two versions
- Clarify and bump supported cirq version: latest two versions (unless this is difficult to maintain, then we'd just support the latest version and drop it from the matrix)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,37 +43,32 @@ jobs:
       - name: Lint
         run: check/pylint
   pytest:
-    name: Pytest Ubuntu
-    runs-on: ubuntu-18.04
+    name: Pytest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # On each operating system, check latest version of python and cirq
+        os: [ ubuntu-latest, macos-latest, windows-latest ]
+        python-version: [ '3.9' ]
+        cirq-version: [ '~=0.13.0', '>=0.14.0.dev' ]
+        # Also check least-supported versions (linux only)
+        include:
+          - os: ubuntu-latest
+            python-version: 3.8
+            cirq-version: '~=0.12.0'
+      fail-fast: false
     steps:
-      - uses: actions/checkout@v1
-      - uses: actions/setup-python@v1
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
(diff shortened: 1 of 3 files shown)


mpharrigan:
If there's a true need for a recent Cirq version, that would be made explicit by bumping the min version in `install_requies` and removing that version from the testing matrix. 

In practice, openfermion has rarely been in this situation as of late.

MichaelBroughton:
What about the case of a breaking change between cirq 0.13 and 0.14-dev ?

mpharrigan:
we should fix openfermion to not break with 0.14-dev?

MichaelBroughton (approved):

mpharrigan:
@ncrubin what do you think

ncrubin (dismissed):
LGTM

ncrubin:
I agree with Matt.  if issues come up we can resolve and bump min version.

mpharrigan:
Since this changes the names of expected checks, we need to configure the repository settings or it will keep waiting for the old ci job names

mpharrigan:
Actually I guess TFQ has next-cirq running in a nightly build. This is better, as it won't hold up unrelated PRs (but perhaps is more easily ignored?)

mpharrigan:
@ncrubin @MichaelBroughton my push dismissed your LGTMs. Please re-approve. I'll force-merge and then edit the list of required checks to match the new name. When we bump Cirq or Python it will change the name of the matrix build and we (I) will have to change the list of required checks. If this is too onerous, there's a clunky workaround (where you make one final job that depends on the matrix build job succeeding), but I don't think that will be necessary.

mpharrigan:
bump @MichaelBroughton @ncrubin

ncrubin (approved):
LGTM

mpharrigan:
I force merged this and I've updated the required test for future PRs. Please ping me if there are issues with any future PRs

PR #756, opened 2021-12-03
Direct download the missing file

MarkDaoust:
This file is no longer included in the pip package causing the colab and docs import to fail.

Download it from github.


MarkDaoust:
Wait. This is still broken.

PR #770, opened 2022-02-06
Reduced BCS type Hamiltonians

cvmxn1:
This PR aims to facilitate the use of reduced Bardeen-Cooper-Schrieffer Hamiltonians. Although these Hamiltonians are exactly solvable using the Richardson-Gaudin equations, are useful for serving as a benchmark for strongly correlated systems where the attractive regime is prominent and the equations break down, while maintaining a simple form, allowing for easier implementation in NISQ devices where the connectivity between qubits remains an important factor.

@ncrubin please share your thoughts and feedback on this.

### DIFF
--- a/src/openfermion/ops/representations/doci_hamiltonian_test.py
+++ b/src/openfermion/ops/representations/doci_hamiltonian_test.py
@@ -10,6 +10,7 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 """Tests for doci_hamiltonian.py."""
+import sys
 import os
 import unittest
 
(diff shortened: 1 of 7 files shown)


src/openfermion/ops/representations/doci_hamiltonian_test.py
```diff
@@ -23,15 +24,20 @@
 from openfermion.ops.representations.doci_hamiltonian import (
     DOCIHamiltonian, get_tensors_from_doci, get_projected_integrals_from_doci,
     get_doci_from_integrals)
+from openfermion import (get_fermion_operator, InteractionOperator, \
+                         normal_ordered)
+
+numpy.set_printoptions(linewidth=2000, threshold=sys.maxsize)
```
ncrubin, line 30: this can be remove right? Since the print statements are removed below.

ncrubin (approved):
small nit but LGMT otherwise

src/openfermion/ops/representations/doci_hamiltonian_test.py
cvmxn1: Yes, these print options can be removed since they was used mostly for debugging purposes. However, print statements still exist in the test_fermionic_hamiltonian_from_integrals, and are also useful if the assert statements raise an error in the construction of an object of interest.

ncrubin:
@cvmxn1 can I merge this?

cvmxn1:
Yes, this branch is ready for merge, all docstrings and relevant information have been updated.

PR #785, opened 2022-06-22
Improve runtime type checking in symmetry_conserving_bravyi_kitaev

kwkbtr:
Since the current runtime type checking in `symmetry_conserving_bravyi_kitaev`  uses `type()`, it raises an error when one uses a subclass of `FermionOperator`.
To cover such cases, it is better to use `isinstance()` instead of `type()`.


mpharrigan (approved):
don't know why this was never merged, but LGTM

PR #793, opened 2022-07-25
Making OF compatible with cirq>0.15

ncrubin:
With the removal of final_state_vector from the Circuit object
we move the final_state_vector to the simulator.simulate result.

update requirements and CI

pin to cirq 1.0

unpin from 1.0

update ci workflow

bump min version in CI since 3.6 is no longer supported

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,12 +50,12 @@ jobs:
         # On each operating system, check latest version of python and cirq
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ '3.9' ]
-        cirq-version: [ '~=0.13.0' ]
+        cirq-version: [ '~=0.15.0' ]
         # Also check least-supported versions (linux only)
         include:
           - os: ubuntu-latest
             python-version: 3.8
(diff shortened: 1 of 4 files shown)


mpharrigan:
for anyone reading along: you could also do `cirq.final_state_vector(circuit)`

.github/workflows/ci.yml
```diff
@@ -50,12 +50,12 @@ jobs:
...
         os: [ ubuntu-latest, macos-latest, windows-latest ]
         python-version: [ '3.9' ]
-        cirq-version: [ '~=0.13.0' ]
+        cirq-version: [ '~=0.15.0' ]
         # Also check least-supported versions (linux only)
         include:
           - os: ubuntu-latest
             python-version: 3.8
-            cirq-version: '~=0.12.0'
+            cirq-version: '~=1.0.0'
```
mpharrigan, line 58: the comment in this `include` block says this is supposed to test the least-supported versions (linux only). Here, you've changed it to check newest version of cirq but with oldest version of python

mpharrigan (approved):
sweeet

PR #795, opened 2022-08-08
Fix two openfermion docs

dstrain115:
- These two were using pre-Cirq 1.0 formalisms.
- Change cirq.DropNegligble to cirq.drop_negligible_operations
- Change cirq_google.optimized_for_xmon to
  cirq.optimize_for_target_gateset


ncrubin (approved):
LGTM

PR #805, opened 2022-10-10
Fix convention of PolynomialTensor basis change

aoowweenn:
In general, we rotate a matrix M to M' is to apply a rotation matrix R: M' = R @ M @ R.T
The corresponding Einstein notation is
M'^{p_1p_2} = R^{p_1}_{a_1} R^{p_2}_{b_1} M^{a_1a_2}


mhucka:
@arettig Would you be able to take a look at this and check the technical correctness? (No rush.)

src/openfermion/ops/representations/polynomial_tensor_test.py
```diff
@@ -474,6 +474,40 @@ def test_rotate_basis_reverse(self):
         polynomial_tensor.rotate_basis(rotation_matrix_reverse)
         self.assertEqual(polynomial_tensor, want_polynomial_tensor)
 
+    def test_rotate_basis_90deg(self):
```
arettig: This test actually does not catch the bug you pointed out since for this specific case $$U^T M U$$ gives the same results as $$U M U^T$$. It would be a good idea to change this test to use a different basis rotation.

arettig (changes requested):
This is a good catch. The `PolynomialTensor` rotation code currently does not match the expression in the [documentation](https://quantumai.google/openfermion/tutorials/circuits_1_basis_change). We should change this to match standard convention (and our own documentation), but this is a potentially breaking change for anyone using this function currently.

It would also be a good idea to change the 90 degree rotation test added in this PR to some different rotation as this test actually passes with the current (incorrect) implementation.

src/openfermion/ops/representations/polynomial_tensor_test.py
arettig: Changed to a cyclical rotation test, which can distinguish between the incorrect previous code and the current fixed code.

arettig (approved):
LGTM

PR #806, opened 2022-10-19
Rename `openfermion/third_party/` to `openfermion/contrib/`

tanujkhattar:
The rename is required to make the library compliant with one of the internal checks that disallows having a directory named `third_party`. 

As discussed offline, accessed to objects from `third_party` module in the top-level namespace (`openfermion.object_name`) should remain unchanged and all modules inside the third_party/ directory are private. Also, the code is not expected to be used much and the rename is therefore considered safe to do. 

cc @ncrubin


fdmalone (approved):
LGTM.

PR #814, opened 2023-04-19
Change two_body.dtype comparison from numpy.float to numpy.float64 in DiagonalCoulombOperator

fdmalone:
numpy.float was deprecated in v1.24.0. Replace comparison with numpy.float64 instead. This is causing CI to fail in #813.


fdmalone:
CI now seems to be regressing due to an issue with scipy.linalg that crops up past 1.10

fdmalone:
Closing this in favour of #816 as we need both changes for CI to succeed.

PR #817, opened 2023-04-27
Apply suggestions from CommonTypos checker

pavoljuhas:
No change in code function.


pavoljuhas:
@fdmalone - can you PTAL?

fdmalone:
LGTM. As an aside we should probably be testing the notebooks similar to cirq.

fdmalone (approved):

PR #818, opened 2023-05-23
Use pip-compile for CI dependencies

mpharrigan:
1. factor requirements into their own, specific input `deps/thing.txt` files
2. pip-compile pinned versions to `thing.env.txt`
3. use pinned environments for CI
4. drop `cirq-google` as a dependency since it's not a dependency. 
5. bump ci to python 3.10


fdmalone (approved):
LGTM

PR #822, opened 2023-06-20
[PBC Resources Estimates 2/4] Add Integral Factorization Helpers

fdmalone:
Part 2 of #821. Here we add utility classes to store the factorized integrals for later consumption by pyscf coupled cluster classes and / or resource estimation.


fdmalone:
@mpharrigan ptal when you get a chance

src/openfermion/resource_estimates/pbc/thc/factorizations/thc_jax.py
```diff
@@ -63,23 +63,23 @@ def load_thc_factors(chkfile_name: str) -> KPointTHC:
     xi = None
     with h5py.File(chkfile_name, "r") as fh5:
         chi = fh5["chi"][:]
-        G_mapping = fh5["G_mapping"][:]
-        num_kpts = G_mapping.shape[0]
+        g_mapping = fh5["g_mapping"][:]
```
mpharrigan, line 66: how was this ever working?

fdmalone: I think this was some merge issue where I lowercased everything which was previously uppercase.

mpharrigan: so the lowercase hdf5 string is correct? it would be nice if this sort of thing was covered by a test

mpharrigan (approved):

Issue #825, opened 2023-07-22
Resource estimation code not tested as part of the CI


fdmalone:
Both the molecular and (soon-to-be-merged) pbc resource estimation code is not tested through the CI, largely because there are some external dependencies (pyscf, btas,). I think we should enable these as part of the CI, possibly only run if the resource estimation code is touched. An alternative would be to migrate these to open-fermion pyscf or similar.

Either way, tracking if the upsteam dependencies don't break things is important.

PR #828, opened 2023-08-01
Enable testing for resource estimates

fdmalone:
Previously the resource estimation code was not being tested as part of the CI (#825) . This PR enables it, I've marked a few tests as slow and will open an issue to track the fact that one dependency (btas/pybtas) doesn't have any tests. I think we should add a weekly workflow which will run the slow tests and run a yet to be written pybtas unit test.

I'm only testing the resource estimate code for linux and mac as I don't think pyscf works on windows.


.github/workflows/ci.yml
```diff
@@ -71,6 +71,29 @@ jobs:
       - name: Pytest check
         run: check/pytest
         shell: bash
+  pytest_resource_estimates:
+    name: Pytest Resource Estimates
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        # On each operating system, check latest version of python and cirq
```
mpharrigan, line 79: remove these comments, as I think they were copy pasted from other jobs but aren't accurate here

.github/workflows/ci.yml
```diff
@@ -71,6 +71,29 @@ jobs:
...
+        os: [ubuntu-latest, macos-latest]
+        python-version: [ '3.10' ]
+        cirq-version: [ '~=1.0' ]
+        # Also check least-supported versions (linux only)
```
mpharrigan, line 83: remove this comment

src/openfermion/resource_estimates/__init__.py
```diff
@@ -10,10 +9,10 @@
...
-#coverage: ignore
-import pytest
 
 try:
+    import jax
     import pyscf
-except ImportError:
-    pytest.skip('Need pyscf for resource estimates', allow_module_level=True)
+    HAVE_DEPS_FOR_RESOURCE_ESTIMATES = True
+except ModuleNotFoundError:
```
mpharrigan, line 17: what's the difference between `ImportError` and `ModuleNotFoundError`? I usually use import error for this. Is one a subclass of another?

src/openfermion/resource_estimates/__init__.py
```diff
@@ -10,10 +9,10 @@
...
+    HAVE_DEPS_FOR_RESOURCE_ESTIMATES = False
```
mpharrigan, line 18: should we re-raise the exception with a helpful error message? presumably if these imports fail something else will give a weird error message when you actually go and try to use any of the functionality

check/pytest
```diff
@@ -19,15 +19,21 @@ cd "$(git rev-parse --show-toplevel)"
...
+        RESOURCE_ESTIMATES=1
     else
         PYTEST_ARGS+=("${arg}")
     fi
 done
 
-if [ -z "${ACTUALLY_QUIET}" ]; then
+
+if [ "${RESOURCE_ESTIMATES}" -eq 1 ]; then
+    pytest -c dev_tools/conf/pytest.ini -m "not slow" "${PYTEST_ARGS[@]}" src/openfermion/resource_estimates
```
mpharrigan, line 35: Doesn't pytest just go through all the source files and run anything that looks like a test? If I run pytest *without* the `--resource-estimates` flag ... won't it run the resource estimation tests anyways.

Please provide an alternative rationale if you have one: but I would think that we would always tell pytest to try to test all the test files it can and your mechanism of skipping tests if the dependencies aren't installed will just cause those to be skipped if the dependencies aren't installed. That is -- instead of relying on this flag we just rely on the precense of the dependencies.

If you *do* want a flag: the behavior I would expect is a sort of tri-state logic: If resource estimates are disabled, never run the tests regardless of presence of dependencies. If it's "auto" (the default) run the tests if the dependencies exist, otherwise they will be skipped with your skipping mechanism. If it's set to "on": always try to run the tests and treat the skips as failures.

This probably isn't worth the effort, so I'd just remove the flag and always use the "auto" behavior I've described

src/openfermion/resource_estimates/__init__.py
fdmalone: I think ModuleNotFoundError is designed to catch a narrower set of errors (can't find the module), if the module is there but there's another error raised when importing it it would raise an ImportError. So I guess ImportError will catch a wider range of issues with the two dependencies.

check/pytest
fdmalone: good point. I was having some difficulty passing literal quotes through to check/pytest (-m "not slow")

src/openfermion/resource_estimates/__init__.py
fdmalone: I don't think so, wouldn't this cause an import error (pytest will error rather than skip the tests)

check/pytest
fdmalone: Fixed this "$@" vs $@

fdmalone:
@mpharrigan PTAL

src/openfermion/resource_estimates/__init__.py
mpharrigan: ah good point. Maybe print out a message if you think that could be helpful?

check/pytest
```diff
@@ -19,7 +19,7 @@ cd "$(git rev-parse --show-toplevel)"
 
 PYTEST_ARGS=()
 ACTUALLY_QUIET=""
-for arg in $@; do
+for arg in "$@"; do
```
mpharrigan, line 22: why is this changed? are you sure?

mpharrigan:
the output from the ci job looks like the desired tests aren't being run (??)

```
Run check/pytest -m 'not slow' src/openfermion/resource_estimates
============================= test session starts ==============================
platform linux -- Python 3.10.12, pytest-[7](https://github.com/quantumlib/OpenFermion/actions/runs/5735347442/job/15542949189?pr=828#step:5:8).3.1, pluggy-1.0.0
rootdir: /home/runner/work/OpenFermion/OpenFermion/dev_tools/conf
configfile: pytest.ini
plugins: cov-4.0.0, asyncio-0.21.0, xdist-3.3.1
asyncio: mode=strict
collected 25 items / 7 deselected / 1[8](https://github.com/quantumlib/OpenFermion/actions/runs/5735347442/job/15542949189?pr=828#step:5:9) selected

dev_tools/conf/utils_test.py .....                                       [ 27%]
dev_tools/conf/df/compute_cost_df_test.py ..                             [ 38%]
dev_tools/conf/df/compute_lambda_df_test.py .                            [ 44%]
dev_tools/conf/sf/compute_cost_sf_test.py ..                             [ 55%]
dev_tools/conf/sf/compute_lambda_sf_test.py .                            [ 61%]
dev_tools/conf/sparse/costing_sparse_test.py ..                          [ 72%]
dev_tools/conf/thc/compute_cost_thc_test.py ..                           [ 83%]
dev_tools/conf/thc/compute_lambda_thc_test.py .                          [ 88%]
dev_tools/conf/thc/spacetime_test.py ..                                  [[10](https://github.com/quantumlib/OpenFermion/actions/runs/5735347442/job/15542949189?pr=828#step:5:11)0%]

=============================== warnings summary ===============================
```

mpharrigan:
or is that a weird artifact of the runner? those test files aren't in `dev_tools/conf` and are indeed the correct tests

mpharrigan (dismissed):

check/pytest
fdmalone: $@ splits quoted args on whitespace so the option `-m "not slow"` gets converted to '-m' 'not' 'slow', rather than '-m' 'not slow'. At least I couldn't figure out how to pass in quoted args otherwise. Happy to be schooled on bash though

fdmalone:
very odd, I have no idea why it's giving those paths...

fdmalone:
Maybe to do with pytest saying the root_dir is dev_tools/conf, maybe I used the wrong command to specify the pytest.ini file.

mpharrigan:
maybe that grep expression is doing something weird

fdmalone:
Nah the grep is quieting the output I think. I needed to pass --rootdir through to pytest, somehow specifying the pytest.ini file sets the rootdir to where the ini file is (dev_tools/conf). Should be fixed now.

mpharrigan (approved):
still lgtm. Do reviews get disqualified if you post another commit? We should make that not happen

fdmalone:
Seems like it needed a re-review. Thanks!

PR #829, opened 2023-08-04
Use docker to pin dependencies

mpharrigan:


.github/workflows/ci.yml
```diff
@@ -40,56 +39,59 @@ jobs:
...
     name: Pytest
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # On each operating system, check latest version of python and cirq
         os: [ ubuntu-latest, macos-latest, windows-latest ]
-        python-version: [ '3.10' ]
-        cirq-version: [ '~=1.0' ]
-        # Also check least-supported versions (linux only)
-        include:
```
fdmalone, line 57: What's the minimum python version we support now?

dev_tools/requirements/re-pip-compile-in-docker.sh
```diff
@@ -0,0 +1,35 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+set -ex
+
+# https://stackoverflow.com/q/59895
+SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &>/dev/null && pwd)"
+readonly SCRIPT_DIR
+
+# We use docker to run the actual `pip-compile` command because its
```
fdmalone, line 23: What is the workflow for regenerating new requirement files now? Should we run this script on our local machine and add?

fdmalone (changes requested):
More of a discussion question, but it looks like we're moving to only testing python3.10, which seems fine, should we add 3.9 / 3.11?

fdmalone:
But yeah, this is generally a much simpler way of managing environments.

.github/workflows/ci.yml
mpharrigan: Instead of abusing the strategy matrix, I moved the "least supported" test to a new job `pytest-minimum`. There, I bumped the minimum python version for shady reasons:

 - The "minimum" environment uses cirq 0.15, which supports python 3.8 so we could run the job on python 3.8
 - however, the way the job works is first you install the full, modern, pinned environment which requires python 3.9 because cirq 1.2 requires python 3.9
 - After the full environment, we install the older version of cirq. But that step would fail with the old python

The "correct" fix would be to pin a minimal environment. The correct way to do this would be to have another Dockerfile that uses python 3.8

dev_tools/requirements/re-pip-compile-in-docker.sh
mpharrigan: If you run `re-pip-compile-in-docker.sh` on a workstation with docker installed, it should just work :tm:

You can also run the python script outside of docker but its output may vary depending on the environment you run it from

mpharrigan:
ok, for better or for worse I have changed my pip-compile driver script to be able to support multiple "platforms" where each platform is a collection of environment (pinned requirement) files. Within a platform they're all consistent. 

So here: the default `envs/` directory has all the normal stuff with the latest cirq whereas `max_compat/` envs all have cirq constrained to be 1.0.0

furthermore: the python version of the docker image is now parameterized so we generate the `envs/` default envs on python 3.10 and the `max_compat/` envs on python 3.8

.github/workflows/ci.yml
mpharrigan: I have unbumped the minimum python version by doing a correct fix

fdmalone:
@mpharrigan is this good to go?

mpharrigan:
Sorry, I'll have to revisit this and give it the once-over to make sure it's good to go

fdmalone:
I need to regenerate as in #843 for black. I'll close #843 in favour of this.

fdmalone (approved):

fdmalone:
Assuming the macos runners are down for the moment.

PR #835, opened 2023-10-02
Switch to using the black formatter

mpharrigan:
- copy `check/format-incremental` from Cirq (one minor edit to remove an unnecessary exclude line)
- copy `black` version pin from Cirq so our open-source dev environments are mildly in sync.

Instead of formatting everything all at once, I plan on just letting the formatting change slowly over time as PRs format only their changed files. cc @ncrubin @fdmalone @tanujkhattar

### DIFF
--- a/check/format-incremental
+++ b/check/format-incremental
@@ -1,22 +1,23 @@
 #!/usr/bin/env bash
 
 ################################################################################
-# Formats lines of python code that have been modified.
+# Formats python files that have been modified.
 #
 # Usage:
-#     check/format-incremental [BASE_REVISION] [--apply]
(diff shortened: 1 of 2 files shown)


fdmalone:
Why not format everything in one go?

fdmalone:
cirq did this all in one go right, with some git-blame-ignore-revs?

tanujkhattar:
> cirq did this all in one go right, with some git-blame-ignore-revs?


Yes

fdmalone:
With that being said, I'd rather switch to black in one way or another. Can follow up with a bulk reformat.

fdmalone:
xref #835

fdmalone (approved):

PR #843, opened 2023-10-19
Update requirements for black

fdmalone:
Following #835 we need to regenerate the requirements files. I'm not really sure why resolver==backtracing is lost in the diff.


fdmalone:
This may depend on #829 which fixes the minimum python version / cirq issue.

fdmalone:
closing in favour of #829

PR #849, opened 2023-10-24
Fix setup resource path

fdmalone:
I was a bit too quick to bump the version it seems. The path was wrong for the resource estimates requirements. Also the manifest file was outdated.


fdmalone:
@mpharrigan PTAL

mpharrigan (approved):

Issue #839, opened 2023-10-17
Inconsistencies in the double factorized chemistry resource estimate costing function


fdmalone:
the costing function computes br for outer prepare and inner prepare differently:

https://github.com/quantumlib/OpenFermion/blob/a034397a13b1ea224e11746d4ee368cccf2bc962/src/openfermion/resource_estimates/df/compute_cost_df.py#L53

Which would be fine but the br fed into cost two should be using the second (hardcoded value) of 7 to output into qrom for state preparation on the second register.

mhucka:
(I'm going around and checking issue statuses.) @fdmalone could you let me know what the status of this is? (Should we unassignn you?)

PR #850, opened 2023-10-27
Move version back to dev.

fdmalone:


src/openfermion/_version.py
```diff
@@ -11,4 +11,4 @@
 #   limitations under the License.
 
 """Define version number here and read it from setup.py automatically"""
-__version__ = "1.6.0"
+__version__ = "1.6.0.dev"
```
mpharrigan, line 14: this should be 1.7.0.dev0 probably

fdmalone: I guess, how about 1.6.1.dev0 as I imagine releasing some bug fixes for #839 etc (what is the 0 after the dev for, it seems cirq doesn't have the zero, qualtran does)

mpharrigan: My reading of the spec: https://peps.python.org/pep-0440/#public-version-identifiers says that it should have a zero. https://peps.python.org/pep-0440/#developmental-releases but everything will work without it https://peps.python.org/pep-0440/#implicit-development-release-number

If we were to do things the right way :tm: patch releases would be made off of a branch. E.g. `master` would move to 1.7.0.dev0; fixes are committed there; and critical bugfixes are cherry-picked to a `1.6` branch that branches off of the tagged release

mpharrigan (approved):

PR #852, opened 2023-10-30
Fix pytest deprecation warnings.

fdmalone:
The diff is huge because of reformatting. May be worth first merging #846.  But just fixed some deprecation warnings (numpy.product, numpy.math, converting size one array to scalar)  emitted by pytest.

### DIFF
--- a/src/openfermion/linalg/davidson_test.py
+++ b/src/openfermion/linalg/davidson_test.py
@@ -20,7 +20,6 @@ import scipy.linalg
 import scipy.sparse
 import scipy.sparse.linalg
 
-from openfermion.ops.operators import QubitOperator
 from openfermion.linalg.davidson import (
     Davidson,
     DavidsonOptions,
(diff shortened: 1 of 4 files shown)


src/openfermion/linalg/davidson_test.py
```diff
@@ -116,7 +117,7 @@ def test_invalid_dimension(self):
 
 
 class DavidsonTest(unittest.TestCase):
-    """"Tests for Davidson class with a real matrix."""
+    """ "Tests for Davidson class with a real matrix."""
```
mpharrigan, line 120: nit there's an extra quote here

fdmalone: Fixed.

fdmalone:
@mpharrigan needs approval

mpharrigan (approved):
thanks for the ping; I was waiting until I could see an actual manageable diff :)

Issue #853, opened 2023-10-31
Nightly tests are broken


fdmalone:
Should fix this and use it as an opportunity to test the slow tests.

mhucka:
This appears to have been fixed in the past. Closing.

PR #869, opened 2024-02-21
Fix doc generation

dstrain115:
This document errors out with the following error:

"RuntimeError: maximum taylor expansion limit reached"

Changing the time to 0.9 eliminates the error.  Better solutions or ideas welcome, but this at least allows the python notebooks to be generated.

### DIFF
--- a/docs/fqe/tutorials/fermi_hubbard.ipynb
+++ b/docs/fqe/tutorials/fermi_hubbard.ipynb
@@ -196,7 +196,7 @@
    "outputs": [],
    "source": [
     "%%time\n",
-    "e_time = 1.0\n",
+    "e_time = 0.9\n",
     "true_evolved_fqe = init_wfn.time_evolve(e_time, hubbard)"
    ]
   },


fdmalone:
Do we know when this started to fail?

dstrain115:
> Do we know when this started to fail?

Not really.  At least a month.   It's been broken for a while, and we were not checking the email alias for doc generation for a while.

fdmalone:
My current guess is this threshold was never met for this example (it's set at 1e-15 (the example reaches 1e-14)). In fqe v0.3.0 this raises an error if the max number of iterations is reached. Looking back to v0.1.0 there was no warning raised. The notebook seems to be quite a bit older given that it's being pip installed as a git repo and not as a tagged release (there is a comment about updating this to pip install fqe once v0.1.0 is released).

It seems like the correct fix would be to update the accuracy threshold but this doesn't appear to be a controllable parameter from time_evolve. So for the moment I think your fix is the only way. 

WDYT @ncrubin

fdmalone:
although this seems pretty environment dependent, are there build logs somewhere / environment information. As a side note we should be testing the Openfermion notebooks ourselves periodically.

fdmalone:
xref #853

dstrain115:
Ah, interesting.  Maybe we can also create a new issue to plumb out the error_threshold as a parameter as well.  Let me know if you would like me to put in this fix for now.

fdmalone:
Yeah I will open a tracking issue, it's probably best to merge now to fix the docs at least. And we can make the unitary prettier again once it's done.

fdmalone:
Will you need a new openfermion release to fully fix this?

PR #898, opened 2024-12-19
Fix comparison in freeze_orbitals

aleksey-uvarov:
Comparing using "is" instead of "==" would sometimes lead to incorrect results


ncrubin (approved):
LGTM

Seems like this is for an integer comparison where == is appropriate.

PR #904, opened 2025-01-29
Block Invariant Symmetry Shift functionality 

SmikPatel:
Added a function to calculate the 1-norm of DF post-processed with block invariant symmetry shifts (DF+LRPS method of https://pubs.acs.org/doi/abs/10.1021/acs.jctc.4c01390), along with a test.


mhucka:
@SmikPatel Thank you for this work, and my apologies for the long delay in replying. This looks reasonable and we'd like to work on getting it into OpenFermion. When you have a moment, could you please update your branch on GitHub? Before we add code review comments, it would be best to work on an updated version of the code.

<div align="center">
image
</div>

PR #921, opened 2025-02-02
Add .zenodo.json file

mhucka:
Zenodo is an open repository maintained by CERN for preserving research data and software. It assigns permanent unique identifiers to each repository so it can be identified and cited. Zenodo also provides an integration mechanism allowing GitHub repository owners to arrange for GitHub releases to be sent automatically to Zenodo when they are made.

In order to provide users with (i) a way to override the values obtained via the GitHub API and (ii) to set other metadata fields available in Zenodo but not in GitHub, the Zenodo-GitHub integration mechanism will also look for a `.zenodo.json` file at the top level of a repository.

This PR adds a `.zenodo.json` file for OpenFermion; separately, we will turn on the GitHub-Zenodo integration for this repository, so that OpenFermion releases are archived in Zenodo automatically.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,248 @@
+{
+  "title": "OpenFermion",
+  "description": "The electronic structure package for quantum computers.",
+  "license": "Apache-2.0",
+  "access_right": "open",
+  "upload_type": "software",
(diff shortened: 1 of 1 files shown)


ncrubin (approved):
LGTM

PR #966, opened 2025-02-10
Add CITATION.cff file

mhucka:
This basically mirrors the `.zenodo.json` file contents.


mpharrigan (approved):
why do we have to put all the authors twice?

mhucka:
> why do we have to put all the authors twice?

That's a good question. The author lists are slightly different: there's one for this repository (based on the list in the README file), and then there's one for the "preferred citation", which is the 2020 paper.

I hate that there's this almost-duplicate-but-not-quite thing going on. As an alternative, we could – if you guys prefer – change the repository author list to a single entry like "The OpenFermion Developers". We'd want to keep the paper author list as-is (that one will never change anyway, it's a said-and-done kind of thing).

The tradeoff of using "The OpenFermion Developers" is easier maintenance over time, at the cost of reducing individual contributor's acknowledgements and visibility.

PR #996, opened 2025-03-04
Improve format-incremental script

mhucka:
This is a rewritten version of `check/format-incremental`. This was motivated by the desire to add a `--help` option and do a little bit of cleaning up the code, and then one thing led to another, and, well, here we are. The final changes include:

- Added a `--help` option
- Added a `--no-color` option
- Added a `--quiet` option
- Make the script more robust in various ways (e.g., declare variables, be more careful in parsing CLI arguments, etc.)
- Don't hard-code ANSI color code sequences in every echo statement
- Make the script more DRY
- Revise the error & info messages to try to be slightly more clear


mpharrigan:
my personal rule is once you start using a for loop in a bash script, it's probably time to port it to something like python. Is there someone with more bash expertise who can review this?

mhucka:
> Is this shared with the other various repos?

Not yet, but that's my intention. I had to add one for the unitary repo, took the opportunity to rewrite it, and now am going around to add it to the other repos.

> Is this the first place the updated script will be committed?

It's in a PR for here and unitary at the momment.

> my personal rule is once you start using a for loop in a bash script, it's probably time to port it to something like python. Is there someone with more bash expertise who can review this?

I think what I hear you saying is that it's become too complicated. OK, I worked on simplifying it just now and will push an update.

Pavol is at least as expert in Bash as me (likely more), so he's the only other one I can think of. I'll add him to the reviewers.

mhucka:
@mpharrigan The latest commit has a simplified version of the code.

mpharrigan:
Thanks @mhucka ! It would be nice if @pavoljuhas could give a thorough review so we can port this to all the repos with confidence

PR #999, opened 2025-03-04
Add workflow to welcome first-time contributors

mhucka:
This adds a workflow that triggers on new issues and PRs. If the author has not contributed to this repository before, the workflow will add a comment with an informative welcome message to the issue or PR.


mhucka:
Let's just bag this. Nice to have, but not essential, and people don't seem to be enthused about it.

Issue #1024, opened 2025-03-27
Effect by Numpy>=2.0, codes with numpy.string_ raise AttributeError


azure1ce:
Hi, I'm runing VQE by using Nvidia's cudaq package, it's 
------------------cudaq.chemistry.create_molecular_hamiltonian()  --------------
function stack with 
--------------line 711 in  file of "openfermion/chem/molecular_data.py"------------
shows:
----------AttributeError: `numpy.string_` was removed in the NumPy 2.0 release. Use `numpy.bytes_` instead.----------
I found that there are 7 usage in the file, not sure how many usage in the whole package
So that warning to may need maintain the code for the future use.

Thank you for building the package

Kee-Wang:
Just come here to say I encounter the `same numpy.string_` problem, although triggered by some other functions.

azure1ce:
> Just come here to say I encounter the same `numpy.string_` problem, although triggered by some other functions.

I tried simply replaced numpy.string_ with numpy.bytes_ and allowed to running forward. You may could try it as well.
Bests

Kee-Wang:
> > Just come here to say I encounter the same `numpy.string_` problem, although triggered by some other functions.
> 
> I tried simply replaced numpy.string_ with numpy.bytes_ and allowed to running forward. You may could try it as well. Bests

Hi, I solved my part of the problem by simply creating a new conda env.

PR #1055, opened 2025-06-02
Partly fix #1008 and #1024: update dependencies and environment files for NumPy 2

### LINKED ISSUE #1008: NumPy v2 support
It seems that OpenFermion depends on `numpy>=1.24,<2.0`, but since `numpy<2.0` does not support Python 3.13, I would like to use `numpy>=2.0`. Do you have any plans to support NumPy v2?

mhucka:
This updates the minimum version of NumPy to 2, and cirq-core to 1.5 (the latter is needed to support NumPy 2).

### DIFF
--- a/dev_tools/requirements/envs/dev.env.txt
+++ b/dev_tools/requirements/envs/dev.env.txt
@@ -6,7 +6,7 @@
 #
 ase==3.24.0
     # via
-    #   -r /Users/mhucka/project-files/google/github/openfermion/dev_tools/requirements/deps/resource_estimates_runtime.txt
+    #   -r deps/resource_estimates_runtime.txt
     #   -r deps/resource_estimates_runtime.txt
 astroid==3.3.8
     # via pylint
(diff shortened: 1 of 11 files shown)


dev_tools/requirements/envs/dev.env.txt
```diff
@@ -58,10 +60,13 @@ iniconfig==2.0.0
 isort==6.0.0
     # via pylint
 jax==0.4.38
-    # via -r deps/resource_estimates_runtime.txt
+    # via
+    #   -r deps/resource_estimates_runtime.txt
+    #   -r deps/resource_estimates_runtime.txt
```
dstrain115: Why are all these comments repeated?

dstrain115 (approved):

dev_tools/requirements/envs/dev.env.txt
mhucka: > Why are all these comments repeated?

Not sure, since this was auto generated. I can delete them. Another commit coming up in a moment.

PR #1056, opened 2025-06-02
Set a fixed random seed during testing

mhucka:
Some of the tests became more flaky with the move to NumPy 2. In all cases, random numbers were involved at some point. Setting a fixed random seed made the tests reproducible and (so far) seems to have stopped the flakiness.

### DIFF
--- a/src/openfermion/conftest.py
+++ b/src/openfermion/conftest.py
@@ -11,9 +11,9 @@
 #   limitations under the License.
 
 import os
+import random
 import pytest
 import numpy as np
-import random
 
 
 def pytest_configure(config):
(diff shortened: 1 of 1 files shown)


dstrain115:
I suppose this is fine, but I am a tad worried we are sweeping some problems under the rug.

mhucka:
> I suppose this is fine, but I am a tad worried we are sweeping some problems under the rug.

A fair point. What if we do this?

- open an issue to flag this as a problem
- use the global seed for now, so we can release OpenFermion with NumPy 2 support
- fix the problem properly after the release

dstrain115:
SGTM.

mhucka:
> SGTM.

Issue is #1060.

PR #1057, opened 2025-06-02
Partly fix #1008 and #1024: adjust files to support NumPy 2

### LINKED ISSUE #1008: NumPy v2 support

mhucka:
Update some files to use constructs compatible with NumPy 2.

A separate PR (#1055) updates the requirements files & dependencies for NumPy 2.

### DIFF
--- a/src/openfermion/chem/molecular_data.py
+++ b/src/openfermion/chem/molecular_data.py
@@ -324,13 +324,12 @@ def geometry_from_file(file_name):
 def antisymtei(two_body_integrals):
     """
     Args:
-
-    two_body_integrals : Numpy array of two-electron integrals with OpenFermion
-        Ordering.
+        two_body_integrals: Numpy array of two-electron integrals with
+            OpenFermion Ordering.
 
     Returns:
-    antisymints : Numpy array of anti-symmetrized integrals
-        <ij||kl> = <ij|kl> - <ij|lk> (physicist ordering).
+        antisymints : Numpy array of anti-symmetrized integrals
+            <ij||kl> = <ij|kl> - <ij|lk> (physicist ordering).
     """
     symints = numpy.copy(two_body_integrals.transpose(0, 1, 3, 2), order='C')
     antisymints = symints - two_body_integrals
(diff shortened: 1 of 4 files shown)


src/openfermion/chem/molecular_data.py
```diff
@@ -324,6 +324,7 @@ def geometry_from_file(file_name):
 def antisymtei(two_body_integrals):
     """
     Args:
+
```
dstrain115: Is this correct formatting?  Usually its Args: followed by an indented argument, no line break.  (I think)

src/openfermion/chem/molecular_data.py
```diff
@@ -403,7 +406,8 @@ def spinorb_from_spatial(one_body_integrals, two_body_integrals):
 
 
 class MolecularData(object):
-    """Class for storing molecule data from a fixed basis set at a fixed
+    """Class for storing molecule data from a fixed basis set at a fixed.
```
dstrain115: This doesn't look like it's a correct period, since it is continued "fixed ....  period"

src/openfermion/ops/operators/binary_code.py
```diff
@@ -293,7 +295,12 @@ def __rmul__(self, factor):
 
     def __str__(self):
         """Return an easy-to-read string representation."""
-        string_return = [list(map(list, self.encoder.toarray()))]
+        # string_return = [list(map(list, self.encoder.toarray()))]
```
dstrain115: Don't keep commented code around.

src/openfermion/chem/molecular_data.py
mhucka: You're right, it's not. The formatter did that. Will fix.

src/openfermion/chem/molecular_data.py
mhucka: You're right. That was an auto-formatting result. The tool must have gotten confused by the blank line after the first line. Will fix.

src/openfermion/ops/operators/binary_code.py
mhucka: Whoops, thanks for catching that.

JerryChen97:
Finally we are going to have `OpenFermion` compatible with NumPy >= 2.0 :partying_face:

PR #1066, opened 2025-06-03
Fix #1064: replace calls to deprecated method `orbital_energies`

### LINKED ISSUE #1064: OpenFermion methods still call deprecated method `orbital_energies()`
### Describe the issue

The method `orbital_energies()` on the option `QuadraticHamiltonian` in `src/openfermion/ops/representations/quadratic_hamiltonians.py` is deprecated (and produces warnings when called), but a few files still contain calls to it. These should be replaced with calls to `diagonalizing_bogoliubov_transform()`.

1.7.1.dev0

Run `check/pytest`. It will produce a number of warnings like this:

```
src/openfermion/ops/representations/quadratic_hamiltonian_test.py::DiagonalizingCircuitTest.test_particle_conserving:317
-> src/openfermion/ops/representations/quadratic_hamiltonian.py:444 DeprecationWarning('The method `orbital_energies` is deprecated. Use the method `diagonalizing_bogoliubov_transform` instead.')
```

mhucka:
Method `orbital_energies` in is deprecated in file `quadratic_determinants.py`; its replacement is method `diagonalizing_bogoliubov_transform`.

### DIFF
--- a/src/openfermion/circuits/slater_determinants.py
+++ b/src/openfermion/circuits/slater_determinants.py
@@ -103,11 +103,9 @@ def gaussian_state_preparation_circuit(
     if not isinstance(quadratic_hamiltonian, QuadraticHamiltonian):
         raise ValueError('Input must be an instance of QuadraticHamiltonian.')
 
-    (
-        orbital_energies,
-        transformation_matrix,
-        _,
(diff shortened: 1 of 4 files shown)


dstrain115 (approved):
Seems fine to me.  Would love @ncrubin to verify this is a valid replacement.

PR #1074, opened 2025-06-04
Fix a number of CodeQL code-scanning alerts

mhucka:
This addresses a number of alerts by CodeQL. Many more remain, but at least it's a start.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,21 +16,23 @@ from setuptools import find_packages, setup
 
 # This reads the __version__ variable from openfermion/_version.py
 __version__ = ''
-exec(open('src/openfermion/_version.py').read())
+with open('src/openfermion/_version.py') as v:
+    exec(v.read())
 
-# Readme file as long_description:
+# The readme file is used as the long_description:
 long_description = '===========\n' + 'OpenFermion\n' + '===========\n\n'
 with open('README.rst', 'r', encoding='utf-8') as readme:
     long_description += readme.read()
 
-# Read in package requirements.txt
-requirements = open('dev_tools/requirements/deps/runtime.txt').readlines()
+# Read in package requirements.txt.
+with open('dev_tools/requirements/deps/runtime.txt') as r:
+    requirements = r.readlines()
 requirements = [r.strip() for r in requirements]
 requirements = [r for r in requirements if not r.startswith('#')]
-# Resource estimates requirements.
-resource_requirements = open(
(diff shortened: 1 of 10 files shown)


setup.py
```diff
@@ -16,21 +16,23 @@
 
 # This reads the __version__ variable from openfermion/_version.py
 __version__ = ''
-exec(open('src/openfermion/_version.py').read())
```
fdmalone, line 19: I wonder if we should follow cirq and use runpy over exec: https://github.com/quantumlib/Cirq/blob/main/setup.py#L23

setup.py
```diff
@@ -16,21 +16,23 @@
...
-# Readme file as long_description:
+# The readme file is used as the long_description:
 long_description = '===========\n' + 'OpenFermion\n' + '===========\n\n'
 with open('README.rst', 'r', encoding='utf-8') as readme:
     long_description += readme.read()
 
-# Read in package requirements.txt
-requirements = open('dev_tools/requirements/deps/runtime.txt').readlines()
+# Read in package requirements.txt.
+with open('dev_tools/requirements/deps/runtime.txt') as r:
```
fdmalone, line 28: I just ran into this recently when mypy complained about with open syntax, apparently pathlib.Path(file).read_text() is preferred?

src/openfermion/chem/molecular_data.py
```diff
@@ -841,6 +841,7 @@ def save(self):
         try:
             os.remove("{}.hdf5".format(self.filename))
         except OSError:
+            # Nothing to do but carry on.
```
fdmalone, line 844: Did codeql complain here?

fdmalone (approved):

src/openfermion/chem/molecular_data.py
mhucka: > Did codeql complain here?

Yeah, it did: https://github.com/quantumlib/OpenFermion/security/code-scanning/459

setup.py
mhucka: That's a good idea. I'll update this.

setup.py
mhucka: Hmm. Mypy doesn't complain about this for some reason. I'm inclined to leave it as-is for now, just to get on with this release. Would that be okay with you @fdmalone?

fdmalone: Maybe it was a linter, anyway sounds fine.

Issue #929, opened 2025-02-03
Function linearize_term may return None by default


mhucka:
Function `linearize_term` in file `src/openfermion/measurements/equality_cosntraint_projection.py` has several returns in an if-elseif conditional, but it doesn't have a default return, which means it can return `None` if none of the conditions match. This is probably wrong, but I don't know what the default return value should be instead.

mhucka:
Looks good. Thanks! Closing this issue as completed by PR #929.

PR #1086, opened 2025-06-28
Raise error for non-physical terms

SamKChang:
Raise ValueError for issue [929](https://github.com/quantumlib/OpenFermion/issues/929) when terms are unexpected because only one- and two-body operators are present in molecular Hamiltonian (See Eq. (4) - (7) in https://www.nature.com/articles/ncomms5213)


mhucka (approved):
LGTM

mhucka:
@SamKChang Thank you for your work!

PR #1096, opened 2025-07-24
Fix typos in two files

dstrain115:
- This fixes two typos of "representation" caught by the CommonTypos tool


mhucka (approved):
Thank you for these!

PR #1098, opened 2025-07-27
Update Dockerfile and its README

Bevacizumab:
The original Dockerfile has issues and is not working. Ubuntu and Python versions are not specified, and the link to Psi4 is expired. This pull request fixes the Dockerfile and updates its README.md to match.

The new configuration has been tested on an Apple Silicon Mac (M4). Customization for other platforms is straightforward and can be understood easily with the updated README.md.

Changes to Dockerfile:
1. Pin versions for Ubuntu (22.04) and Python (3.12).
2. Add `WORKDIR` to mount local files.
3. Set `PATH` with `ENV`.
4. Psi4 is installed with a conda command instead of `wget`.
5. Dependencies are installed inside of a virtual environment (`fermion`).
6. Reduce the number of `RUN` commands to reduce the number of layers.

Changes to its README.md:
1. Put some details on the Dockerfile in the first paragraph of Setting up ~.
2. Add details to customization.
3. Change `docker run` command and introduce a simpler way to mount local files. So some previous explanations are deleted.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -15,7 +15,8 @@ OpenFermion (or any of its plugins) using the standard procedure.
 ## What's included?
 
 - Git
-- Python 3
+- Python 3.12
+- [Miniconda](https://www.anaconda.com/docs/getting-started/miniconda/main)
 - [OpenFermion](https://github.com/quantumlib/OpenFermion)
 - [Cirq](https://github.com/quantumlib/Cirq)
 - [Psi4](http://www.psicode.org)
@@ -27,6 +28,25 @@ OpenFermion (or any of its plugins) using the standard procedure.
 
 ## Setting up Docker for the first time
 
+The Dockerfile is based on the [Ubuntu image](https://hub.docker.com/_/ubuntu) (ver. 22.04).
+It creates a Python (ver. 3.12) virtual environemnt (named `fermion`) using Miniconda and installs all dependencies within it. Psi4 is installed with a conda [command](https://psicode.org/installs/v191/).
+The default configuration uses the Miniconda installer (ver. 25.5.1-1) for Python 3.12 on Linux `aarch64` architecture.
+
+### Customizing the Environment
+You can manually edit the Dockerfile if you need to set up a different development environment (e.g., change the Ubuntu, Python, Miniconda, or Psi4 version).
+
+If your local machine builds Linux `x86_64` architecture with the Dockerfile, the `wget` command 
+for the Miniconda installer (Line 40 in the Dockerfile)
+```
+wget https://repo.anaconda.com/miniconda/Miniconda3-py312_25.5.1-1-Linux-aarch64.sh -O ~/miniconda3/miniconda.sh && \
+```
+must be changed to 
+```
+wget https://repo.anaconda.com/miniconda/Miniconda3-py312_25.5.1-1-Linux-x86_64.sh -O ~/miniconda3/miniconda.sh && \
+```
+You can check other Miniconda installers [here](https://repo.anaconda.com/miniconda/).
+
+### Building Docker Image
 You first need to install [Docker](https://www.docker.com/).
 Once Docker is setup, one can navigate to the folder containing the
 Dockerfile for building the OpenFermion image (docker/dockerfile) and run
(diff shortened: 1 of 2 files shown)


mhucka:
> I completed the CLA last week, right after I saw the bot's comment. I've just confirmed that the CLA status is green in the "Checks" tab. It seems you might have been viewing an older notification without refreshing it.

Oh, great. Thanks for doing that.
 
> Please let me know if you have any feedback on the code changes themselves. Thank you!

Yes, I will review it.

Bevacizumab:
Hi @mhucka, I've just rebased my branch on top of the latest master. Sorry for the long wait.

docker/README.md
```diff
@@ -27,9 +28,28 @@ OpenFermion (or any of its plugins) using the standard procedure.
 
 ## Setting up Docker for the first time
 
+The Dockerfile is based on the [Ubuntu image](https://hub.docker.com/_/ubuntu) (ver. 22.04).
+It creates a Python (ver. 3.12) virtual environemnt (named `fermion`) using Miniforge and installs all dependencies within it. Psi4 is installed with a conda [command](https://psicode.org/installs/v191/).
```
mhucka:
```suggestion
It creates a Python (ver. 3.12) virtual environment (named `fermion`) using Miniforge and installs all dependencies within it. Psi4 is installed with a conda [command](https://psicode.org/installs/v191/).
```

docker/README.md
```diff
@@ -27,9 +28,28 @@ OpenFermion (or any of its plugins) using the standard procedure.
...
+The default configuration uses the latest Miniforge installer on Linux `aarch64` architecture.
```
mhucka: Why does this change the default from the previous x86_64 architecture? From our experiences, x86_84-based systems are still more common than arm64 systems.

docker/README.md
```diff
@@ -39,32 +59,21 @@ where "openfermion_docker" is just an arbitrary name for our docker image.
...
-random and generated automatically). Using this name, one can then copy
-files into the active Docker session from other terminal using:
-
+where "openfermion_container" is an arbitrary choice for the name of our docker container. This command will mount your current local directory to `/root/workspace` inside the running container.
+By default, the virtual environment `fermion` is automatically activated in the running container.
+After installing the `Dev Containers` extension in Visual Studio Code, 
+you can open the container with the following option:
 ```
-docker cp [path to file on disk] [container name]:[path in container]
+Command Palette -> Dev Containers: Attach to Running Container..
```
mhucka: This change refers to a GUI interface, correct? As far as I know, `docker` does not provide a GUI. There is a separate tool, Docker Desktop, that provides a GUI, but we can't assume that users will be using that. (Also, Docker Desktop has license restrictions.)

Please revert this back to CLI-based instructions.

docker/dockerfile
```diff
@@ -12,48 +12,46 @@
 
 # Dockerfile for OpenFermion, Cirq, and select plugins.
 
-FROM ubuntu
+FROM ubuntu:22.04
 
 USER root
```
mhucka: This was in the original Dockerfile, and not a new addition, but it may as well be fixed as part of this PR. The Dockerfile should not need root, so it should be possible to omit this line entirely. (I'm not sure why the original had it in the first place.) Removing the line will solve the check failure about "USER should not be root".

docker/dockerfile
```diff
@@ -12,48 +12,46 @@
...
-ENTRYPOINT bash
+WORKDIR /root/workspace
+COPY . /root/workspace
+
+# Set PATH for miniforge
+ENV PATH="/root/conda/bin:${PATH}"
+# Set PATH for pyscf
+ENV PYTHONPATH=/root/pyscf
+
+RUN apt-get update && \
```
mhucka, line 23: For DL3009, I will add a configuration setting to `.hadolint.yaml` to disable this particular check because it's just not worth the hassle.

For DL3015, go ahead and add the recommended `--no-install-recommends` flag.

For DL3008, we should probably pin the versions, although there are pros and cons to doing that. One way to find out the versions that can be used is to do roughly the following:
1. switch to the x86_64 architecture as discuss in a previous comment above
2. run the docker build as-is
3. figure out what versions of the packages were installed by the build
4. edit the file to add those vesions to the apt install command.

I don't know if it's possible with `apt-get install`, but try to see if it will let you specify just major.minor versions, not full major.minor.patch.whatever; i.e., "12.3" instead of "12.3.0.build34" etc.

mhucka (changes requested):

Bevacizumab:
@mhucka Thank you for your good and detailed review comments. It helped me a lot for improving the Dockerfile. I will explain the changes in the commit.

Dockerfile
- `USER root` is deleted.
- I put `--no-install-recommends` flag for `apt-get install`.
- I thank you for adding the configuration setting.
- I fixed the package versions for `apt-get install`. Unfortunately, it requires the full strings. I've manually checked all the installed packages' versions for both arm64 and amd64 architectures. Both architectures install the same version for each package.
- I've made changes to support both linux/arm64 and linux/amd64 architectures (Psi4 only supports these two linux architectures). It is possible by putting `uname` and `uname -m` for the wget link of Miniforge.
- Additionally, I solved DL3003 warning. Each `cd` was changed to `WORKDIR`.

README
- I added information about the architecture support to help anyone who is not familiar with the concept.
- I corrected the typo `environemnt`.
- I deleted the explanation of "the default architecture".
- I reverted to CLI-based instructions. Mounting local files is a good docker practice. But some might not want to do this. So I provide instructions on how to `COPY` or not, and how to mount or not.
I split the explanation of copying local files and git repositories into a new subsection for clearance.

PR #1100, opened 2025-07-29
revise the created pauli_term_grouping.py

hofong428:
Optimized Pauli term grouping strategies for measurement reduction in fermionic simulations.
Based on research showing up to 50% measurement reduction compared to traditional techniques.


mhucka:
@hofong428 thank you for this work. It looks like Google lacks evidence of a Contributor License Agreement (CLA) having been accepted – could you check the instructions and do that? (It won't ask you to give up any rights.)

Also, could you provide a somewhat more informative title for this PR? It looks like the revision has to do with optimizing performance, so perhaps it should mention that. This will make it easier for people to understand what's going on in the future.

hofong428:
I've tried to conduct the PRs as codebatai(my startup company), or is only
personal PR allowable?

PR #1101, opened 2025-07-29
Create pauli_term_grouping_test.py

hofong428:


mhucka:
@hofong428 thank you for this work. There are some failures in the CI checks, as you can see. Also, the CodeQL scanner flagged some unused Python imports that can be removed from the files.

mhucka:
> So.. what should I do now?

If this PR contains test cases for the contribution in PR #1100, then please add the file from this PR to PR #1100, close this PR, and continue at #1100.

The problems flagged by the CI here will still need to be addressed. It is up to your preferences whether you want to address them first and then move the file to the other PR, or move the file first and then address the problems. (When you push another commit to PR #1100 to add the file, it will trigger the CI jobs to run again, and the same errors will occur there until they are addressed.)

I hope I answered the question you were asking. If I misunderstood, please let me know.

hofong428:
Crystal clear. Thanks for your help. I will re-submit later.

PR #1109, opened 2025-08-25
Remove CodeQL scanner & move others to new weekly schedule

mhucka:
GitHub enhanced their CodeQL scanner and its configuration options since the time I originally wrote the custom workflow file here. It's now easier to simply use GitHub's repository options for setting up CodeQL. No need to maintain our own workflow anymore.

Also, the other scanners (OSV and Scorecard) are run on every PR, so there is not much value in running also running them nightly. I moved them to a new weekly workflow because it's still worth running them regularly.

Finally, this updates the workflows to the latest versions used in our template repo.


.github/workflows/osv-scanner.yaml
```diff
@@ -69,9 +85,12 @@ jobs:
       security-events: write
       # Needed to upload SARIF file to CodeQL.
       contents: read
+    env:
+      # Setting Bash SHELLOPTS here takes effect for all shell commands below.
+      SHELLOPTS: ${{inputs.debug && 'xtrace' || '' }}
     steps:
       - name: Check out a copy of the git repository
-        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
pavoljuhas: Is this intentional?  If so, please comment in the PR description why is it necessary to downgrade.  Same for other action downgrades below.

mhucka: Ah, no. This is the consequence of having copies of the workflow :-(. I'll update this and push a new commit momentarily.

Thanks for spotting that.

pavoljuhas (changes requested):
Please check the downgrades of actions in this PR.

Also, quite a few workflows run with `write-all` permissions.
Can we change that to something more restrictive?
For example, the OSV scanner example at https://google.github.io/osv-scanner/github-action/#scan-on-pull-request, grants `write` permission for `security-events` only.

.github/workflows/osv-scanner.yaml
mhucka: Updated the versions & hashes everywhere that needed it.

mhucka:
> Also, quite a few workflows run with `write-all` permissions. Can we change that to something more restrictive? For example, the OSV scanner example at [google.github.io/osv-scanner/github-action#scan-on-pull-request](https://google.github.io/osv-scanner/github-action/#scan-on-pull-request), grants `write` permission for `security-events` only.

A fair point. I committed a change that restricts the permissions. It _seems_ to work, although I remember having a lot of trouble finding working permissions and settled on write-all partly out of desperation. We will have to wait and see if this works once merged into the default branch.

.github/workflows/scorecard-scanner.yaml
```diff
@@ -72,6 +86,14 @@ jobs:
           publish_results: true
 
       - name: Upload results to code-scanning dashboard
-        uses: github/codeql-action/upload-sarif@96f518a34f7a870018057716cc4d7a5c014bd61c # v3
+        uses: github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3
```
pavoljuhas:
```suggestion
        uses: github/codeql-action/upload-sarif@51f77329afa6477de8c49fc9c7046c15b9a4e79d # v3.29.5
```

.github/workflows/osv-scanner.yaml
```diff
@@ -99,18 +119,19 @@ jobs:
...
+            --include-git-root
             --format=json
             --output=new-results.json
             --recursive
             ./
 
       - name: Run the OSV scanner reporter
         # yamllint disable rule:line-length
-        uses: google/osv-scanner-action/osv-reporter-action@456ceb78310755116e0a3738121351006286b797 # v2.2.1
+        uses: google/osv-scanner-action/osv-reporter-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.0.0
```
pavoljuhas:
```suggestion
        uses: google/osv-scanner-action/osv-reporter-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.1.0
```

.github/workflows/osv-scanner.yaml
```diff
@@ -82,10 +101,11 @@ jobs:
 
       - name: Run OSV scanner on existing code
         # yamllint disable rule:line-length
-        uses: google/osv-scanner-action/osv-scanner-action@456ceb78310755116e0a3738121351006286b797 # v2.2.1
+        uses: google/osv-scanner-action/osv-scanner-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.0.0
```
pavoljuhas:
```suggestion
        uses: google/osv-scanner-action/osv-scanner-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.1.0
```

.github/workflows/osv-scanner.yaml
```diff
@@ -99,18 +119,19 @@ jobs:
 
       - name: Run OSV scanner on new code
         # yamllint disable rule:line-length
-        uses: google/osv-scanner-action/osv-scanner-action@456ceb78310755116e0a3738121351006286b797 # v2.2.1
+        uses: google/osv-scanner-action/osv-scanner-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.0.0
```
pavoljuhas:
```suggestion
        uses: google/osv-scanner-action/osv-scanner-action@b00f71e051ddddc6e46a193c31c8c0bf283bf9e6 # v2.1.0
```

pavoljuhas (approved):
LGTM with a couple of comment updates.

Thank you for taking care of permissions!

.github/workflows/osv-scanner.yaml
mhucka: @pavoljuhas Do you have an automated way of checking the versions for a PR review like this, or do you run something like frizbee by hand?

mhucka: (When I ran frizbee, it did not change the comments to add the more specific versions. So I'm wondering how you found it.)

PR #1111, opened 2025-08-25
Fix #925: inconsistent return in df_integrals.py::double_factorize

### LINKED ISSUE #925: Inconsistent returns in double_factorize
In file src/openfermion/resource_estimates/pbc/df/df_integrals.py, [lines 159-172](https://github.com/quantumlib/OpenFermion/blob/36ced6d5f55b9bad5c09fb3f3352e181a37cdaf0/src/openfermion/resource_estimates/pbc/df/df_integrals.py#L159-L172), the function `double_factorize` is declared as returning None, but there's a return statement where it returns something other than none:

```python
        if self.df_factors is not None:
            return self.df_factors
```

mhucka:
The `double_factorize` method in the `DFABKpointIntegrals` class had an inconsistent return. It was type-hinted to return `None`, but had a code path that returned a value. This early return also caused a caching issue where the method would incorrectly return a cached result even when called with a new `thresh` value, preventing re-computation.

This commit removes the premature return statement. This allows the `thresh` parameter to take effect as intended and makes the method's behavior consistent with its type hint.


dstrain115 (approved):

fdmalone:
I think the actual issue here is that df_factors doesn't seem to be used anywhere...

mhucka:
> I think the actual issue here is that df_factors doesn't seem to be used anywhere...

I did some searching in the code base, and it looks like you're right. In that case, I propose to change this PR to remove this property from the class – unless you think that we've stumbled on a deeper problem (e.g., that something _should_ be setting the value, but isn't)?

mpharrigan:
@fdmalone does that sounds good to you? @mhucka do you want to make the change?

fdmalone:
yep

PR #1112, opened 2025-08-25
Fix #885: type coercion due to slightly incorrect default value

### LINKED ISSUE #885: Bad behavior when working with sympy
I found that when using sympy expression as coefficients in `FermionOperator`, the expression would be transformed automatically into float numbers. For example:
```
from openfermion.ops import FermionOperator
from sympy import Rational

a = FermionOperator('0^ 0', Rational(1, 2))
a
```

This gives the output:
```
1/2 [0^ 0]
```

But when doing:
```
a = FermionOperator('0^ 0', Rational(1, 2))
b = FermionOperator('1^ 1', Rational(1, 2))
a + b
```

Then it outputs:
```
1/2 [0^ 0] +
0.500000000000000 [1^ 1]
```

Which is not what I would expect. This is probably due to that, in the `SymbolicOperator` class' addition method (line 435-436):
```
            for term in addend.terms:
                self.terms[term] = self.terms.get(term, 0.0) + addend.terms[term]
```
When `term` isn't find in self, its coefficient is set to `0.0`, which might cause some type conversion for sympy expression. Since it originates in `SymbolicOperator` , it would probably happen in other operator classes.

mhucka:
When adding a new term to a `SymbolicOperator` with a sympy coefficient, the coefficient was being cast to a float. This was because the default value for a new term was `0.0`, which is a float. This commit changes the default value to `0`, which is an integer. This ensures that the type of the coefficient is preserved when adding new terms.

### DIFF
--- a/src/openfermion/ops/operators/symbolic_operator_test.py
+++ b/src/openfermion/ops/operators/symbolic_operator_test.py
@@ -9,13 +9,15 @@
 #   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
+
 """Tests  symbolic_operator.py."""
+
 import copy
+import numpy
+import sympy
 import unittest
 import warnings
 
-import numpy
-import sympy
 from openfermion.config import EQ_TOLERANCE
 from openfermion.testing.testing_utils import EqualsTester
 
--- a/src/openfermion/ops/operators/symbolic_operator_test.py
+++ b/src/openfermion/ops/operators/symbolic_operator_test.py
@@ -20,6 +20,7 @@ from openfermion.config import EQ_TOLERANCE
 from openfermion.testing.testing_utils import EqualsTester
 
 from openfermion.ops.operators.symbolic_operator import SymbolicOperator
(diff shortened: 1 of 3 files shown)


src/openfermion/ops/operators/symbolic_operator_test.py
```diff
@@ -9,17 +9,20 @@
...
 import copy
+import numpy
+import sympy
 import unittest
 import warnings
 
-import numpy
-import sympy
 from openfermion.config import EQ_TOLERANCE
 from openfermion.testing.testing_utils import EqualsTester
```
pavoljuhas: The imports are usually grouped in 3 alphabetically sorted blocks: (1) builtin imports (2) third_party imports (3) current package imports.

Can you please put back the numpy, sympy imports _after_ builtins?

Same in the test module above.

mhucka:
> LGTM, but please restore the import order per inline comment.

Yeah, I had done it by hand (the pylint setup for this project needs to be updated and doesn't help in this case) and predictably, I got it wrong.

I ran the files through `isort`. @pavoljuhas Hopefully they're correct now?

pavoljuhas (approved):

PR #1114, opened 2025-08-25
Fix #764: wrong SymbolicOperator.isclose() result

### LINKED ISSUE #764: `SymbolicOperator.isclose` bug for operators with many terms?
the following [lines of code](https://github.com/quantumlib/OpenFermion/blob/6a2fdacdd77c5101bc47f3d2bde821f0db717b28/src/openfermion/ops/operators/symbolic_operator.py#L635-L636)  (v1.3.0 release branch) implementing `SymbolicOperator.isclose` are doing something strange when comparing two operators:

```
if not (isinstance(a, sympy.Expr) or isinstance(b, sympy.Expr)):
    tol *= max(1, abs(a), abs(b))
```

This behavior means that operators `x` and `y` whose terms have large coefficients will satisfy `x == y` even when `x - y > 0`. Here is a minimal reproducing example

```
from openfermion import FermionOperator
x = FermionOperator("0^ 0")
y = FermionOperator("0^ 0")

# construct two identical operators up to some number of terms
num_terms_before_ineq = 30
for i in range(num_terms_before_ineq):
    x += FermionOperator(f" (10+0j) [0^ {i}]")
    y += FermionOperator(f" (10+0j) [0^ {i}]")
    
# add a final term that is equal within tol but gets missed by the isclose check
xfinal = FermionOperator(f" (1+0j) [0^ {num_terms_before_ineq + 1}]")
yfinal = FermionOperator(f" (2+0j) [0^ {num_terms_before_ineq + 1}]")
# these two terms are not equal within tol..
print(xfinal == yfinal) # >>> False
print(xfinal - yfinal) # (-1+0j) [0^ 31]

# ...but these two terms will be, because the `tol` argument to `isclose` was corrupted
x += xfinal
y += yfinal
print(x == y) # >>> True
print(x - y) # (-1+0j) [0^ 31]
```
you can tweak `num_terms_before_ineq` smaller and the output will correctly reflect that `x` and `y` are not equal within the desired tolerance. 

Otherwise if this is the intended behavior (i.e. this comparison is meant to be equality within a relative tolerance w/r to magnitude of the coefficient vectors for `x` and `y`), the docstring for `SymbolicOperator.isclose` should be updated accordingly.

mhucka:
Changes:

1. `isclose()` now accepts relative and absolute tolerances. This allows for more precise control over the comparison, just like in NumPy.
2. Adjusted the comparison logic. The old implementation used a single tolerance value that it scaled in a way that could produce incorrect results. The new logic uses a common formula for checking if 1 numbers are close: 
   ```
   abs(a - b) <= abs_tol + rel_tol * max(abs(a), abs(b))
   ```
3. Updated `__eq__` in `SymbolicOperator` to use the tolerance values.
4. Added a new test `symbolic_operator_test.py`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -100,8 +100,9 @@ jobs:
       gha: ${{steps.filter.outputs.gha}}
       gha_files: ${{steps.filter.outputs.gha_files}}
 
-      # The following all test both the relevant file condition & the CI config
-      # because a change in the CI workflows can affect the CI check results.
+      # The rest all test both the relevant files and ALSO tools & scripts b/c
+      # a change in the CI workflows or scripts can affect the check results.
+
       python: ${{steps.filter.outputs.python || steps.filter.outputs.ci}}
       python_files: ${{steps.filter.outputs.python_files}}
 
@@ -159,9 +160,10 @@ jobs:
           # The outputs will be variables named "foo_files" for a filter "foo".
           filters: |
             ci:
-              - './.github/workflows/ci.yml'
-              - './.github/workflows/codeql.yaml'
-              - './.github/workflows/osv-scanner.yaml'
+              - './.github/workflows/**'
+              - './.github/problem-matchers/**'
+              - './check/**'
+              - './dev_tools/**'
             cff:
               - added|modified:
                   - '**/CITATION.cff'
@@ -200,7 +202,7 @@ jobs:
         uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v4
 
       - name: Set up Python with caching of pip dependencies
-        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        uses: actions/setup-python@e797f83bcb11b83ae66e0230d6156d7c80228e7c # v6.0.0
         with:
           python-version: ${{inputs.python_ver || env.python_ver}}
           architecture: 'x64'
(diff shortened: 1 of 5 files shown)


src/openfermion/ops/operators/symbolic_operator.py
```diff
@@ -618,15 +618,17 @@ def __next__(self):
...
 
         Comparison is done for each term individually. Return True
         if the difference between each term in self and other is
-        less than EQ_TOLERANCE
+        less than the specified tolerance.
 
         Args:
             other(SymbolicOperator): SymbolicOperator to compare against.
+            rel_tol(float): Relative tolerance.
+            abs_tol(float): Absolute tolerance.
```
mpharrigan: consider rtol and atol for consistency with numpy/scipy

src/openfermion/ops/operators/symbolic_operator.py
```diff
@@ -618,15 +618,17 @@ def __next__(self):
         term, coefficient = next(self._iter)
         return self.__class__(term=term, coefficient=coefficient)
 
-    def isclose(self, other, tol=EQ_TOLERANCE):
+    def isclose(self, other, rel_tol=EQ_TOLERANCE, abs_tol=EQ_TOLERANCE):
```
mpharrigan: this is technically a breaking API change. are we ok with that? weigh confusion vs. code compatibility to keeping one of them named `tol`

dstrain115: Yeah, maybe we could keep 'tol' as a deprecated parameter and set atol=tol if it exists.

PR #1115, opened 2025-08-25
Do small updates to ci.yaml and rename the nightly Cirq test workflow

mhucka:
* Fix the version comments in `ci.yaml` for `actions/checkout`. The comments said `v4` but in fact the hash is for version `v5`.

* Remove the call to `osv-scanner.yaml` in `ci.yaml`. The scanner file invokes itself on PRs, so this is unnecessary.

* Make `ci.yaml` work for branch `main` as well as `master`. In anticipation of renaming the default branch, the `branches` test for the workflow should work for both.

* Rename nightly pytest workflow. Use a leading underscore in the file name, because that seems to be more conventional in other repos using sub-workflows. Also, Rename the run-name of the workflow itself to be more accurate and informational.


.github/workflows/nightly.yaml
```diff
@@ -34,14 +34,31 @@ permissions: read-all
...
       args: ''
       reason: '(nightly)'
+      cirq-release: 'stable'
 
   cirq-pre:
     name: Nightly Pytest Cirq pre-release
-    uses: ./.github/workflows/nightly-pytest.yaml
+    uses: ./.github/workflows/_nightly-cirq-test.yaml
     with:
       args: '--pre'
```
pavoljuhas: Note the `--pre` argument applies to everything on the dependency chain and can cause spurious failures if some of cirq dependencies publish a breaking pre-release.

```suggestion
      args: 'cirq-core~=1.0.dev'
```

(I have tested `pip install -U "cirq-core" "cirq-core~=1.0.dev0"` installs the latest Cirq pre-release.)

.github/workflows/nightly.yaml
```diff
@@ -34,14 +34,31 @@ permissions: read-all
...
       reason: '(nightly)'
+      cirq-release: 'pre-release'
+
+  osv:
+    name: Nightly OSV code scan
+    uses: ./.github/workflows/osv-scanner.yaml
+    permissions: write-all
```
pavoljuhas: Can this and the scorecard use the same permissions as in the referenced workflows?

Alternatively, you could just add the daily `schedule:` trigger to osv-scanner.yaml and scorecard-scanner.yaml so you don't need to maintain permissions in two places.

mhucka: > Can this and the scorecard use the same permissions as in the referenced workflows?
> 
> Alternatively, you could just add the daily `schedule:` trigger to osv-scanner.yaml and scorecard-scanner.yaml so you don't need to maintain permissions in two places.

Thanks for the feedback. You're right that avoiding a controlling workflow (nightly.yaml) avoids some duplication, simplifies permissions, and is easier to maintain. I was seeking to put some organization on the whole scheme so that people had an easier time understanding the relationships and how different pieces are triggered. But it is turning into more work than expected.

I'll remove nightly.yaml and update the other workflows to do the schedules themselves.

.github/workflows/nightly.yaml
mhucka: Thanks for that info – I actually didn't know the effect of `--pre` extended to other packages. I made this change in the latest commits.

mhucka:
The workflow changes won't show up in the CI checks because they're for weekly runs. Triggering them manually, here are examples of the results of running them in my fork after merging the changes into the master branch:

* OSV: https://github.com/mhucka/OpenFermion/actions/runs/17351346741
* Scorecard: https://github.com/mhucka/OpenFermion/actions/runs/17351127105/

Issue #880, opened 2024-04-30
get_sparse_operator fails on non-simplified QubitOperators


jagunther:
Hi,

I noticed that `get_sparse_operator()` assumes that Paulistrings in a QubitOperator are simplified, i.e. there should be no Paulis within a Paulistring acting on the same tensor-factor. Normally when initialising a QubitOperator the simplification is done automatically, however, the Symmetry-conserving Bravyi-Kitaev transform does not do this. Below is a small example:

```
from openfermion import (
    get_sparse_operator,
    FermionOperator,
    symmetry_conserving_bravyi_kitaev,
)

fermion_op = FermionOperator("0^ 1^")
qubit_op = symmetry_conserving_bravyi_kitaev(fermion_op, 4, 2)
print(qubit_op)

# throws error
get_sparse_operator(qubit_op)
```

noobyogi0010:
Hey @jagunther 
If no one is working on this issue I would like to take it up.

matti-sarjala:
A useful way to frame the fix may be to separate algebraic validity from canonical representation.

The operator produced by `symmetry_conserving_bravyi_kitaev` can still be algebraically meaningful, but `get_sparse_operator()` appears to assume that each Pauli string has already been simplified so that there is at most one Pauli action per qubit.

Possible acceptance criteria:

1. The reproducer in the issue should no longer raise.
2. The resulting sparse operator should match the sparse operator obtained after explicitly simplifying/canonicalizing the `QubitOperator`.
3. Add a regression test for a `QubitOperator` with repeated actions on the same qubit within one term.
4. Decide where the invariant should live:
   - `symmetry_conserving_bravyi_kitaev` always returns a simplified `QubitOperator`, or
   - `get_sparse_operator()` defensively canonicalizes before conversion.

The user-facing boundary would then be clear: `get_sparse_operator()` accepts a mathematically valid `QubitOperator`, not only one that has already been normalized into canonical Pauli-string form.

PR #1128, opened 2025-09-21
Fix: Ensure QubitOperator is simplified before sparse conversion

Aniketsy:
#880 
This PR fixes an issue where `get_sparse_operator` fails when given a non-simplified `QubitOperator`

Please let me know if my approach or fix needs any improvements . I’m open to feedback and happy to make changes based on suggestions.
Thankyou !


mhucka:
@Aniketsy Thanks for this contribution. I took the liberty of merging the latest changes from the main branch and pushing a small change to your fork, in an attempt to resolve some remaining errors and warnings. However, it turns out that the warnings about simplification now show up all over the place in the tests. This is surely an indication that the change to get_sparse_operator() is not the right way to solve the underlying problem. Unfortunately, I also lack the domain expertise to figure out the correct solution.

Aniketsy:
@mhucka thanks for pushing the changes, and I apologize for keeping this PR stale for log time, as i was envolved in other things but I will take a look in this again and i'll try to fix this! :)

Issue #856, opened 2024-01-08
slight modification to function generate_hamiltonian ?


michaelkaicher:
Hi, 

when playing with some active space Hamiltonians with pyscf and generating them as FermionOperator objects (or rather InteractionOperator objects), I found that the resulting Hamiltonian included a lot of unnecessary terms: namely, operators such as 

h_{kklm} a_k^dag a_k^dag a_l a_m, 

where h_{kklm}  was different than zero and the operator thus included (but the operator a_k^dag a_k^dag a_l a_m is zero, so one should just set h_{kklm} to zero right away (k,l,m are here spinorbital indices). 

Maybe this problem could be solved by just adding: 
`if not (p==q or r==s):`
before 
`# Same spin
                    two_body_coefficients[2 * p, 2 * q, 2 * r, 2 *
                                          s] = (two_body_integrals[p, q, r, s] /
                                                2.)
                    two_body_coefficients[2 * p + 1, 2 * q + 1, 2 * r +
                                          1, 2 * s +
                                          1] = (two_body_integrals[p, q, r, s] /
                                                2.)`
in line 272 of [https://github.com/quantumlib/OpenFermion/blob/v1.6.0/src/openfermion/hamiltonians/hartree_fock.py#L245-L291](url)

Would this cause any problems? 

Cheers, 
Michael.

ANUKOOL324:
allow me to look on this issue @fdmalone

matti-sarjala:
A useful way to frame the fix may be to distinguish tensor coefficient storage from the actual fermionic operator algebra.

Even if a two-body tensor entry is nonzero, the corresponding normal-ordered term can still vanish algebraically when it contains repeated creation or repeated annihilation operators on the same spin orbital, e.g.

`a_k^\dagger a_k^\dagger a_l a_m = 0`

or

`a_k^\dagger a_l^\dagger a_m a_m = 0`.

Possible acceptance criteria:

1. Terms with repeated creation indices or repeated annihilation indices should not appear in the generated `FermionOperator`.
2. The generated Hamiltonian should be operator-equivalent to the previous result, but with fewer zero terms.
3. Add a small regression test where a same-spin two-body coefficient would previously generate a term with repeated fermionic operators.
4. Check that this does not remove valid opposite-spin terms, where the spin-orbital indices are different even if they originate from the same spatial orbital.

[...]

PR #1129, opened 2025-09-21
Fix: Exclude zero-valued same-spin two-body terms in generate_hamiltonian

Aniketsy:
#856 

This pull request fixes issue  by adding a condition to the `generate_hamiltonian` function that skips the inclusion of two-body Hamiltonian terms with repeated creation .


mhucka:
I'm not certain about this. I converted the related issue into a discussion. I'm going to close this one-line PR for now.

PR #1144, opened 2025-09-26
Add check/shellcheck

mhucka:
This adds `check/shellcheck` for checking shell scripts. It's adapted from Cirq's check/shellcheck.

### DIFF
--- a/check/shellcheck
+++ b/check/shellcheck
@@ -1,25 +1,38 @@
 #!/usr/bin/env bash
 
-###############################################################################
-# Execute shellcheck for all shell scripts in this repository
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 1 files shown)


check/shellcheck
```diff
@@ -0,0 +1,80 @@
...
+#     check/shellcheck [--dry-run] [shellcheck-arguments]
+#
+# Use the '--dry-run' option to print out the shellcheck command line without
+# running it.  This displays all shell script files found in the repository.
+
+set -uo pipefail
+
+# Change working directory to the repository root.
+thisdir=$(dirname "${BASH_SOURCE[0]:?}")
+repo_dir=$(git -C "${thisdir}" rev-parse --show-toplevel)
```
pavoljuhas, line 29: Please either append `|| exit $?` here as in Cirq or use `set -euo pipefail` above.  With the second option you can remove `exit $?` on the line below.

mhucka: Darn it. I actually had added that exit code in my final version but apparently didn't push my final commit. Thanks for pointing it out. I'll do a fixup PR at some point.

pavoljuhas: Already done in #1147  :-)

mhucka: OK, Mr. Speedy. Thanks!

PR #1146, opened 2025-09-26
Add check/all

mhucka:
This is adapted from, but heavily modified from, Cirq's check/all.

### DIFF
--- a/dev_tools/requirements/Dockerfile
+++ b/dev_tools/requirements/Dockerfile
@@ -19,7 +19,10 @@ WORKDIR /pip-compile
 
 # Step 0: install pip-tools
 COPY envs/pip-tools.env.txt ./
-RUN pip install -r pip-tools.env.txt
+
+# hadolint global ignore=DL3013
+RUN pip install -U pip --no-cache-dir --root-user-action ignore && \
+    pip install -r pip-tools.env.txt --no-cache-dir --root-user-action ignore
 
 # Step 1: compile a complete & consistent environment with all dependencies
 COPY deps/ ./deps/
--- a/dev_tools/requirements/deps/runtime.txt
+++ b/dev_tools/requirements/deps/runtime.txt
@@ -1,4 +1,4 @@
-cirq-core>=1.4.1
+cirq-core>=1.4.1,<1.6.0
 deprecation
 h5py>=3.10.0
 networkx
@@ -9,3 +9,13 @@ requests~=2.32.2
 scipy~=1.15
 
 sympy
+
+# Transitive dependency. Versions 1.3.3+ require Python 311. Constraining it
+# lets us produce a single set of requirements files for Python 3.10 - 3.12.
+contourpy<1.3.3
+
+# Another transitive dependency. We don't care about the version of this
+# backport because it's only needed for a pytest extension. However, because we
+# run pip-compile in Python 3.10 for maximum compatibility, we need to add an
+# explicit constraint to keep this package out of the requirements files.
+backports-asyncio-runner ; python_full_version < '3.11'
--- a/dev_tools/requirements/envs/dev.env.txt
+++ b/dev_tools/requirements/envs/dev.env.txt
@@ -4,58 +4,64 @@
 #
 #    pip-compile --output-file=envs/dev.env.txt deps/format.txt deps/mypy.txt deps/pip-tools.txt deps/pylint.txt deps/pytest.txt deps/resource_estimates_runtime.txt deps/runtime.txt
 #
-ase==3.25.0
+ase==3.27.0
     # via -r deps/resource_estimates_runtime.txt
-astroid==3.3.10
+astroid==3.3.11
     # via pylint
-attrs==25.3.0
+attrs==25.4.0
     # via
     #   cirq-core
     #   jsonschema
     #   referencing
+backports-asyncio-runner==1.2.0 ; python_full_version < "3.11"
+    # via
+    #   -r deps/runtime.txt
+    #   pytest-asyncio
 black==26.3.1
     # via -r deps/format.txt
-build==1.2.2.post1
+build==1.4.0
     # via pip-tools
-certifi==2025.4.26
+certifi==2026.2.25
     # via requests
-charset-normalizer==3.4.2
+charset-normalizer==3.4.5
     # via requests
 cirq-core==1.5.0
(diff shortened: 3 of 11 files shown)


check/all
```diff
@@ -0,0 +1,125 @@
...
+Usage: ${0##*/} [--help] [--only-changed-files] [--apply-format-changes] [BASE_REV]
+
+If the first argument given on the command line is the option --help or -h,
+this program prints usage information and then exits.
+
+With no arguments, this runs multiple checks on the code base. These tests
+are based on the programs in the checks/ subdirectory.
+
+If the option --no-coverage is specified, check/pytest will be run instead of
+check/pytest-and-incremental-coverage.
```
pavoljuhas: I fail to see any code for `--no-coverage`.  Either remove this paragraph or handle the option in the script.  Since passing coverage is required by the CI, it seems better to just change the text and keep running pytest-and-incremental-coverage unconditionally.

check/all
```diff
@@ -0,0 +1,125 @@
...
+the argument BASE_REV is passed on to tests that can use it, such as
+check/pytest-and-incremental-coverage."
+
+set -eo pipefail -o errtrace
+shopt -s inherit_errexit
+
+# Get the working directory to the repo root.
+thisdir=$(dirname "${BASH_SOURCE[0]:?}") || exit $?
+repo_dir=$(git -C "${thisdir}" rev-parse --show-toplevel) || exit $?
+cd "${repo_dir}" || exit $?
```
pavoljuhas: Nit - remove `|| exit $?` because it is implied by the errexit (-e) option on line 41.

check/all
```diff
@@ -0,0 +1,125 @@
...
+
+function error() {
+    echo >&2 "ERROR: ${*}"
+}
+
+# ~~~~ Parse the CLI arguments and gather some data ~~~~
+
+declare rev
+declare apply_arg
+declare only_changed
```
pavoljuhas: Nit - it is safer to set initial value, e.g., `declare rev=""`, otherwise the `rev` value could be picked from the `rev` environment variable.  Also, as `rev` and `apply_arg` are optional for format-incremental, it is better to declare them as arrays which may be empty (and would not trigger shellcheck):
```
declare -a rev=()
declare -a apply_arg=()
```

check/all
```diff
@@ -0,0 +1,125 @@
...
+        --only-changed-files)
+            only_changed="true"
+            shift
+            ;;
+        -*)
+            if [[ -n "${rev}" ]]; then
+                error "Invalid arguments."
+                echo "${usage}"
+                exit 1
+            fi
```
pavoljuhas: This silently skips `--no-coverage` option or any other unknown option if there is no `rev` argument.  How about
```suggestion
            error "Invalid option '${arg}'"
            error "See '$0 --help' for the list of supported options."
            exit 1
```

check/all
```diff
@@ -0,0 +1,125 @@
...
+            ;;
+        *)
+            if [ "$(git cat-file -t "${arg}" 2> /dev/null)" != "commit" ]; then
+                error "No revision '${arg}'"
+                exit 1
+            fi
+            rev="${arg}"
```
pavoljuhas: This will reject annotated tag wrapping a commit, because its type is "tag" (this is an issue in Cirq too).  Here is a better way:
```suggestion
            if ! rev=( "$(git rev-parse --verify --end-of-options "${arg}^{commit}")" ); then
                error "No revision '${arg}'"
                exit 1
            fi
```

check/all
```diff
@@ -0,0 +1,125 @@
...
+            ;;
+    esac
+done
+
+# ~~~~ Run the tests ~~~~
+
+declare -a errors
+declare program
```
pavoljuhas: `bash` seems to initialize `errors[0]` from the environment.  The `program` is not used on global scope.
 ```suggestion
declare -a errors=()
```

File: check/all [hunk 7]
```diff
@@ -0,0 +1,125 @@
...
+
+function run() {
+    local -r program="$1"
+    echo "Running $program ..."
+    $program || errors+=( "${program} failed" )
+    echo
```
pavoljuhas: Passing the entire command line as one argument (e.g., line 102) is fragile, in case any of desired arguments contain space or a glob-like character which would get expanded.  It is safer to pass the command line arguments exactly:
```suggestion
    echo "Running $* ..."
    "$@" || errors+=( "$* failed" )
    echo
```

check/all
```diff
@@ -0,0 +1,125 @@
...
+}
+
+if [[ -n "${only_changed}" ]]; then
+    run "check/format-incremental ${rev} ${apply_arg}"
+    run "check/pylint-changed-files ${rev}"
```
pavoljuhas: With `rev` and `apply_arg` declared as arrays, we can pass arguments to the `run` function exactly (similar change needed on lines 105, 109 below):
```suggestion
    run check/format-incremental "${rev[@]}" "${apply_arg[@]}"
    run check/pylint-changed-files "${rev[@]}"
```

check/all
mhucka: I removed the text. The next push will have the change.

check/all
mhucka: The next push will have the change.

check/all
mhucka: I added the declarations. (I did not know that declared variables could be picked up from the environment.)

W.r.t. using arrays, the values are never arrays; they are only a single value, so I feel it's more clear not to pretend they are, and leave them as strings. (When I see "declare -a" in someone's script, I then look for how the variable could have more than one value, whether it's referenced using `${foo[@]}`, etc.) Plus, I don't get any shellcheck warnings – do you?

pavoljuhas: The point is both `rev` and `apply_arg` are _optional_ arguments to other scripts which can be expressed as an array of zero or one element.  Having them as a string would make shellcheck complain if unquoted.  If quoted, an empty `"${rev}"` is passed as an actual argument of an empty string, which may cause unexpected behavior.  Example:

```bash
$ s=""; sh -c 'echo $#' -- a "${s}" b
3

$ declare -a opt=(); sh -c 'echo $#' -- a "${opt[@]}" b
2

$ opt=( value ); sh -c 'echo $#' -- a "${opt[@]}" b
3
```

mhucka: Next commit will do this.

PR #1150, opened 2025-09-30
Avoid early exit in check/shellcheck due to spurious read failure

pavoljuhas:
Problem: In bash `read -r -d "" VAR` expects a NUL character to terminate
the input line.  If missing read returns a failure exit status and thus
quits the shellcheck script due to its errexit option.

Solution: Terminate read input with the NUL character.


mhucka (approved):
Thanks for noticing this.

PR #1257, opened 2026-04-06
Replace assert in reverse_jordan_wigner.py

mhucka:
Replace an `assert` in a non-test context with raising an exception.

### DIFF
--- a/src/openfermion/transforms/opconversions/reverse_jordan_wigner.py
+++ b/src/openfermion/transforms/opconversions/reverse_jordan_wigner.py
@@ -82,7 +82,7 @@ def reverse_jordan_wigner(qubit_operator, n_qubits=None):
 
                 # Get next non-identity operator acting below 'working_qubit'.
                 if len(working_term.terms) != 1:
-                    raise ValueError('Qubit operator term needs to be a single term')
+                    raise ValueError('QubitOperator must contain exactly one term')
                 working_qubit = pauli_operator[0] - 1
                 for working_operator in reversed(list(working_term.terms)[0]):
                     if working_operator[0] <= working_qubit:
--- a/src/openfermion/transforms/opconversions/reverse_jordan_wigner.py
+++ b/src/openfermion/transforms/opconversions/reverse_jordan_wigner.py
@@ -82,7 +82,10 @@ def reverse_jordan_wigner(qubit_operator, n_qubits=None):
 
                 # Get next non-identity operator acting below 'working_qubit'.
                 if len(working_term.terms) != 1:
-                    raise ValueError('QubitOperator must contain exactly one term')
+                    raise ValueError(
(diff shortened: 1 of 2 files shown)


src/openfermion/transforms/opconversions/reverse_jordan_wigner_test.py
```diff
@@ -160,6 +161,14 @@ def test_bad_type(self):
         with self.assertRaises(TypeError):
             reverse_jordan_wigner(3)
 
+    def test_reverse_jw_multi_term_error(self):
+        with patch(
+            'openfermion.transforms.opconversions.reverse_jordan_wigner.' 'QubitOperator.__mul__'
+        ) as mock_mul:
+            mock_mul.return_value = QubitOperator('X0') + QubitOperator('Y0')
+            with self.assertRaisesRegex(ValueError, 'QubitOperator must contain exactly one term'):
+                reverse_jordan_wigner(QubitOperator('X1'))
```
mpharrigan, line 172: why do you need to mock stuff?

src/openfermion/transforms/opconversions/reverse_jordan_wigner.py
```diff
@@ -81,7 +81,8 @@ def reverse_jordan_wigner(qubit_operator, n_qubits=None):
                     working_term.terms[list(working_term.terms)[0]] = 1.0
 
                 # Get next non-identity operator acting below 'working_qubit'.
-                assert len(working_term.terms) == 1
+                if len(working_term.terms) != 1:
+                    raise ValueError('QubitOperator must contain exactly one term')
```
mpharrigan: maybe include the actual count or the repr of the terms in the error message to aid debuggging

mhucka: OK. Done in the latest push.

src/openfermion/transforms/opconversions/reverse_jordan_wigner_test.py
mhucka: It's to simulate an internal failure condition. The code uses unittest.mock.patch to replace the `__mul__` method of `QubitOperator`, specifically where it is used inside the `reverse_jordan_wigner` module, so that it can force any multiplication operation to return a multiterm operator (`X0 + Y0`) instead of a single Pauli string.

Maybe I'm missing another, easier way? (Entirely possible …)

mpharrigan: Naively, I would think that if you can't trigger the failure condition without mucking up the library code itself (i.e. it can't be triggered by bad user input), then I would think an assertion would be the more idiomatic way to guard this

mhucka:
I suspect the problem is that the function being tested should be changed somehow. Anyway, I'm going to close this because this doesn't seem worth spending more time on.

PR #1262, opened 2026-04-10
Avoid inefficient list construction in `qubit_tapering_from_stabilizer.py`

mhucka:
This replaces some inefficient `list += [item]` patterns with `list.append(item)` in `_reduce_terms` and `_reduce_terms_keep_length` of the file `qubit_tapering_from_stabilizer.py`. This alternative approach avoids unnecessary temporary list creation and is a standard Python optimization for iterative list building.


mpharrigan:
do you have any performance benchmarks that back up this change?

mhucka:
> do you have any performance benchmarks that back up this change?

@mpharrigan Here is a short benchmark program to demonstrate the performance of `append`:

```python
import timeit

setup_code = """
def construct_with_iadd(n):
    l = []
    for i in range(n):
        l += [i]
    return l

def construct_with_append(n):
    l = []
    for i in range(n):
        l.append(i)
    return l
"""

sizes = [10, 100, 1000, 10000, 100000]

print("| N items | `list += [item]` (ms) | `list.append(item)` (ms) | Speedup |")
print("|---------|-----------------------|--------------------------|---------|")

for n in sizes:
    # Scale iterations down for larger lists to keep the benchmark fast
    number = 10000 if n < 100000 else 1000

    t_iadd = timeit.timeit(f"construct_with_iadd({n})", setup=setup_code, number=number)
    t_append = timeit.timeit(f"construct_with_append({n})", setup=setup_code, number=number)

    # Calculate average time per call in milliseconds
    avg_iadd = (t_iadd / number) * 1000
    avg_append = (t_append / number) * 1000
    speedup = avg_iadd / avg_append

    print(f"| {n:<7} | {avg_iadd:<21.4f} | {avg_append:<24.4f} | **{speedup:.2f}x** |")
```

Running this program on a Linux VM using Python 3.12.11 produces the following output:

| N items | `list += [item]` (ms) | `list.append(item)` (ms) | Speedup |
|---------|-----------------------|--------------------------|---------|
| 10      | 0.0043                | 0.0010                   | **4.33x** |
| 100     | 0.0460                | 0.0147                   | **3.12x** |
| 1000    | 0.8182                | 0.3715                   | **2.20x** |
| 100000  | 100.7308              | 26.2301                  | **3.84x** |


Would this be a sufficient benchmark?

mpharrigan:
I was thinking more of an openfermion-based benchmark, but if you think it's worth it then lgtm

mpharrigan (approved):

PR #1279, opened 2026-04-23
Move pytest options to pyproject.toml

### LINKED ISSUE #1268: Move tool configuration settings into `pyproject.toml`

OpenFermion currently puts the configuration files for some tools like pylint into `dev_tools/conf/`. This location is unexpected for most modern developers, who probably expect to find the config files at the top level of the repo.

 We should follow the more modern approach of putting such configurations in the top-level `pyproject.toml` file.

None

mhucka:
In addition to being the more modern approach, this lets us start to consolidate tool options into a single well-known location, and also lets us simplify the check scripts.

One of the PRs that will resolve https://github.com/quantumlib/OpenFermion/issues/1268.


pyproject.toml
```diff
@@ -23,17 +23,23 @@ skip-magic-trailing-comma = true
...
 
-# Omit files outside the current working directory.
-# Note: this means coverage must be run from the openfermion repo root.
-# Failure to do so will result in false positives.
+# Omit some paths. Note: this means coverage must be run from the OpenFermion
+# repo root. Failure to do so will result in false positives.
 omit = [
-    "./profiling/performance_benchmarks.py",
-    "./dev_tools/*",
+    "./git/*",
```
pavoljuhas: I think it should be `.git`, `.github` instead of `git`, `github`.

That said are these necessary at all?  I would expect that coverage would skip the `.git` directory by default as that must be a very common case.  `.github` does not have any python files so excluding it is redundant.

pavoljuhas (approved):
Please fix the `omit` list in pyproject.toml before submitting.

Otherwise LGTM.

pyproject.toml
mhucka: Oh, yah, that was meant to be `.git`. Great catch.

I agree it would make sense for python-cov to skip that, but when I looked, I couldn't find information that it did that. I just looked again in the python-cov repo and the coveragepy repo, and can't find evidence that it does.

mhucka: Ah, hold on, what if it gets the paths from pytest? Pytest by default does ignore ".*" unless `norecursedirs` is set to a value (https://docs.pytest.org/en/stable/reference/reference.html#confval-norecursedirs). OpenFermion's configuration doesn't set that variable, so the default should be in effect.

I'll take out that value.

PR #1280, opened 2026-04-23
Move mypy options to pyproject.toml

### LINKED ISSUE #1268: Move tool configuration settings into `pyproject.toml`

mhucka:
In addition to being the more modern approach, this lets us start to consolidate tool options into a single well-known locadtion instead of burying them in a nonstandard location like `dev_tools/conf/`.

One of the PRs that will resolve https://github.com/quantumlib/OpenFermion/issues/1268.


pyproject.toml
```diff
@@ -22,3 +22,42 @@ skip-magic-trailing-comma = true
 # Speed up pytest-cov on Python 3.12+ by enabling sys.monitoring if possible.
 core = "sysmon"
 disable_warnings = ["no-sysmon"]
+
+[tool.mypy]
+implicit_optional = true
+# TODO: this is temporary for 1.7.0 release. Remove this & fix the errors.
+ignore_errors = true
```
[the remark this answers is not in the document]
mhucka: The errors should be fixed, but for the purposes of moving the mypy settings to pyproject.toml, we should keep the settings the same for a cleaner change history.

pyproject.toml
```diff
@@ -22,3 +22,42 @@ skip-magic-trailing-comma = true
...
+    "numpy.*",
+    "pytest.*",
+    "scipy.*",
+    "_pytest.*",
+    "h5py",
+    "pyscf.*",
+    "pubchempy",
+    "pybtas",
+    "absl",
+    "tensorflow_docs"
```
pavoljuhas: Consider sorting these.

pavoljuhas (approved):
LGTM with a tiny suggestion.

pyproject.toml
mhucka: Done in the next push.

pavoljuhas (approved):

PR #1288, opened 2026-04-27
Skip Pubchem live API test in CI

mhucka:
Even with many retries and long delays, the test fails for almost every PR. I don't know why this has been happening lately; perhaps Pubchem's servers have become overloaded lately (possibly due to an increase in the use of AI agents?). In any case, there seems to be no good solution to stop flaky failures in CI except to simply skip the test in CI.

The approach implemented here tests an environment variable set by GitHub in their runner environments. If that variable is set, the live API test is skipped; otherwise, it's run as usual. This means that developers in their normal local environments will run the tests as usual, and don't have to do anything special. In addition, our CI workflows don't need to change either.

### DIFF
--- a/src/openfermion/chem/pubchem_test.py
+++ b/src/openfermion/chem/pubchem_test.py
@@ -78,7 +78,7 @@ using_pubchempy = pytest.mark.skipif(
 
 # Skip if running in a CI environment.
 skip_in_ci = pytest.mark.skipif(
-    os.environ.get('CI') == 'true',
+os.environ.get('CI'),
     reason='Skipping Pubchem API tests in CI to avoid failures due to busy servers.',
 )
 
--- a/src/openfermion/chem/pubchem_test.py
+++ b/src/openfermion/chem/pubchem_test.py
@@ -158,7 +158,7 @@ class TestOpenFermionPubChem:
             _ = geometry_from_pubchem('water', structure='foo')
 
     @skip_in_ci
-    @pytest.mark.flaky(retries=4, delay=30, only_on=[pubchempy.ServerBusyError])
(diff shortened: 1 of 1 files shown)


src/openfermion/chem/pubchem_test.py
```diff
@@ -74,6 +76,12 @@ def mock_get_compounds(name, searchtype, record_type='2d'):
     module_importable('pubchempy') is False, reason='Not detecting `pubchempy`.'
 )
 
+# Skip if running in a CI environment.
+skip_in_ci = pytest.mark.skipif(
+    os.environ.get('CI', False) != False,
```
pavoljuhas: It is usually an anti-pattern to compare booleans; also environment variables are strings.  Let us rewrite in a cleaner equivalent way:

```suggestion
    'CI' in os.environ,
```

src/openfermion/chem/pubchem_test.py
```diff
@@ -74,6 +76,12 @@ def mock_get_compounds(name, searchtype, record_type='2d'):
     module_importable('pubchempy') is False, reason='Not detecting `pubchempy`.'
 )
 
+# Skip if running in a CI environment.
+skip_in_ci = pytest.mark.skipif(
```
pavoljuhas: This is used only once.  Consider using directly as a test decorator without an extra variable indirection.

pavoljuhas (changes requested):
LGTM after addressing inline suggestions.

src/openfermion/chem/pubchem_test.py
mhucka: Good point! Done in the latest push.

mhucka:
> LGTM after addressing inline suggestions.

Thanks for those great suggestions.

pavoljuhas (approved):

PR #1293, opened 2026-04-28
Fix 2 problems in `_multitensor.py`

mhucka:
The `dual_basis` parameter to the `MultiTensor` class constructor was defaulted to `DualBasis()`. In Python, default arguments are evaluated only once at definition time, meaning every `MultiTensor` instance created without an explicit basis shared the same global `DualBasis` object. When tests were run in parallel or repeated, data from one test would leak into others, leading to errors.

In addition, `MultiTensor.add_dual_elements` used `list.extend()` instead of `list.append()` when adding `DualBasisElement` objects. Since `DualBasisElement` is iterable (yielding its internal terms), using `extend()` incorrectly decomposed the object into its constituent tuples instead of adding the object as a single unit. This caused `MultiTensor.synthesize_dual_basis()` to fail in some cases.

### DIFF
--- a/src/openfermion/contrib/representability/_multitensor.py
+++ b/src/openfermion/contrib/representability/_multitensor.py
@@ -23,7 +23,7 @@ class TMap(object):
 
 
 class MultiTensor(object):
-    def __init__(self, tensors, dual_basis=DualBasis()):
+    def __init__(self, tensors, dual_basis=None):
         """
         A collection of tensor objects with maps from name to tensor
 
@@ -46,6 +46,8 @@ class MultiTensor(object):
         self.off_set_map = self.make_offset_dict(self.tensors)
 
         # An iterable object that provides access to the dual basis elements
+        if dual_basis is None:
+            dual_basis = DualBasis()
         self.dual_basis = dual_basis
         self.vec_dim = sum([vec.size for vec in self.tensors])
 
@@ -82,7 +84,7 @@ class MultiTensor(object):
             raise TypeError("dual_element variable needs to be a DualBasisElement type")
 
         # we should extend TMap to add
-        self.dual_basis.elements.extend(dual_element)
+        self.dual_basis.elements.append(dual_element)
 
     def synthesize_dual_basis(self):
         """
--- a/src/openfermion/contrib/representability/_multitensor.py
+++ b/src/openfermion/contrib/representability/_multitensor.py
@@ -95,7 +95,7 @@ class MultiTensor(object):
 
         :returns: sparse matrix
         """
-        # go throught the dual basis list and synthesize each element
+        # go through the dual basis list and synthesize each element
         dual_row_indices = []
         dual_col_indices = []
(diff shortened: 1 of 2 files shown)


dstrain115 (approved):

src/openfermion/contrib/representability/_multitensor.py
```diff
@@ -82,7 +84,7 @@ def add_dual_elements(self, dual_element):
             raise TypeError("dual_element variable needs to be a DualBasisElement type")
 
         # we should extend TMap to add
-        self.dual_basis.elements.extend(dual_element)
+        self.dual_basis.elements.append(dual_element)
```
mpharrigan, line 86: can you update or remove the comment above this.

Are we sure `add` is the intended operation here?

mpharrigan: The function name is plural but the argument name is singular. The docstring says "add" but the inline comment says "extend". maybe this was originally an extend-type operation and was changed to an append-type?

mhucka: > can you update or remove the comment above this.

Ah, yes, good point. Thanks for catching that. Fixed in the next push.

> The function name is plural but the argument name is singular. The docstring says "add" but the inline comment says "extend". maybe this was originally an extend-type operation and was changed to an append-type?

Yeah, something probably happened there. I also think the bug was never caught before now because the tests in `_multitensor_test.py` were incomplete. (I only ran into the problem because I was working on getting parallel Pytest working, which revealed the problem of initializing `MultiTensor` with a mutable object, which led to more poking around.)

Here's what's happening:

1. `DualBasisElement` objects are iterable because they implement the `__iter__()` magic method (c.f. `_dualbasis.py`). Each call to that method returns a tuple of 3 items.

2. In Python, doing `somelist.extend(someobject)` invokes the `__iter__` method on `someobject`. On a `DualBasisElement` object, doing that ends up adding a tuple of 3 items to `somelist`. That's in contrast with the operation `somelist.append(someobject)`, which puts the `DualBasisElement` object itself on `somelist`.

3. The previous `_multitensor_test.py` only tested adding an item to the list `dual_basis.elements`, and never tested _using_ the results. The place where the results are used are in the method `synthesize_dual_basis()`. That method eventually calls the `synthesize_element()` method on each item on the list, and therein lies the problem: [`synthesize_element()` does a loop](https://github.com/quantumlib/OpenFermion/blob/40caf92b0396e5b52debce1f38d93644d32d2e49/src/openfermion/contrib/representability/_multitensor.py#L135) like this:

    ```python
    for tlabel, velement, coeff in element:
    ```
    That works as expected if the values on the `elements` list in `DualBasis` are actual `DualBasisElement` objects because their `__iter__()` method will get called in the `for` loop (yielding the 3 items above), but that _won't_ happen if the values on the `elements` list are tuples. In the latter case, an error will result.

Issue #1060, opened 2025-06-02
Remove global random seed used for testing


mhucka:
PR #1056 modified the pytest configuration to set a global random number seed, to deal with flaky tests that failed sporadically because some numerical comparisons exceeded tolerances undpredictably. Setting a global seed is not good practice. The seeds for test scenarios should be revised:

* Avoid global seeds. It's generally better to avoid setting a global seed for all tests. Global seeds can cause unexpected behavior when different parts of the code rely on random numbers.
 
* Use `numpy.random.default_rng`. It's considered better to create a `Generator` instance with a seed at the beginning of each test or test suite. This ensures that each test has its own independent random state.
 
* Pass `Generator` objects. Pass the `Generator` object around as an argument. This makes the random number generation explicit and avoids relying on global state.

matti-sarjala:
A safe way to split this might be to avoid removing the global seed in the first PR.

Possible incremental path:

1. Identify tests that currently rely on implicit global randomness.
2. Convert one small test module to use local `numpy.random.default_rng(seed)` instances.
3. Pass the `Generator` explicitly where helper functions need randomness.
4. Confirm that the converted module still passes when run repeatedly.
5. Only after enough modules are converted, remove the global pytest seed as a final cleanup PR.

This would reduce the risk of turning one global-seed cleanup into many unrelated flaky-test investigations.

A minimal first PR could target just one test file and follow the invariant:

“randomness used by a test should be local, explicit, and reproducible from the test itself.”

fzlzjerry:
I'll take the first incremental slice in `_multitensor_test.py`: replace its uses of the legacy `np.random` global state with a fresh, deterministically seeded `numpy.random.Generator` fixture passed into each randomized test. I'll leave the global fixture in place so unrelated modules are not destabilized, and run this module repeatedly with varied `pytest-randomly` seeds.

PR #1308, opened 2026-05-14
Use local random seeds in slater determinant tests

mzl2233:
This updates the Slater determinant Gaussian-state tests to keep their random choices local to the test case. The Hamiltonian construction now receives explicit per-test seeds, and the excited-state orbital selections use a local NumPy generator instead of the module-level random state, which makes this module less dependent on the global pytest seed while preserving deterministic coverage. Verified with `python3 -m compileall -q src/openfermion/circuits/slater_determinants_test.py`, `python3 -m pytest src/openfermion/circuits/slater_determinants_test.py -q -k 'JWGetGaussianStateTest'`, and `git diff --check`. Part of #1060.

### DIFF
--- a/src/openfermion/circuits/slater_determinants_test.py
+++ b/src/openfermion/circuits/slater_determinants_test.py
@@ -137,13 +137,16 @@ class GaussianStatePreparationCircuitTest(unittest.TestCase):
 class JWGetGaussianStateTest(unittest.TestCase):
     def setUp(self):
         self.n_qubits_range = range(2, 10)
+        self.rng = numpy.random.default_rng(5387)
+        self.hamiltonian_seed_sequence = numpy.random.SeedSequence(5387)
 
     def test_ground_state_particle_conserving(self):
         """Test getting the ground state of a Hamiltonian that conserves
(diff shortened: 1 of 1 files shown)


mhucka:
@mzl2233 Thank you for this work! When you get a chance, could you please take care of signing the contributor license agreement? There is some more information about what to do at https://github.com/quantumlib/OpenFermion/pull/1308/checks?check_run_id=76033891071

(It looks complicated, but it's pretty straightforward. We need it so that we are legally allowed to distribute contributions as part of this open-source project.)

mhucka:
@mzl2233 Hello again! Just checking back to ask if there's anything I can do to help with the matter of the CLA (contributor license agreement). Please let me know.

mhucka:
@mzl2233 In addition to the points above, please also read `CONTRIBUTING.md` for information about how to do proper testing for this project. For example, we don't use `python3 -m compileall`; we use other command and scripts.

So far, this PR has the look of something produced blindly by an LLM. Without confirmation that you are actively involved in this work, I will have to close this PR.

PR #1320, opened 2026-05-30
Added shellcheck dependency

### LINKED ISSUE #1318: shellcheck dependency is not installed
### Describe the issue

Running `check/shellcheck` (and therefore `check/all`) will fail unless the user has manually installed shellcheck. This could potentially be avoided with [shellcheck-py](https://pypi.org/project/shellcheck-py/).

1.7.2.dev0

Run `check/shellcheck`

arettig:
Added shellcheck-py as a development dependency as it is used in `check/shellcheck` and the optional pre-commit hooks.


mhucka (approved):
Clean and focused. Nice!

PR #1325, opened 2026-06-02
Fix RDM basis set rotation

### LINKED ISSUE #603: InteractionRDM.rotate_basis rotates incorrectly?
It looks like InteractionRDM.rotate_basis calls the same PolynomialTensor routine as rotate_basis on InteractionOperator.  RDM rotation should be
```
opdm = of.general_basis_change(opdm, v, (0, 1))
tpdm = of.general_basis_change(tpdm, v, (0, 0, 1, 1))
```

and not
```
opdm = of.general_basis_change(opdm, v, (1, 0))
tpdm = of.general_basis_change(tpdm, v, (1, 1, 0, 0))
```

arettig:
This fixes #603, where RDM basis set rotations incorrectly used the same rotation matrices as operators instead of the conjugate. 

The `rotate_basis()` method is now overridden in `interaction_rdm.py`. 

Added new tests that check the `rotate_basis()` method and will fail if old convention is used.

### DIFF
--- a/src/openfermion/ops/representations/interaction_rdm.py
+++ b/src/openfermion/ops/representations/interaction_rdm.py
@@ -61,6 +61,23 @@ class InteractionRDM(PolynomialTensor):
         """Set the value of the two-body tensor."""
         self.n_body_tensors[1, 1, 0, 0] = value
 
+    def rotate_basis(self, rotation_matrix):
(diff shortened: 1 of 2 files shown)


mhucka (approved):
Nice! Thank you for this.

PR #1339, opened 2026-06-04
Rerun flaky test in vpe_circuits_test.py

arettig:
This test was causing intermittent failures of the CI in the 5% of runs where the measured value is outside 2 standard deviations from the expected value. The built in `@cirq.testing.retry_once_with_later_random_values` decorator will now rerun the test with a new random seed.


arettig:
Looks like `@cirq.testing.retry_once_with_later_random_values` was added in cirq v1.5.0 but our `ci.yaml` uses v1.4.1:
 https://github.com/quantumlib/OpenFermion/blob/efaadaeecd36631ffe45afa45e451324a7e0078e/.github/workflows/ci.yaml#L203

@mhucka  Is continuing support for 1.4.1 a hard requirement, or is it ok to bump that up in the CI?

mhucka:
> @mhucka Is continuing support for 1.4.1 a hard requirement, or is it ok to bump that up in the CI?

It is probably not a hard requirement, and now that you mention it, we should upgrade it to support the latest Cirq version if possible. 

The pin in dev_tools/requirements/deps/runtime is actually a bit wider:

```
cirq-core>=1.4.1,<1.6.0
```

It ends up being 1.4.1 when pip-compile processes all the transitive dependencies, but I can no longer remember what the problem was the last time I tried to go higher than 1.6.

I'll open an issue about this.

mhucka:
@arettig Quick update: I believe I worked out the changes needed to the requirements to make it possible to get higher versions of Cirq on Python versions > 3.10. I should have a PR ready tomorrow.

arettig:
@mhucka Thanks for fixing the Cirq versioning! It looks like this flaky test fix works now.

mhucka:
> @mhucka Thanks for fixing the Cirq versioning! It looks like this flaky test fix works now.

There is one thing that needs to be adjusted in the current PR. The decorator `cirq.testing.retry_once_with_later_random_values` is only available in Cirq versions 1.5.0+, which only supports Python 3.11+, but OpenFermion still can be installed in Python 3.10. If someone tries to run the tests in Python 3.10, they'll get an error about the undefined symbol. So, we need to protect against that. One way is to put something like this at the top of the file:

```python
# Try to import the retry decorator (introduced in Cirq 1.5.0).
# If running on Cirq < 1.5.0, fall back to a no-op decorator.
try:
    retry_once_with_later_random_values = cirq.testing.retry_once_with_later_random_values
except AttributeError:
    def retry_once_with_later_random_values(func):
        return func
```

and then change the use of the decorator to be

```python
@retry_once_with_later_random_values
def test_single_timestep():
```

This is not a great solution, but it should work.

Can I leave it to you to make the necessary changes to this PR?

arettig:
I saw there is a `wrapped.py` file for sorting out versioning issues with Cirq so I moved this fix there. In the case of an older Cirq version, I manually define it using the implementation copied from Cirq rather than a dummy function.

mhucka (approved):
Whoa, even better. Thanks!

PR #1343, opened 2026-06-06
Fix QASM Pauli exponential Rz angle

### LINKED ISSUE #721: Angle of rotation in trotterize_exp_qubop_to_qasm is off by a factor of 2
Hello,

I'm a student using OpenFermion for a project, and I'm having a problem with openfermion.circuits.trotterize_exp_qubop_to_qasm. I was trying to use it to trotterize the UCCSD operator, but it was working strangely.

I saw that it uses trotter_operator_grouping to decompose the operators, then it calls pauli_exp_to_qasm to trotterize each operator.

And I don't understant something regarding the Z rotation performed on the last qubit of 'qids' in pauli_exp_to_qasm. 

```
                    ret_list = ret_list + [
                        "Rz {} {}".format(term_coeff * evolution_time, qids[-1])
                    ]
```

Why is the angle (term_coeff * evolution_time), rather than double that? Wouldn't the goal be applying e^(-i * term_coeff * evolution_time * Z) through Rz(2 * term_coeff * evolution_time)? Am I missing something?

Thanks in advance!

marko1olo:
Fixes #721.

## Summary
- doubles the uncontrolled `Rz` angle emitted by `pauli_exp_to_qasm` so it implements `exp(-i * theta * Z)` rather than `exp(-i * theta * Z / 2)`
- updates existing QASM expectations for Pauli-Z strings
- adds a direct regression test with non-default `evolution_time`

## Validation
- `C:\hades\oss\.venvs\openfermion-1317\Scripts\python.exe -m pytest src\openfermion\circuits\trotter_exp_to_qgates_test.py -q`
- `C:\hades\oss\.venvs\openfermion-1317\Scripts\python.exe -m black --check src\openfermion\circuits\trotter_exp_to_qgates.py src\openfermion\circuits\trotter_exp_to_qgates_test.py`
- `C:\hades\oss\.venvs\openfermion-1317\Scripts\python.exe -m compileall -q src\openfermion\circuits\trotter_exp_to_qgates.py src\openfermion\circuits\trotter_exp_to_qgates_test.py`
- `git diff --check`

AI-assisted contribution: implemented with OpenAI Codex and reviewed locally before submission.

### DIFF
--- a/src/openfermion/circuits/trotter_exp_to_qgates.py
+++ b/src/openfermion/circuits/trotter_exp_to_qgates.py
@@ -255,7 +255,9 @@ def pauli_exp_to_qasm(qubit_operator_list, evolution_time=1.0, qubit_list=None, 
                     ]
             else:
                 if len(qids) > 0:
-                    ret_list = ret_list + ["Rz {} {}".format(term_coeff * evolution_time, qids[-1])]
+                    ret_list = ret_list + [
+                        "Rz {} {}".format(2 * term_coeff * evolution_time, qids[-1])
+                    ]
 
             # 4. Second set of CNOTs
             ret_list = ret_list + cnots2
--- a/src/openfermion/circuits/trotter_exp_to_qgates_test.py
+++ b/src/openfermion/circuits/trotter_exp_to_qgates_test.py
@@ -210,10 +210,15 @@ class TrottQasmTest(unittest.TestCase):
         # Correct string
         strcorrect = '''5
 CNOT 3 4
-Rz 0.6 4
+Rz 1.2 4
 CNOT 3 4'''
 
         self.assertEqual(qasmstr, strcorrect)
+
+    def test_pauli_exp_to_qasm_uses_rz_angle_for_exp_minus_i_theta_z(self):
+        qasmstr = "\n".join(pauli_exp_to_qasm([QubitOperator('Z0', 0.5)], evolution_time=3))
+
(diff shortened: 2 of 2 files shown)


marko1olo:
CI note: the non-CLA failure is from `Python compatibility tests`, not from the QASM `Rz` angle regression path changed in this PR.

The failing test is:

```text
src/openfermion/circuits/primitives/state_preparation_test.py::test_prepare_gaussian_state[4-True-None-2]
ValueError: matmul: dimension mismatch with signature (n,k=16),(k=8,1?)->(n,1?)
```

The log shows `cirq-core==1.4.1` in the max-compat job. The simulated state vector is length 8 while `quad_ham_sparse` is 16x16, so this looks like a compatibility/baseline test issue rather than a failure caused by the QASM export change.

src/openfermion/circuits/trotter_exp_to_qgates.py
arettig, line 246: As pointed out by the review bot, the controlled version is still incorrect. I believe the coefficients for the c-phase gate and the two Rz gates will also need to change.

arettig (changes requested):
Thanks for fixing this, the uncontrolled version looks good! There is also a controlled version of this exponential that is still incorrect - can you also correct the parameters on the cphase gate and the two Rz gates contained in the controlled branch of the if statement?

arettig:
superseded by #1380

Issue #1282, opened 2026-04-23
Mypy reports many errors, and they should be fixed


mhucka:
### Describe the issue

Unsetting the "ignore_errors" flag for mypy in the config file reveals there are many mypy type-checking errors. These should be fixed properly.

<details>

[31msrc/openfermion/ops/operators/symbolic_operator.py:123: error: Incompatible types in assignment (expression has type "None", base class "object" defined the type as "Callable[[], int]")  [assignment]
src/openfermion/resource_estimates/surface_code_compilation/physical_costing.py:86: error: Argument "failure_rate" to "MagicStateFactory" has incompatible type "float"; expected "int"  [arg-type]
src/openfermion/resource_estimates/surface_code_compilation/physical_costing.py:128: error: Argument "rounds" to "MagicStateFactory" has incompatible type "float"; expected "int"  [arg-type]
src/openfermion/resource_estimates/surface_code_compilation/physical_costing.py:129: error: Argument "failure_rate" to "MagicStateFactory" has incompatible type "float"; expected "int"  [arg-type]
src/openfermion/resource_estimates/pbc/thc/factorizations/kmeans.py:139: error: Incompatible types in assignment (expression has type "floating[Any]", variable has type "float")  [assignment]
src/openfermion/resource_estimates/pbc/hamiltonian/hamiltonian.py:115: error: Incompatible types in assignment (expression has type "type[complex128]", variable has type "type[floating[_64Bit]]")  [assignment]
src/openfermion/resource_estimates/pbc/hamiltonian/hamiltonian.py:116: error: Incompatible types in assignment (expression has type "dtype[Any]", variable has type "type[floating[_64Bit]]")  [assignment]
src/openfermion/resource_estimates/pbc/hamiltonian/hamiltonian.py:127: error: Need type annotation for "tao" (hint: "tao: list[<type>] = ...")  [var-annotated]
src/openfermion/ops/representations/polynomial_tensor.py:120: error: Incompatible types in assignment (expression has type "None", base class "object" defined the type as "Callable[[], int]")  [assignment]
src/openfermion/linalg/wave_fitting.py:71: error: Incompatible types in assignment (expression has type "tuple[Any, ...]", variable has type "ndarray[Any, Any]")  [assignment]
src/openfermion/contrib/representability/_namedtensor.py:46: error: Incompatible types in assignment (expression has type "None", variable has type "int")  [assignment]
[...]

PR #1349, opened 2026-06-08
Fix mypy annotations in fermion_partitioning (#1282)

rosspeili:
First slice for #1282: correct generator return types and local variable annotations in measurements/fermion_partitioning.py. No runtime or API behavior changes.

### DIFF
--- a/src/openfermion/measurements/fermion_partitioning.py
+++ b/src/openfermion/measurements/fermion_partitioning.py
@@ -10,13 +10,16 @@
 #   See the License for the specific language governing permissions and
 #   limitations under the License.
 
+from collections.abc import Generator, Iterable
+from typing import Any
+
 import numpy
 from openfermion.measurements import partition_iterator
 
 MAX_LOOPS = 1e6
 
 
-def pair_within(labels: list) -> list:
+def pair_within(labels: list[Any]) -> Generator[tuple[Any, ...], None, None]:
     """
     Generates pairings of labels that contain each pair at least once.
 
@@ -81,7 +84,9 @@ def pair_within(labels: list) -> list:
             yield pairing1 + pairing2
 
 
-def pair_between(frag1: list, frag2: list, start_offset: int = 0) -> tuple:
+def pair_between(
+    frag1: list[Any], frag2: list[Any], start_offset: int = 0
+) -> Generator[tuple[Any, ...], None, None]:
     """Pairs between two fragments of a larger list
 
     A pairing of a list is a set of pairs of list elements. E.g. a pairing of
@@ -197,7 +202,7 @@ def _gen_pairings_between_partitions(parta, partb):
                 yield pair_a + pair_b + pair_ab
 
 
-def pair_within_simultaneously(labels: list) -> tuple:
+def pair_within_simultaneously(labels: list[Any]) -> Generator[tuple[Any, ...], None, None]:
     """Generates simultaneous pairings between four-element combinations
 
     A pairing of a list is a set of pairs of list elements. E.g. a pairing of
@@ -230,7 +235,7 @@ def pair_within_simultaneously(labels: list) -> tuple:
     for partition in _gen_partitions(labels):
         generator_list = [_loop_iterator(pair_within, partition[j]) for j in range(len(partition))]
         for dummy1 in range(len(partition[-2]) - 1 + len(partition[-2]) % 2):
-            pairing = tuple()
+            pairing: tuple[Any, ...] = ()
             for generator in generator_list[::2]:
                 pairing = pairing + next(generator)[0]
             for dummy2 in range(len(partition[-1]) - 1 + len(partition[-1]) % 2):
@@ -277,7 +282,9 @@ def _get_padding(num_bins, bin_size):
         trial_size += 1
 
 
-def _asynchronous_iter(iterators, flatten=False):
+def _asynchronous_iter(
+    iterators: Iterable[Iterable[Any]], flatten: bool = False
+) -> Generator[tuple[Any, ...], None, None]:
     """
     Iterates over a set of K iterators with max L elements to
     generate all pairs between them in O(L^2 + 2L log(L) + log(L)^2),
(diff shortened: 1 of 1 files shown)


mhucka (changes requested):
Thank you for doing this!

There is one change I would recommend. Although `Any` is valid, it is better to define so-called generic parameter types, along the following lines:

```python
from typing import TypeVar

T = TypeVar('T')

def pair_within(labels: list[T]) -> Generator[tuple[T, ...], None, None]:
    """Generates pairings of labels that contain each pair at least once."""
    ...
```

Applying this across other matching functions (like `pair_between` and `pair_within_simultaneously`) allows static analyzers to check that the output tuples contain elements of the exact same type as the input.

Would you be able to try to make the appropriate changes to this PR?

rosspeili:
@mhucka reworked based on suggestion, mypy is clean on this file and all tests pass. Let me know how it looks, and thanks for the tip. <3

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -10,13 +10,22 @@
...
 
+T = TypeVar('T')
+
+# A yielded pairing is a tuple whose elements are usually (T, T) pairs.
+# Some code paths also append bare T values for unpaired leftovers.
+Pairing = tuple[tuple[T, T] | T, ...]
+
 
-def pair_within(labels: list) -> list:
+def pair_within(labels: list[T]) -> Generator[Pairing, None, None]:
```
mhucka: Here, the declaration should be changed to `Pairing[T]`.

```suggestion
def pair_within(labels: list[T]) -> Generator[Pairing[T], None, None]:
```

Using `Pairing` without its type argument causes static type checkers like mypy to treat it as `Pairing[Any]`, which often passes the type check, but does not give us the real benefits of the type declarations. The lack of the subscript `T` breaks the connection between the type `T` of the input list `labels` and the yielded element types.

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -197,7 +211,7 @@ def _gen_pairings_between_partitions(parta, partb):
                 yield pair_a + pair_b + pair_ab
 
 
-def pair_within_simultaneously(labels: list) -> tuple:
+def pair_within_simultaneously(labels: list[T]) -> Generator[Pairing, None, None]:
```
mhucka:
```suggestion
def pair_within_simultaneously(labels: list[T]) -> Generator[Pairing[T], None, None]:
```

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -367,7 +383,9 @@ def _parallel_iter(iterators, flatten=False):
             yield tuple(next_result)
 
 
-def pair_within_simultaneously_binned(binned_majoranas: list) -> tuple:
+def pair_within_simultaneously_binned(
+    binned_majoranas: list[list[T]],
+) -> Generator[Pairing, None, None]:
```
mhucka:
```suggestion
) -> Generator[Pairing[T], None, None]:
```

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -426,7 +444,9 @@ def pair_within_simultaneously_binned(binned_majoranas: list) -> tuple:
             yield pairing
 
 
-def pair_within_simultaneously_symmetric(num_fermions: int, num_symmetries: int) -> tuple:
+def pair_within_simultaneously_symmetric(
+    num_fermions: int, num_symmetries: int
+) -> Generator[Pairing, None, None]:
```
mhucka: This one is not a `T` but instead, an `int`:

```suggestion
) -> Generator[Pairing[int], None, None]:
```

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -230,7 +244,7 @@ def pair_within_simultaneously(labels: list) -> tuple:
     for partition in _gen_partitions(labels):
         generator_list = [_loop_iterator(pair_within, partition[j]) for j in range(len(partition))]
         for dummy1 in range(len(partition[-2]) - 1 + len(partition[-2]) % 2):
-            pairing = tuple()
+            pairing: Pairing = ()
```
mhucka:
```suggestion
            pairing: Pairing[T] = ()
```

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -115,6 +128,7 @@ def pair_between(frag1: list, frag2: list, start_offset: int = 0) -> tuple:
     num_pairs = min(len(frag1), len(frag2))
 
     for index_offset in range(start_offset, num_iter):
+        pairing: Pairing
```
mhucka:
```suggestion
        pairing: Pairing[T]
```

src/openfermion/measurements/fermion_partitioning.py
```diff
@@ -81,7 +92,9 @@ def pair_within(labels: list) -> list:
             yield pairing1 + pairing2
 
 
-def pair_between(frag1: list, frag2: list, start_offset: int = 0) -> tuple:
+def pair_between(
+    frag1: list[T], frag2: list[T], start_offset: int = 0
+) -> Generator[Pairing, None, None]:
```
mhucka:
```suggestion
) -> Generator[Pairing[T], None, None]:
```

mhucka (changes requested):
I'm not done with the review on the update, but here's a start for a minor detail.

rosspeili:
@mhucka applied all the `Pairing[T]` suggestions from above, but that made strict mypy complain about the `None` sentinel in `pair_within`, so I fixed that in a follow-up push, same behavior, all tests pass. Let me know how this looks or if I am missing something. Thanks again for the tips and direction <3

mhucka (approved):
Looks good. Thank you for working on this so diligently!

rosspeili:
@mhucka thanks for merging #1349 

Happy to keep going with the same small-PR approach. Any preference for the next batch? I was thinking another pure-annotation file with a small error count, unless you'd rather prioritize a specific error family first.

mhucka:
@rosspeili Thank you for your work on this. I can't think of a reason to prefer any particular order, so it's up to you.

Issue #1340, opened 2026-06-04
Raise the version of Cirq to the latest if possible


mhucka:
### Is your feature request related to a use case or problem?

The current requirements files end up pinning Cirq to 1.4.1, which is increasingly problematic. We should try to support higher versions of Cirq.

### What solution or approach do you envision?

A starting point would be to remove all the version pins in dev_tools/requirements/deps, run pip-compile, and try to resolve any problems that arise.

### How urgent is this for you?

P1 – needed by the next release

rosspeili:
Hey @mhucka happy to take a dive into this.

mhucka:
> @rosspeili happy to take a dive into this.

Thank you for offering. I had actually assigned it to myself last week, and a PR merged yesterday (https://github.com/quantumlib/OpenFermion/pull/1352) should hopefully resolve this. I see I failed to reference this issue there and close this. Sorry if that was misleading …

rosspeili:
Nw at all @mhucka. <3

PR #1352, opened 2026-06-09
Replace `pip-compile`-based dependency management with `uv`-based approach

mhucka:
Replace the Docker file and scripts in dev_tools/requirements with a simpler scheme that uses `uv`. Among other things, `uv`'s smarter method for producing lock files makes it possible to deal with Cirq version interactions with Python's versions, so that users can use the latest Cirq version if they are also using recent Python versions.


dev_tools/requirements/create-env-files.sh
```diff
@@ -0,0 +1,107 @@
...
+Options:
+  -h   Show this help message and exit
+
+All other options on the command line will be passed directly to
+'uv pip compile'. Run 'uv pip compile --help' to learn about the
+options available."
+
+while [[ $# -gt 0 ]]; do
+    case "${1}" in
+        -h) echo "${usage}"; exit 0 ;;
```
pavoljuhas: Nit - can we handle `--help` too for a good measure?

dev_tools/requirements/create-env-files.sh
```diff
@@ -0,0 +1,107 @@
...
+
+uv pip compile "$@" \
+    -o dev_tools/requirements/envs/shellcheck.env.txt \
+    -c dev_tools/requirements/envs/dev.env.txt \
+    dev_tools/requirements/deps/shellcheck.txt
+
+# ~~~~ Generate max_compat files ~~~~
+
+uv pip compile "$@" \
+    -o dev_tools/requirements/max_compat/dev.env.txt \
```
pavoljuhas: Is this still needed?  The output in `dev_tools/requirements/max_compat/dev.env.txt` is nearly identical to `dev_tools/requirements/envs/pytest.env.txt`.  It appears to be compiled for Python >= 3.12 so the constraints in oldest-versions.txt have no effect.

dev_tools/requirements/create-env-files.sh
```diff
@@ -0,0 +1,107 @@
...
+    -c dev_tools/requirements/deps/oldest-versions.txt \
+    dev_tools/requirements/deps/pytest.txt \
+    dev_tools/requirements/deps/runtime.txt
+
+uv pip compile "$@" \
+    -o dev_tools/requirements/max_compat/pytest-max-compat.env.txt \
```
pavoljuhas, line 98: Looks like a duplicate of `dev_tools/requirements/max_compat/dev.env.txt`.
Please fix or delete.

dev_tools/requirements/deps/oldest-versions.txt
```diff
@@ -1,3 +1,3 @@
 # Used as a constraint file for pytest-max-compat
-cirq-core~=1.4.1
-cirq-google~=1.4.1
+cirq-core ~= 1.4.1; python_version < "3.11"
+cirq-google ~= 1.4.1; python_version < "3.11"
```
pavoljuhas: I don't see any reference to `cirq_google` in the repo, please delete if redundant.

dev_tools/requirements/deps/runtime.txt
```diff
@@ -1,21 +1,10 @@
-cirq-core>=1.4.1,<1.6.0
+cirq-core >= 1.4.1, < 1.6.0; python_version < "3.11"
+cirq-core >= 1.6.0; python_version >= "3.11"
```
pavoljuhas, line 2: I am a bit puzzled why is this needed.  cirq-core carries a `python_requires` metadata so if installing to 3.10, pip should find the latest version 1.5.0 (and later versions for later Pythons).

pavoljuhas (changes requested):
Please see inline comments regarding max_compat files, otherwise LGTM.

Also, if all `.env.txt` files will be only generated by the `uv`, consider marking them up with the `linguist-generated` git attribute like [here](https://github.com/quantumlib/Cirq/blob/6922063c70b2ef6d1a13bc39a0921185cebfffeb/.gitattributes#L6-L7) (can be a separate PR).

dev_tools/requirements/create-env-files.sh
mhucka: Done in the next push.

dev_tools/requirements/deps/oldest-versions.txt
mhucka: Good catch. Must have been some historical hold-over.

mhucka:
> Also, if all `.env.txt` files will be only generated by the `uv`, consider marking them up with the `linguist-generated` git attribute like [here](https://github.com/quantumlib/Cirq/blob/6922063c70b2ef6d1a13bc39a0921185cebfffeb/.gitattributes#L6-L7) (can be a separate PR).

Thanks for the great idea. Done in PR #1355

dev_tools/requirements/deps/runtime.txt
mhucka: IIRC, this is related to some transitive dependency (possibly contourpy, brought in by matplotlib), where some pip is unable to find a compatible version on Python 3.10 if one tries to install Cirq 1.5 but it succeeds with 1.4.1.

dev_tools/requirements/create-env-files.sh
mhucka: > Is this still needed? The output in `dev_tools/requirements/max_compat/dev.env.txt` is nearly identical to `dev_tools/requirements/envs/pytest.env.txt`. It appears to be compiled for Python >= 3.12 so the constraints in oldest-versions.txt have no effect.

They look pretty different to me. Can you double-check the files again?

With respect to the versions, the oldest-version file is needed to provide the condition on Python 3.11. In both of the files above, the following can be found:

```
cirq-core==1.4.1 ; python_full_version < '3.11' \
    --hash=sha256:869db60413265c41a8206854c1d4ca9bad5fac9cfd7c6a10685b5a6d516defa0
    # via
    #   -c dev_tools/requirements/deps/oldest-versions.txt
    #   -r dev_tools/requirements/deps/runtime.txt
cirq-core==1.6.1 ; python_full_version >= '3.11' \
    --hash=sha256:fbc809d7c6a228762ad92bb7870dc16f55b76728fcc7ef4c7aaceb7e5c63e8e6
```

dev_tools/requirements/create-env-files.sh
mhucka: > Looks like a duplicate of `dev_tools/requirements/max_compat/dev.env.txt`. Please fix or delete.

You're right. Fixed in the next push.

dev_tools/requirements/create-env-files.sh
pavoljuhas: NVM - I can see the difference in the cirq-core requirement.  Sorry about that.

dev_tools/requirements/create-env-files.sh
```diff
@@ -0,0 +1,102 @@
...
+    -o dev_tools/requirements/envs/shellcheck.env.txt \
+    -c dev_tools/requirements/envs/dev.env.txt \
+    dev_tools/requirements/deps/shellcheck.txt
+
+# ~~~~ Generate max_compat files ~~~~
+
+uv pip compile "$@" \
+    -o dev_tools/requirements/max_compat/pytest-max-compat.env.txt \
+    -c dev_tools/requirements/deps/oldest-versions.txt \
+    -c dev_tools/requirements/max_compat/dev.env.txt \
```
pavoljuhas: This does not exist anymore.

```suggestion
```

pavoljuhas (approved):
LGTM, please see one last suggestion before merging.

mhucka:
Resolved by #1340

PR #1354, opened 2026-06-09
Fix planewave based jellium model

### LINKED ISSUE #808: Is the Jellium model Hamiltonian incorrect?
I have a quick question related to the correctness of the Jellium (plane wave basis) Hamiltonian - in particular the potential term in 2 dimensions. 

A minimal code snippet to generate the potential term of a spinless Jellium Hamiltonian (with the plane wave basis states living on a 2D, 3x3 grid) reads:

```
from openfermion.hamiltonians.jellium import plane_wave_potential
from openfermion import Grid
import math
grid = Grid(dimensions=2, length=3, scale=2*math.pi)
potential = plane_wave_potential(grid, spinless=True, e_cutoff = None)
```

There are two potential issues that I see (and I'm curious if I'm missing something maybe?):

1) The code tries to implement the second line of Eq 7 in [1]. Looking at `plane_wave_potential`, the outer loop over \mu (or `omega_indices` in the code defined on line ~146 of `OpenFermion/src/openfermion/hamiltonians/jellium.py`), will iterate over momenta values ([0,0] is skipped): 
[-1. -1.], [-1.  0.], [-1.  1.], [ 0. -1.], [0. 1.], [ 1. -1.], [1. 0.], [1. 1.]
This seems potentially incomplete to me. I would think there should be other momenta values here like [-2, 0], for example. This is because (see Eq 7 in [1]), orbitals that correspond to index, (for example): \mu + p = [-2, 0] + [1,0] = [-1, 0] are still on the grid of allowable basis states. In other words, the \mu index (in code, effectively related to `omega_indices`) should include cases corresponding to higher momenta values than what the base 3x3 grid includes. 

2) The authors of [1] write eq 7 for a 3d case. The code, however, seems to be written in a general way, and allows other dimensions (say dimension==2). However, `plane_wave_potential` scales each term of the potential energy by a 1/momenta_squared (see line ~162 of `OpenFermion/src/openfermion/hamiltonians/jellium.py`) factor, where momenta_squared is calculated from the outer index (i.e index \mu). This seems incorrect for dimensions other than 3. For example, book [2] (see equations 1.19 and 1.68) shows that this scaling should go like 1/momenta (no square) in 2 dimensions. 

I'd love to know if I'm missing something or if these are issues with the current Jellium implementation in the code. 
thanks

[1] Babbush et al, PRX 8, 011044 (2018)
[2] Giuliani, G. & Vignale, G. Quantum Theory of the Electron Liquid

arettig:
This fixes #808, accounting for the 2 issues raised there:

1. The planewave basis set is truncated inconsistently. The planewave basis set is truncated based on Ecut as expected for the kinetic term, but in the electron-electron repulsion, only the momentum change q = G1 - G2 is truncated, allowing terms arising from planewaves that should be excluded from the basis. This has been corrected by truncating consistently with the kinetic term.
2. The coulomb operator was hardcoded for 3-dimensional systems but allowed calculations with other numbers of dimensions (and in fact was tested on 1-dimensional systems). The coulomb kernel was generalized to allow for 1 and 2 dimensional systems as well (using a soft coulomb potential for 1D), and will crash if not using 1-3 dimensions. The reference values for several tests that utilized 1-dimensional hamiltonians were adjusted.

### DIFF
--- a/src/openfermion/hamiltonians/jellium.py
+++ b/src/openfermion/hamiltonians/jellium.py
@@ -15,6 +15,7 @@ import math
 from typing import Optional
 
 import numpy
+import scipy.special
 
 from openfermion.ops.operators import FermionOperator, QubitOperator
 from openfermion.utils.grid import Grid
@@ -59,6 +60,37 @@ def wigner_seitz_length_scale(
     length_scale = volume ** (1.0 / dimension)
 
     return length_scale
+
+
+def coulomb_potential_momentum(
+    momenta_squared: float, dimension: int, volume: float, a_1d: float = 1.0
+) -> float:
+    """Return the momentum space Coulomb potential for a given dimension.
+
+    For 1d systems, a soft coulomb potential is used with a regularization
+    parameter of a_1d: v(r) = 1/sqrt(r^2 + a_1d^2)
+
+    Args:
+        momenta_squared (float): The squared momentum vector.
+        dimension (int): The dimension of the system (1, 2, or 3).
+        volume (float): The volume (or area, or length) of the simulation cell.
+        a_1d (float): The regularization parameter for 1D systems (default 1.0).
+
+    Returns:
+        float: The potential coefficient.
+    """
+    if momenta_squared == 0:
+        return 0.0
+    q = numpy.sqrt(momenta_squared)
+    if dimension == 3:
+        return 2.0 * numpy.pi / (volume * momenta_squared)
+    elif dimension == 2:
+        return numpy.pi / (volume * q)
+    elif dimension == 1:
+        # use a soft coulomb potential with reg. param a_1d
+        return scipy.special.k0(q * a_1d) / volume
+    else:
+        raise ValueError(f'Unsupported dimension {dimension}.')
 
 
 def plane_wave_kinetic(
@@ -119,7 +151,6 @@ def plane_wave_potential(
         operator (FermionOperator)
     """
     # Initialize.
-    prefactor = 2.0 * numpy.pi / grid.volume_scale()
     operator = FermionOperator((), 0.0)
     spins = [None] if spinless else [0, 1]
     if non_periodic and period_cutoff is None:
@@ -169,7 +200,9 @@ def plane_wave_potential(
             continue
 
         # Compute coefficient.
-        coefficient = prefactor / momenta_squared
+        coefficient = coulomb_potential_momentum(
+            momenta_squared, grid.dimensions, grid.volume_scale()
+        )
         if non_periodic:
             coefficient *= 1.0 - numpy.cos(period_cutoff * numpy.sqrt(momenta_squared))
 
@@ -228,7 +261,6 @@ def dual_basis_jellium_model(
     """
     # Initialize.
     n_points = grid.num_points
-    position_prefactor = 2.0 * numpy.pi / grid.volume_scale()
     operator = FermionOperator()
     spins = [None] if spinless else [0, 1]
     if potential and non_periodic and period_cutoff is None:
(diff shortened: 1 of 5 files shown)


src/openfermion/hamiltonians/jellium.py
```diff
@@ -61,6 +62,37 @@ def wigner_seitz_length_scale(
...
+        dimension (int): The dimension of the system (1, 2, or 3).
+        volume (float): The volume (or area, or length) of the simulation cell.
+        a_1d (float): The regularization parameter for 1D systems (default 1.0).
+
+    Returns:
+        float: The potential coefficient.
+    """
+    if momenta_squared == 0:
+        return 0.0
+    q = numpy.sqrt(momenta_squared)
```
mhucka, line 84: Since `momenta_squared` is a Python `float`, if the value is a non-negative scalar, it should be possible to use Python's regular `math.sqrt`; this would be slightly faster as it will avoid some overhead in NumPy's version.

If this is changed to a `math.sqrt`, it also would be better to test that the value of momenta_squared is non-negative before taking the square root. The rest of this function tests other values and uses an `else` to deal with everything else, but it will be too late by then.

Finally, I see that the original code in this file used numpy functions and constants instead of Python `math` versions. I'm not sure if there was a reason for that – it might be a historical hold-over from pre-Python 3 days, or something else. If only the new code in this PR is changed to use the math versions, then that would be a bit inconsistent. So, an option might be to leave everything in this PR as you have it now (using numpy things), and leave the type changes to a different PR.

src/openfermion/hamiltonians/jellium.py
```diff
@@ -61,6 +62,37 @@ def wigner_seitz_length_scale(
...
+    if dimension == 3:
+        return 2.0 * numpy.pi / (volume * momenta_squared)
```
mhucka, line 86: Similar as above, it should be possible to use Python's regular `math.pi` here; this would be slightly faster as it will avoid some overhead in NumPy's version.

src/openfermion/hamiltonians/jellium.py
```diff
@@ -61,6 +62,37 @@ def wigner_seitz_length_scale(
...
+    elif dimension == 2:
+        return numpy.pi / (volume * q)
```
mhucka, line 88: Ditto for the `numpy.pi` here.

src/openfermion/hamiltonians/jellium.py
```diff
@@ -61,6 +62,37 @@ def wigner_seitz_length_scale(
     return length_scale
 
 
+def coulomb_potential_momentum(
+    momenta_squared: float, dimension: int, volume: float, a_1d: float = 1.0
+) -> float:
+    """Return the momentum space Coulomb potential for a given dimension.
+
+    For 1d systems, a soft coulomb potential is used with a regularization
+    parameter of a_1d: v(r) = 1/sqrt(r^2 + a_1d^2)
```
mhucka: We should probably format this equation using LaTeX notation. This will render better when the docstrings are processed for the web documentation pages.

src/openfermion/hamiltonians/jellium.py
```diff
@@ -61,6 +62,37 @@ def wigner_seitz_length_scale(
     return length_scale
 
 
+def coulomb_potential_momentum(
+    momenta_squared: float, dimension: int, volume: float, a_1d: float = 1.0
+) -> float:
+    """Return the momentum space Coulomb potential for a given dimension.
+
+    For 1d systems, a soft coulomb potential is used with a regularization
```
mhucka: Nit:

```suggestion
    For 1-D systems, a soft coulomb potential is used with a regularization
```

mhucka (changes requested):
Thank you for doing this!

I have some comments about some possible slightly optimizations, but they might be better off left to another PR.

The only other thing is a request for formatting the formula in one of the docstrings. Other than that, it looks good!

src/openfermion/hamiltonians/jellium.py
arettig: This is a good point, thanks! I will make another PR to replace the NumPy functions.

src/openfermion/hamiltonians/jellium.py
arettig: Done, thanks for catching this!

src/openfermion/hamiltonians/jellium.py
arettig: Done, thanks!

mhucka (approved):
LGTM. Thanks!

PR #1355, opened 2026-06-10
Add `.gitattributes` file

mhucka:
This adds a `.gitattributes` file to tell GitHub's Linguist tool how to treat certain special cases.


.gitattributes
```diff
@@ -0,0 +1,25 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Mark files generated by uv to be collapsed by default in PRs.
+# See: https://github.com/github/linguist/#generated-code
+dev_tools/requirements/envs/*.txt linguist-generated=true
```
pavoljuhas: This does not catch files in dev_tools/requirements/max_compat/

```suggestion
*.env.txt linguist-generated=true
```

Verification: empty output from the command below, i.e., every file has the _autogenerated_ pattern.

```shell
git ls-files "*.env.txt" | xargs grep -L This.file.is.autogenerated
```

.gitattributes
```diff
@@ -0,0 +1,25 @@
...
+
+# Include Markdown files in GitHub's file statistics for this repo.
+*.md linguist-detectable
+
+# Exclude config files from GitHub's file statistics. Most are YAML files and
+# Linguist already treats .yaml files as data, so only add some exceptions.
+.markdownlintrc linguist-documentation
+.shfmt          linguist-documentation
```
pavoljuhas: Let us remove patterns that do not exist:

```suggestion
```

pavoljuhas (changes requested):
Please see inline, otherwise LGTM.

pavoljuhas (approved):

PR #1356, opened 2026-06-10
Fix integer type in `binary_code_transform`

JdelArco98:
on openfermion/transforms/opconversions/binary_code_transform.py line 163 it breaks (with numpy>2.) because (-1) ** updated_parity being updated_parity numpy.count_nonzero(fermionic_indices[:op_idx] < op_tuple[0]) line 146 (therefore a numpy.int) is not instance of int, as it is checked on openfermion/ops/operators/symbolic_operator.py  line 124


mhucka (approved):
Thank you for catching and fixing this!

PR #1358, opened 2026-06-11
Remove use of NumPy for scalar math in jellium.py

arettig:
Previously, NumPy functions were used for scalar math in `jellium.py`. Using the Python standard math library instead should yield slight speedups.

### DIFF
--- a/src/openfermion/hamiltonians/jellium.py
+++ b/src/openfermion/hamiltonians/jellium.py
@@ -14,7 +14,6 @@
 import math
 from typing import Optional
 
-import numpy
 import scipy.special
 
 from openfermion.ops.operators import FermionOperator, QubitOperator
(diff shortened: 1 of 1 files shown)


mhucka (commented):
@arettig This looks good, with one possible issue. The `math.cos` function may given an error if `momenta.dot()` returns an array (even a 1-D array). 

```python
math.cos(momenta.dot(differences))
```

(Credit goes to Gemini for flagging this.) I didn't trace the code carefully to see if that can actually happen, but if there's a chance of that, maybe cast the dot product output explicitly to a scalar `float` before passing it to `math.cos`:

```python
math.cos(float(momenta.dot(differences)))
```

Otherwise LGTM.

arettig:
> This looks good, with one possible issue. The `math.cos` function may given an error if `momenta.dot()` returns an array (even a 1-D array).
> 
> ```python
> math.cos(momenta.dot(differences))
> ```
> 
> (Credit goes to Gemini for flagging this.) I didn't trace the code carefully to see if that can actually happen, but if there's a chance of that, maybe cast the dot product output explicitly to a scalar `float` before passing it to `math.cos`:
> 
> ```python
> math.cos(float(momenta.dot(differences)))
> ```
> 
> Otherwise LGTM.

It should be impossible to get a non-scalar from `momenta.dot()` since `Grid.momentum_vector()` and `Grid.position_vector()` return 1-D vectors, so I am inclined to leave this as is.

mhucka (approved):
LGTM

PR #1359, opened 2026-06-12
Add missing term to trotter error loose bounds

### LINKED ISSUE #819: Incorrect Bounds on Trotter Error
The 'loose' and possibly 'tight' bounds calculated by `openfermion.circuits.error_bound` are incorrect. Evidence:

- The 'loose' bound is numerically smaller than the 'tight' bound for simple 2D Ising models, which can be easily verified. See image below.
- For a two-summand Hamiltonian $H=A+B$, the loose bound is essentially set to $\mathcal E \le ||[A,[B,A] ||$. The correct bound is $\mathcal E_{2} \le \frac{1}{12} (||[A,[B,A] || + \frac{1}{2}||[B,[B,A]|| )$. In the case of the 2D Ising model, the second term is combinatorially large in the number of lattice edges and quickly overwhelms the first term, resulting in a bound much larger than the provided loose bound.
- For my specific problem, I derived a formula for the value of the second order error by hand. The 'loose' bound is always equal to the term |[A[B,A]]|, which is incorrect as mentioned. The 'tight' bound seems to overshoot my analytical expression, which is acceptable, but indicates that there might be some room for simple improvements here.

Tight on time, I'm going to commit a faux-paux here and post a screenshot rather than minimal code, and I'll come back and add minimal code to reproduce after my sprint if anyone needs it :).

image

arettig:
This fixes #819. The loose error bound for trotter decomposition is currently implemented as equation 16 from [this paper](https://arxiv.org/pdf/1406.4920):

$$\Delta E_{\text{loose}} =  4 \sum_{\alpha}\lVert H_{\alpha} \rVert  \bigg(\sum_{\beta}' \lVert H_{\beta} \rVert\bigg)^2 \Delta t^2$$

In the paper they mention this is only an upper bound on the second term of equation 6, but we should be giving an upper bound of both terms. I've changed the loose error bound to account for both terms:

$$\Delta E_{\text{loose}} =4 \sum_{\alpha} \lVert H_{\alpha} \rVert \left( \sum_{\beta}' \lVert H_{\beta} \rVert \right)^2 \Delta t^2 + 4 \sum_{\alpha} \lVert H_{\alpha} \rVert^2\sum_{\beta}' \lVert H_{\beta} \rVert \Delta t^2$$

The loose error bound unit test is also changed accordingly.


src/openfermion/circuits/trotter/trotter_error.py
```diff
@@ -154,6 +171,7 @@ def error_bound(terms, tight=False):
                     ):
                         error_a += abs(coefficient_b)
 
+                error += 4.0 * abs(coefficient_a) ** 2 * error_a
                 error += 4.0 * abs(coefficient_a) * error_a**2
```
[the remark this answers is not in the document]
mhucka: Wow, GCA actually read the paper? 😮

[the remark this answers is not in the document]
arettig: > Wow, GCA actually read the paper? 😮

Not very well I'm afraid 😔. The factor of 1/12 is in equation 9 for the tight bound. There is no 1/12 in the loose bound. 

I believe the equation is correct as implemented.

src/openfermion/circuits/trotter/trotter_error_test.py
```diff
@@ -132,7 +132,7 @@ def test_error_bound_xyz_tight(self):
 
     def test_error_bound_xyz_loose(self):
         terms = [QubitOperator('X1'), QubitOperator('Y1'), QubitOperator('Z1')]
-        self.assertTrue(numpy.isclose(error_bound(terms, tight=False), 4.0 * (2**2 + 1**2)))
+        self.assertTrue(numpy.isclose(error_bound(terms, tight=False), 4.0 * (2**2 + 1**2 + 2 + 1)))
```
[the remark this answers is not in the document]
arettig: as stated above, I believe this is incorrect.

src/openfermion/circuits/trotter/trotter_error.py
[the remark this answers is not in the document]
mhucka: Oh, LOL.

mhucka: Are both lines supposed to be kept? If yes, the above could be made slightly more efficient:

* `4 * abs(coefficient_a)` is done twice – it'd be faster to compute once, save in a variable, & reuse it
* In Python, it turns out that `x * x` is faster than `x ** 2` because the latter invokes a generic exponentiation function
* Since the second assignment to `error` doesn't depend on the first, the two lines look like they could be combined into 1 line

src/openfermion/circuits/trotter/trotter_error.py
```diff
@@ -154,6 +171,7 @@ def error_bound(terms, tight=False):
                     ):
                         error_a += abs(coefficient_b)
```
mhucka, line 172: `coefficient_b` is not used between here and when it's assigned above, which means the `(coefficient_b,) = term_b.terms.values()` could be moved inside the `if` block. That'll skip a needless assignment in cases where `coefficient_b` would not get used.

mhucka, line 172: Actually, upon further examination, it looks like the use of a doubly-nested loop is not necessary. The value computed by the inner loop does not depend on `coefficient_a`, which implies `error_a` could be computed in a separate pass, turning this _O(N²)_ algorithm into _O(N)_.

mhucka (changes requested):
It looks like there may be some opportunities for efficiency improvements.

src/openfermion/circuits/trotter/trotter_error.py
arettig: Good point! I combined these two lines into one line, cached abs(coefficient_a) in a variable, and removed the `**2`. The number of arithmetic operations here should be reduced.

src/openfermion/circuits/trotter/trotter_error.py
arettig: Great catch, done!

arettig: I think we have to keep the nested loop here unfortunately. The commutation check in the inner loop depends on both alpha and beta.

mhucka (approved):
Sounds good.

mhucka:
@arettig I'll leave it to you to merge this when you're ready.

PR #1360, opened 2026-06-13
Fix bogoliubov_transform for spin-preserving Gaussian transformations

### LINKED ISSUE #776: The implementation of bogoliubov for a general fermionic gaussian that doesn't mix spin up and spin down is broken
When bogoliubov_transform is called with a transformation_matrix that has shape N x 2N the check that determines whether or not the transformation mixes spin up and spin down and the subsequent code is broken. The current check (_is_spin_block_diagonal) and some of the subsequent code is only valid for a square transformation_matrix.

To reproduce:

Call bogoliubov_transform with the following arguments:
qubits = [cirq.LineQubit(0), cirq.LineQubit(1)]
trasformation_matrix = np.ndarray([[-9.57167901e-01-2.89533435e-01j,  0.00000000e+00+0.00000000e+00j,
        -3.33066907e-16-5.55111512e-17j,  0.00000000e+00+0.00000000e+00j],
       [ 0.00000000e+00+0.00000000e+00j,  1.00000000e+00+0.00000000e+00j,
         0.00000000e+00+0.00000000e+00j,  0.00000000e+00+0.00000000e+00j]])

blazingphoenix7:
Fixes #776.

`bogoliubov_transform` checks whether the transformation matrix is spin-block-diagonal before it ever looks at the matrix shape, and that check only makes sense for a square N x N matrix. For the N x 2N matrix you get out of a general (non-particle-conserving) Bogoliubov transformation it ends up inspecting the wrong blocks, decides that a transform which doesn't actually mix spin is "block diagonal", and then slices it into pieces of the wrong shape. The matrix from the issue dies with:

```
ValueError: Bad shape for transformation_matrix. Expected (1, 1) or (1, 2) but got (1, 3).
```

### What I changed

I moved the block detection into a helper, `_spin_blocks`, that handles both shapes. For the N x 2N case, write the matrix as `W = (W1 W2)` (it acts on all the creation operators followed by all the annihilation operators). The transform leaves the two spin sectors alone exactly when both `W1` and `W2` are block diagonal, and in that case each sector is its own (N/2) x N Gaussian transform built from the matching diagonal blocks of `W1` and `W2`. The recursion that was already there handles the rest. As a bonus, for spin-symmetric systems (BCS-style pairing Hamiltonians, say) you now do two half-size Givens decompositions instead of one full-size one.

### The part that took some digging

I didn't want to call this correct based on eigenstate energies alone, so I checked the circuit at the operator level: does it actually give `U a_p† U⁻¹ = b_p†` for every mode? That caught something the energy checks can't.

Under Jordan-Wigner a spin-down ladder operator drags a Z string across all of the spin-up qubits, so it anticommutes with the spin-up parity operator. If the spin-up circuit flips spin-up parity (which happens whenever its Gaussian decomposition uses an odd number of particle-hole transforms), then splitting the sectors apart flips the sign of every spin-down operator. What makes it sneaky is that it's invisible to energy tests: starting from a fixed occupation state, that bad sign is just a global phase, so the state still comes out as the correct eigenstate with the correct energy even though the underlying unitary is wrong.

So when the spin-up circuit flips parity I add a layer of Z gates on the spin-down qubits to put the signs back (`_preserves_parity` decides whether that's needed). With a fixed computational-basis initial state it's only a global phase, so I skip it there.

I also ran into the general Gaussian decomposition not reliably diagonalizing exactly-block-structured matrices (`U† H U` comes out non-diagonal for some perfectly valid block inputs), so I went with always splitting these instead of trying to detect when it's safe to fall back. Splitting is the more correct path here, not just the faster one.

### Tests

- `test_bogoliubov_transform_spin_block_gaussian_regression`: the actual matrix from #776, normalized so it satisfies `W1 W1† + W2 W2† = I`. This used to raise ValueError.
- `test_bogoliubov_transform_spin_block_operator_algebra`: checks `U a_p† U⁻¹ = b_p†` on the full unitary, for both a parity-preserving and a parity-flipping spin-up sector. This is the one that fails without the sign fix.
- `test_bogoliubov_transform_spin_block_gaussian`: eigenstate energies through the state-prep path.
- `test_bogoliubov_transform_spin_mixing_gaussian_not_split`: a transform that genuinely mixes spin should still go through the general path rather than getting split.

Outside the committed tests I ran a wider sweep (k = 2, 3, 4, real and complex, a range of seeds) cross-checking three things that each catch a sign error on their own: the operator algebra, whether `U† H U` is diagonal, and the eigenstate energies. Everything passes, including 27 of the parity-flipping cases. The full `circuits/` suite passes and black/pylint are clean.

### DIFF
--- a/src/openfermion/circuits/primitives/bogoliubov_transform.py
+++ b/src/openfermion/circuits/primitives/bogoliubov_transform.py
@@ -92,13 +92,14 @@ def bogoliubov_transform(
         initial_state = _occupied_orbitals(initial_state, n_qubits)
     initially_occupied_orbitals = cast(Optional[Sequence[int]], initial_state)
 
-    # If the transformation matrix is block diagonal with two blocks,
-    # do each block separately
-    if _is_spin_block_diagonal(transformation_matrix):
-        up_block = transformation_matrix[: n_qubits // 2, : n_qubits // 2]
+    # If the transformation matrix does not mix the two spin sectors, do each
+    # sector separately. This is both faster (two half-size decompositions) and
+    # more robust: the general fermionic Gaussian decomposition is numerically
+    # unreliable on exactly block-structured matrices. See issue #776.
+    spin_blocks = _spin_blocks(transformation_matrix)
+    if spin_blocks is not None:
+        up_block, down_block = spin_blocks
         up_qubits = qubits[: n_qubits // 2]
-
-        down_block = transformation_matrix[n_qubits // 2 :, n_qubits // 2 :]
         down_qubits = qubits[n_qubits // 2 :]
 
         if initially_occupied_orbitals is None:
@@ -109,6 +110,17 @@ def bogoliubov_transform(
             down_orbitals = [
                 i - n_qubits // 2 for i in initially_occupied_orbitals if i >= n_qubits // 2
             ]
+
+        # Spin-down ladder operators carry a Jordan-Wigner Z string over all
+        # spin-up qubits, so they anticommute with the spin-up parity operator.
+        # If the spin-up circuit flips spin-up parity (an odd number of
+        # particle-hole transformations) the factorized circuit would realize
+        # the spin-down operators with a flipped sign. A layer of Z gates on the
+        # spin-down qubits restores the documented action U a_p^dagger U^-1 =
+        # b_p^dagger. When an initial computational basis state is fixed this
+        # only contributes an irrelevant global phase, so it is skipped there.
+        if initially_occupied_orbitals is None and not _preserves_parity(up_block):
+            yield (cirq.Z(q) for q in down_qubits)
 
         yield bogoliubov_transform(up_qubits, up_block, initial_state=up_orbitals)
         yield bogoliubov_transform(down_qubits, down_block, initial_state=down_orbitals)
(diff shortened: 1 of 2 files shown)


blazingphoenix7:
@googlebot I signed it!

mhucka (approved):
Thank you for this work! Looks good to me. I requested Adam to review it too because this is a bit out of my depth.

src/openfermion/circuits/primitives/bogoliubov_transform.py
```diff
@@ -131,6 +175,28 @@ def _is_spin_block_diagonal(matrix) -> bool:
...
+    """
+    n, m = transformation_matrix.shape
+    if m == n:
+        return True
+    # Rearrange exactly as _gaussian_basis_change does, so the particle-hole
+    # count matches the circuit that will actually be generated.
+    left_block = transformation_matrix[:, :n]
+    right_block = transformation_matrix[:, n:]
+    rearranged = numpy.block([numpy.conjugate(right_block), numpy.conjugate(left_block)])
+    decomposition, _, _, _ = linalg.fermionic_gaussian_decomposition(rearranged)
```
arettig: It would be nice to avoid running the entire decomposition just to check the parity if possible. I think there should be a straightforwardish way to check the parity of the transform just based on the `transformation_matrix` (maybe based on the singular values of the `right_block` or maybe the determinant?) that would speed this up a lot.

arettig (changes requested):
Looks good and great catch on the parity flipping! Your tests look thorough as well.

I have one potential concern about the performance of the code - namely that the entire decomposition is run just to check the parity. I think there should be a way to do this more efficiently - I left a couple thoughts below. Let me know if it's not as straightforward as I'm thinking...

src/openfermion/circuits/primitives/bogoliubov_transform.py
blazingphoenix7: Good idea, done in abd17da. The parity is the determinant of the orthogonal Majorana transform, which is just the determinant of the unitary Bogoliubov-de Gennes extension `[[W1, W2], [W2*, W1*]]` of the transformation matrix (it comes out real and equal to +/-1). So I compute that determinant directly now instead of running the decomposition. I confirmed it matches the old decomposition-based result on 200 random spin-block transforms (even and odd parity, real and complex, sizes 2 through 6).

arettig (approved):
LGTM

PR #1361, opened 2026-06-13
Accept NumPy scalar coefficients across operator and representation types

### LINKED ISSUE #1097: `QubitOperator` incompatible with `numpy==2.3.1`
### Describe the issue

This issue was encountered through our effort of dropping Python 3.10 support in `pennylane` (see [PR](https://github.com/PennyLaneAI/pennylane/pull/7935)). This change bumped our CI version of `numpy` from `2.2.6` to `2.3.1` which resulted in two of our `qchem` tests that use `openfermion` as the backend failing. After some investigation it seems that the `SymbolicOperator` class is incompatible with some changes in `2.3.1`.

1.7.1

The following code works with `numpy==2.2.6` but fails with `numpy==2.3.1`,
```python
import openfermion
import numpy as np

l = [0, 1, 0, 0, 1, 1]
x = np.count_nonzero(l)

openfermion.ops.QubitOperator(term=(), coefficient=(-1)**x)
```
The error message is,
```
    122 def __init__(self, term=None, coefficient=1.0):
    123     if not isinstance(coefficient, COEFFICIENT_TYPES):
--> 124        raise ValueError('Coefficient must be a numeric type. Got {}'.format(type(coefficient)))
    126     # Initialize the terms dictionary
    127     self.terms = {}

ValueError: Coefficient must be a numeric type. Got <class 'numpy.int64'>
```

A simple fix would be to use the `numbers` package instead of `COEFFICIENT_TYPES`,
```python
import numbers
...
    def __init__(self, term=None, coefficient=1.0):
        if not isinstance(coefficient, numbers.Number):
            raise ValueError('Coefficient must be a numeric type. Got {}'.format(type(coefficient)))
...
```

blazingphoenix7:
Fixes #1097.

The actual problem behind #1097 is that NumPy's scalar types aren't subclasses of Python's `int`, `float`, or `complex`. `numpy.int64`, `numpy.float32`, `numpy.complex64` and the rest all fail an `isinstance(x, (int, float, complex))` check, which is the idiom the coefficient validation uses in a few places. (The one that slips through is `numpy.float64`, since it does subclass `float`, which is probably why this went unnoticed for a while.)

The `SymbolicOperator` family already got fixed by adding `numbers.Number` to its accepted types, so `FermionOperator`, `QubitOperator` and so on are fine now. But the same `(int, float, complex)` pattern shows up in other spots that the change didn't reach, and those still reject NumPy scalars:

- `PolynomialTensor` (so `InteractionOperator`, `InteractionRDM`, ...)
- `DOCIHamiltonian`
- `MajoranaOperator`
- the `majorana_operator` builder in `special_operators.py`

This adds `numbers.Number` to each of them, matching the existing fix. `numbers.Number` covers Python and NumPy numeric scalars equally, so a coefficient behaves the same whether it came from Python or NumPy, and the plain int/float/complex cases stay exactly as they were.

One thing worth calling out: dividing or multiplying a real-dtype tensor by a complex scalar still raises, because NumPy won't do the in-place cast from complex to float. That happens with a plain Python `complex` too, so it's pre-existing and unrelated to this change, and I left it alone.

Each affected module gets a test asserting that NumPy scalar coefficients (`int64`, `float32`, `complex64`) give the same result as the equivalent Python scalar, including the in-place operators on `MajoranaOperator`. The tests fail on `main` and pass with this change.

### DIFF
--- a/src/openfermion/hamiltonians/special_operators.py
+++ b/src/openfermion/hamiltonians/special_operators.py
@@ -17,6 +17,8 @@ import openfermion.config as config
 from openfermion.ops.operators import BosonOperator, FermionOperator
 from openfermion.utils.indexing import down_index, up_index
 
+COEFFICIENT_TYPES = config.COEFFICIENT_TYPES
+
 
 def s_plus_operator(n_spatial_orbitals: int) -> FermionOperator:
     r"""Return the s+ operator.
@@ -237,7 +239,7 @@ def majorana_operator(
     Returns:
         FermionOperator
     """
-    if not isinstance(coefficient, config.COEFFICIENT_TYPES):
+    if not isinstance(coefficient, COEFFICIENT_TYPES):
         raise ValueError('Coefficient must be scalar.')
 
     # If term is a string, convert it to a tuple
--- a/src/openfermion/hamiltonians/special_operators.py
+++ b/src/openfermion/hamiltonians/special_operators.py
@@ -11,9 +11,9 @@
 #   limitations under the License.
 """Commonly used operators (mainly instances of SymbolicOperator)."""
 
-import numbers
 from typing import Optional, Union, Tuple
 
+import openfermion.config as config
 from openfermion.ops.operators import BosonOperator, FermionOperator
 from openfermion.utils.indexing import down_index, up_index
 
@@ -237,7 +237,7 @@ def majorana_operator(
     Returns:
         FermionOperator
     """
-    if not isinstance(coefficient, (int, float, complex, numbers.Number)):
+    if not isinstance(coefficient, config.COEFFICIENT_TYPES):
         raise ValueError('Coefficient must be scalar.')
 
     # If term is a string, convert it to a tuple
--- a/src/openfermion/hamiltonians/special_operators.py
+++ b/src/openfermion/hamiltonians/special_operators.py
@@ -11,6 +11,7 @@
 #   limitations under the License.
 """Commonly used operators (mainly instances of SymbolicOperator)."""
 
+import numbers
 from typing import Optional, Union, Tuple
 
 from openfermion.ops.operators import BosonOperator, FermionOperator
@@ -236,7 +237,7 @@ def majorana_operator(
     Returns:
         FermionOperator
     """
-    if not isinstance(coefficient, (int, float, complex)):
+    if not isinstance(coefficient, (int, float, complex, numbers.Number)):
         raise ValueError('Coefficient must be scalar.')
 
     # If term is a string, convert it to a tuple
--- a/src/openfermion/ops/operators/majorana_operator.py
+++ b/src/openfermion/ops/operators/majorana_operator.py
@@ -179,7 +179,7 @@ class MajoranaOperator:
         return minuend
 
     def __mul__(self, other):
-        if not isinstance(other, (type(self), *COEFFICIENT_TYPES)):
+        if not isinstance(other, (type(self), COEFFICIENT_TYPES)):
             return NotImplemented
 
         if isinstance(other, COEFFICIENT_TYPES):
@@ -198,7 +198,7 @@ class MajoranaOperator:
         return MajoranaOperator.from_dict(terms)
 
     def __imul__(self, other):
-        if not isinstance(other, (type(self), *COEFFICIENT_TYPES)):
+        if not isinstance(other, (type(self), COEFFICIENT_TYPES)):
             return NotImplemented
 
         if isinstance(other, COEFFICIENT_TYPES):
--- a/src/openfermion/ops/operators/majorana_operator.py
+++ b/src/openfermion/ops/operators/majorana_operator.py
@@ -13,7 +13,11 @@
 
 import copy
 import itertools
+import numbers
+
 import numpy
+
+COEFFICIENT_TYPES = (int, float, complex, numbers.Number)
 
 
 class MajoranaOperator:
@@ -78,7 +82,7 @@ class MajoranaOperator:
 
     def commutes_with(self, other):
         """Test commutation with another MajoranaOperator"""
-        if isinstance(other, (int, float, complex)):
+        if isinstance(other, COEFFICIENT_TYPES):
             return True
 
         if not isinstance(other, type(self)):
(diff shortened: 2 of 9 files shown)


blazingphoenix7:
Thanks for the review. Both points are addressed in 91517da:

- Moved the three new NumPy-scalar tests (in `polynomial_tensor_test.py`, `doci_hamiltonian_test.py`, and `special_operators_test.py`) into their existing `unittest.TestCase` classes and switched to `self.assertEqual`, so they're collected under plain `unittest` discovery and not just pytest. I confirmed they run via `python -m unittest`.

- On the `isinstance` checks in `MajoranaOperator.__mul__`/`__imul__`: I kept a single `isinstance` but dropped the tuple unpacking, using `(type(self), COEFFICIENT_TYPES)` instead of `(type(self), *COEFFICIENT_TYPES)`. That removes the per-call unpacking that was flagged while staying one `isinstance` call. Splitting it into two `isinstance` calls trips pylint's `consider-merging-isinstance` (R1701), which would fail the lint check, so the nested-tuple form is the version that satisfies both.

src/openfermion/ops/operators/majorana_operator.py
```diff
@@ -175,10 +179,10 @@ def __sub__(self, other):
         return minuend
 
     def __mul__(self, other):
-        if not isinstance(other, (type(self), int, float, complex)):
+        if not isinstance(other, (type(self), COEFFICIENT_TYPES)):
```
mhucka: This tuple is constructed on every call in a couple of places in this class. It would be slightly more efficient and more readable to define a tuple for this,

```python
_MAJORANA_MUL_TYPES = (MajoranaOperator, int, float, complex, numbers.Number)
```

and then use it in place of `(type(self), COEFFICIENT_TYPES)`.

src/openfermion/ops/operators/majorana_operator.py
```diff
@@ -13,8 +13,12 @@
 
 import copy
 import itertools
+import numbers
+
 import numpy
 
+COEFFICIENT_TYPES = (int, float, complex, numbers.Number)
```
mhucka: This definition of `COEFFICIENT_TYPES` is repeated in multiple files. An approach to avoiding this, to keep the code more DRY and maintainable, would be to define this constant in one file (namely `src/openfermion/config.py`), like this:

```python
import numbers

COEFFICIENT_TYPES = (int, float, complex, numbers.Number)
```

and change the other files to do the equivalent of the following:

```python
import openfermion.config as config

COEFFICIENT_TYPES = config.COEFFICIENT_TYPES
```

Doing it this way

1. is more efficient than referencing `config.COEFFICIENT_TYPES` multiiple times all over the code, because it avoids Python having to do multiple symbol table lookups

2. is in keeping with the project principles of not importing individual symbols from modules (i.e., not doing `from openfermion.config import COEFFICIENT_TYPES`.

Can you make the appropriate changes? (Note that the imports may need to be moved to the appropriate place in the imports portion of the files, not put literally right before the uses as I showed above.)

mhucka (changes requested):
Thank you for working on this. I had a couple of minor change requests; otherwise, it's looking good.

blazingphoenix7:
Thanks, both done in 7942bbe.

- Moved `COEFFICIENT_TYPES` into `openfermion/config.py` and changed `majorana_operator.py`, `polynomial_tensor.py`, and `doci_hamiltonian.py` to `import openfermion.config as config` with a local `COEFFICIENT_TYPES = config.COEFFICIENT_TYPES` binding, so it is defined in one place and each module still does a single local lookup. `special_operators.py` uses `config.COEFFICIENT_TYPES` directly for its one check. While I was in `polynomial_tensor.py` I also moved its `EQ_TOLERANCE` onto the same `config` import so the file does not mix `import openfermion.config as config` with `from openfermion.config import ...`.

- Added a module-level `_MAJORANA_MUL_TYPES = (MajoranaOperator,) + COEFFICIENT_TYPES`, built once after the class, and used it in `__mul__` and `__imul__` instead of constructing the tuple on every call.

One thing I left alone: `symbolic_operator.py` also defines a `COEFFICIENT_TYPES`, but its version additionally includes `sympy.Expr` and it is not part of this PR, so I did not touch it. Happy to fold it into the config constant (e.g. `config.COEFFICIENT_TYPES + (sympy.Expr,)`) here or in a follow-up if you would like that unified too.

src/openfermion/ops/operators/majorana_operator.py
blazingphoenix7: Done in 7942bbe. Added a module-level `_MAJORANA_MUL_TYPES = (MajoranaOperator,) + COEFFICIENT_TYPES`, built once after the class, and used it in both `__mul__` and `__imul__`.

src/openfermion/ops/operators/majorana_operator.py
blazingphoenix7: Done in 7942bbe. Moved it to `openfermion/config.py`; the other modules now do `import openfermion.config as config` and bind `COEFFICIENT_TYPES = config.COEFFICIENT_TYPES` locally. I left `symbolic_operator.py` as-is since its version also includes `sympy.Expr` and isn't part of this PR, but I can fold it in too if you want.

src/openfermion/hamiltonians/special_operators.py
```diff
@@ -236,7 +237,7 @@ def majorana_operator(
     Returns:
         FermionOperator
     """
-    if not isinstance(coefficient, (int, float, complex)):
+    if not isinstance(coefficient, config.COEFFICIENT_TYPES):
```
mhucka: I think the latest changes missed this one. The reference to config.COEFFICIENT_TYPES can be pulled up to an assignment at the module level, like you did in majorana_operators.py

mhucka (changes requested):

src/openfermion/hamiltonians/special_operators.py
blazingphoenix7: Good catch, done in 9dc03f2. Pulled it up to a module-level `COEFFICIENT_TYPES = config.COEFFICIENT_TYPES` in `special_operators.py` to match the other files.

mhucka (approved):
Thanks for hanging in there! This looks good. Going to merge it now.

PR #1365, opened 2026-06-16
Fix mypy errors in batch 2 part 3 (#1282)

rosspeili:
plane_wave_hamiltonian: annotate spins as list[Optional[int]]. physical_costing: MagicStateFactory rounds and failure_rate are floats at runtime.

### DIFF
--- a/src/openfermion/hamiltonians/plane_wave_hamiltonian.py
+++ b/src/openfermion/hamiltonians/plane_wave_hamiltonian.py
@@ -11,8 +11,6 @@
 #   limitations under the License.
 """Construct Hamiltonians in plan wave basis and its dual in 3D."""
 
-from typing import Tuple, Optional, Union
-
 import numpy as np
 
 from openfermion.hamiltonians.jellium import jellium_model, jordan_wigner_dual_basis_jellium
@@ -26,10 +24,10 @@ import openfermion.chem.molecular_data as md
 
 def dual_basis_external_potential(
     grid: Grid,
-    geometry: list[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
+    geometry: list[tuple[str, tuple[int | float, int | float, int | float]]],
     spinless: bool,
     non_periodic: bool = False,
-    period_cutoff: Optional[float] = None,
+    period_cutoff: float | None = None,
 ) -> FermionOperator:
     """Return the external potential in the dual basis of arXiv:1706.00023.
 
@@ -55,8 +53,8 @@ def dual_basis_external_potential(
     """
     if non_periodic and period_cutoff is None:
         period_cutoff = grid.volume_scale() ** (1.0 / grid.dimensions)
-    operator = None
-    spins: list[Optional[int]]
+    operator: FermionOperator | None = None
+    spins: list[int | None]
     if spinless:
         spins = [None]
     else:
@@ -116,11 +114,11 @@ def filter_plane_wave_operator(
 
 def plane_wave_external_potential(
     grid: Grid,
-    geometry: list[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
+    geometry: list[tuple[str, tuple[int | float, int | float, int | float]]],
     spinless: bool,
-    e_cutoff: Optional[float] = None,
+    e_cutoff: float | None = None,
     non_periodic: bool = False,
-    period_cutoff: Optional[float] = None,
+    period_cutoff: float | None = None,
 ) -> FermionOperator:
     """Return the external potential operator in plane wave basis.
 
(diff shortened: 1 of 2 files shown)


src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -56,6 +56,7 @@ def dual_basis_external_potential(
     if non_periodic and period_cutoff is None:
         period_cutoff = grid.volume_scale() ** (1.0 / grid.dimensions)
     operator = None
+    spins: List[Optional[int]]
```
mhucka: Hmmm shouldn't this use the Python `list` type instead of the `typing` types?

mhucka (approved):

mhucka (changes requested):
Please check the type annotation in plane_wave_hamiltonians.py

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
mhucka: Oh, LOL, I just read GCA's review telling you to go in the opposite direction.

I think we want to go to `list` if possible. GCA's comment about Python 3.9 is irrelevant as OpenFermion only support 3.10+ anyway. I'd change the other `List` cases in the file (if there are any, as GCA seems to think) to `list` instead. I assume that's what was being used in the other mypy PRs?

rosspeili:
@mhucka thanks and agreed, switched back to `list[Optional[int]]` and updated the other `List[...]` annotations in this file to `list[...]` too similar to previous PRs.

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -56,6 +56,7 @@ def dual_basis_external_potential(
     if non_periodic and period_cutoff is None:
         period_cutoff = grid.volume_scale() ** (1.0 / grid.dimensions)
     operator = None
+    spins: list[Optional[int]]
```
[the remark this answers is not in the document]
mhucka: This project assumes Python 3.10+, so the change above is not necessary.

mhucka:
```suggestion
    spins: list[int | None]
```

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -26,7 +26,7 @@
 
 def dual_basis_external_potential(
     grid: Grid,
-    geometry: List[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
+    geometry: list[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
     spinless: bool,
     non_periodic: bool = False,
     period_cutoff: Optional[float] = None,
```
mhucka: You can use T | None instead of Optional[T]:

```suggestion
    period_cutoff: float | None = None,
```

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -11,7 +11,7 @@
 #   limitations under the License.
 """Construct Hamiltonians in plan wave basis and its dual in 3D."""
 
-from typing import List, Tuple, Optional, Union
+from typing import Tuple, Optional, Union
```
mhucka: Similar to how we're using `list`, we can also use `tuple` instead of `Tuple` in the type annotations. In addition, looks like Union and Optional are not needed. (See comments below.) So, I think this whole import can go away.

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -26,7 +26,7 @@
 
 def dual_basis_external_potential(
     grid: Grid,
-    geometry: List[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
+    geometry: list[Tuple[str, Tuple[Union[int, float], Union[int, float], Union[int, float]]]],
```
mhucka: This can use `tuple` instead of `Tuple`, and `Union[int, float]` can be replaced by `int | float`. The same change applies to the other places where this pattern occurs in this PR, but I will skip repeating the same comment.

```suggestion
    geometry: list[tuple[str, tuple[int | float, int | float, int | float]]],
```

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
```diff
@@ -56,6 +56,7 @@ def dual_basis_external_potential(
     if non_periodic and period_cutoff is None:
         period_cutoff = grid.volume_scale() ** (1.0 / grid.dimensions)
     operator = None
```
mhucka: Missing a type annotation. Looks like that should be:


```suggestion
    operator = FermionOperator | None = None
```

src/openfermion/hamiltonians/plane_wave_hamiltonian.py
mhucka: Note: a similar type declaration occurs in more places below. Please also make the same adjustment to the other ones.

mhucka (changes requested):
Thanks for working on this. I recommend some tweaks below.

rosspeili:
@mhucka thanks for the tip, and sorry for not coming up with this myself, applied all tweaks in latest push, also removed the `typing` import. Let me know how this looks.

mhucka (approved):
Great, thanks for doing this!

PR #1366, opened 2026-06-17
Fix inconsistencies in double factorization cost estimation

### LINKED ISSUE #839: Inconsistencies in the double factorized chemistry resource estimate costing function

arettig:
This fixes #839. Double factorization uses different precisions for $\theta$ for the first and second factorizations (leading to differences in gate counts), but when estimating the cost of double factorization, the first precision is used for both factorizations. This was corrected and the tests updated to reflect the changed cost function.

### DIFF
--- a/src/openfermion/resource_estimates/df/compute_cost_df.py
+++ b/src/openfermion/resource_estimates/df/compute_cost_df.py
@@ -81,13 +81,13 @@ def compute_cost(
         )
 
     # Bits of precision for rotation
-    br = int(np.argmin(oh) + 1)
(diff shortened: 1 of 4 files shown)


mhucka (approved):
LGTM

PR #1368, opened 2026-06-22
Fix mypy errors in batch 3 part 1 (#1282)

rosspeili:
Annotate Tensor optional fields and add guards in _namedtensor. Default beta in compute_thc_resources when None (same pattern as DF). Return empty FermionOperator when plane_wave external potential has no terms.


rosspeili:
@mhucka I moved to batch 3 from #1282, this PR fixes strict-mypy issues in `_namedtensor`, `compute_thc_resources` (default `beta` when `None`, same as DF), and a `plane_wave_hamiltonian` return narrowing leftover from #1365. Let me know how this looks and can move to next part.

mhucka (approved):
Thanks for doing this. This looks very good.

rosspeili:
Thank you @mhucka, moving to next part.

PR #1370, opened 2026-06-22
Add support for arbitrarily large spin states

### LINKED ISSUE #809: multiplicity limitation in molecular data
I wanted to use big hydrogen chains for benchmarking of my code.
However, as it turns out, the biggest Hchain I can created using the following piece of code is of size 12:

```
from openfermion import MolecularData
number_of_hydrogens=12
basis="6-31g"
bond_distance = 1.3
charge = 0
multiplicity = number_of_hydrogens + 1 
geometry = [("H", (0, 0, i * bond_distance)) for i in range(number_of_hydrogens)]
MolecularData(geometry, basis, multiplicity, charge)
```

For higher values of `number_of_hydrogens` I get an error. After some investigation I have realized that it's because of `name_molecule` function, which has a `multiplicity_dict` with keys up to 12. 
I don't even need to get the name of the molecule anywhere, so it's unfortunate that this exception is being raised – I need to either use my own custom version Openfermion or switch to using pySCF.

CC: @sunchong137

arettig:
Fixes #809  Previously the spin state name was obtained using a dictionary mapping integer multiplicity to the name string. Only up to 11 unpaired electrons were supported this way, with an error thrown for anything larger. With this change, larger spins are named via n-multiplet (e.g. 12 unpaired electrons is named 13-multiplet).

Tests are added to check the naming for large spin states is correct, and that non-integer spin states still throw an error.

### DIFF
--- a/src/openfermion/chem/molecular_data.py
+++ b/src/openfermion/chem/molecular_data.py
@@ -275,10 +275,10 @@ def name_molecule(geometry, basis, multiplicity, charge, description):
     }
     if multiplicity in multiplicity_dict:
         name += '_{}'.format(multiplicity_dict[multiplicity])
-    elif isinstance(multiplicity, (int, numpy.integer)) and multiplicity > 0:
+    elif multiplicity > 0:
         name += '_{}-multiplet'.format(multiplicity)
     else:
-        raise MoleculeNameError('Invalid spin multiplicity provided.')
+        raise ValueError('Invalid spin multiplicity provided.')
 
     # Add charge.
     if charge > 0:
@@ -493,7 +493,12 @@ class MolecularData:
         # Metadata fields which must be provided.
         self.geometry = geometry
         self.basis = basis
-        self.multiplicity = multiplicity
+        if isinstance(
+            multiplicity, (int, float, numpy.integer, numpy.floating)
+        ) and multiplicity == int(multiplicity):
+            self.multiplicity = int(multiplicity)
+        else:
+            raise ValueError('Invalid spin multiplicity provided.')
 
         # Metadata fields with default values.
         self.charge = charge
@@ -502,7 +507,7 @@ class MolecularData:
         self.description = description
 
         # Name molecule and get associated filename
-        self.name = name_molecule(geometry, basis, multiplicity, charge, description)
+        self.name = name_molecule(geometry, basis, self.multiplicity, charge, description)
         if filename:
             if filename[-5:] == '.hdf5':
                 filename = filename[: (len(filename) - 5)]
--- a/src/openfermion/chem/molecular_data.py
+++ b/src/openfermion/chem/molecular_data.py
@@ -33,7 +33,9 @@ r"""NOTE ON PQRS CONVENTION:
 
 
 # Define error objects which inherit from Exception.
-class MoleculeNameError(Exception):
+class MoleculeNameError(ValueError):
+    """Exception raised when a molecule name is invalid or cannot be generated."""
+
     pass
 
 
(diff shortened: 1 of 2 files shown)


src/openfermion/chem/molecular_data.py
```diff
@@ -273,10 +273,12 @@ def name_molecule(geometry, basis, multiplicity, charge, description):
...
     }
-    if multiplicity not in multiplicity_dict:
-        raise MoleculeNameError('Invalid spin multiplicity provided.')
-    else:
+    if multiplicity in multiplicity_dict:
         name += '_{}'.format(multiplicity_dict[multiplicity])
+    elif isinstance(multiplicity, (int, numpy.integer)) and multiplicity > 0:
+        name += '_{}-multiplet'.format(multiplicity)
+    else:
+        raise MoleculeNameError('Invalid spin multiplicity provided.')
```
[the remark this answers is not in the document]
arettig: I added a check to the constructor of molecular_data so non-integer values are caught there and the multiplicity parameter of the `name_molecule` will always be an integer.

src/openfermion/chem/molecular_data.py
```diff
@@ -273,10 +273,12 @@ def name_molecule(geometry, basis, multiplicity, charge, description):
...
     }
-    if multiplicity not in multiplicity_dict:
-        raise MoleculeNameError('Invalid spin multiplicity provided.')
-    else:
+    if multiplicity in multiplicity_dict:
         name += '_{}'.format(multiplicity_dict[multiplicity])
+    elif multiplicity > 0:
+        name += '_{}-multiplet'.format(multiplicity)
+    else:
+        raise ValueError('Invalid spin multiplicity provided.')
```
mhucka: This is changing the exception type from `MoleculeNameError` to `ValueError`, which makes it a backward incompatible change. It's not necessarily a bad idea to use `ValueError` here (since semantically, the situation here really does involve a "value error"), there's potential (albeit small) for someone's code to break unexpectedly.

Since `MoleculeNameError` is defined in this file and only used here, one way to get our cake and eat it too is to retain `MoleculeNameError` but make it inherit from `ValueError` instead of the generic `Exception` class:

```python
class MoleculeNameError(ValueError):
    """Exception raised when a molecule name is invalid or cannot be generated."""
```

Then the exception raised on line 281 can be kept as `MoleculeNameError`.

src/openfermion/chem/molecular_data_test.py
```diff
@@ -101,9 +100,37 @@ def test_invalid_multiplicity(self):
         geometry = [('H', (0.0, 0.0, 0.0)), ('H', (0.0, 0.0, 0.7414))]
         basis = 'sto-3g'
         multiplicity = -1
-        with self.assertRaises(MoleculeNameError):
+        with self.assertRaises(ValueError):
```
mhucka: If my prior comment is acted on, this will need to be changed back to `MoleculeNameError`.

(Same comment applies to the other cases of `ValueError` below.

mhucka (changes requested):

src/openfermion/chem/molecular_data.py
arettig: Ah that is a good point about backwards compatibility. I like your solution of making `MoleculeNameError` inherit from `ValueError`. Thanks!

Changed as suggested.

src/openfermion/chem/molecular_data_test.py
arettig: Done.

mhucka (approved):
Thank you for this!

PR #1378, opened 2026-06-23
Fix up a few minor things in `setup.py`

mhucka:
This PR:

* removes an unnecessary header line prepended to the description
* adds a missing `long_description_content_type` value
* improves the short description

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,8 +20,7 @@ __version__ = runpy.run_path('src/openfermion/_version.py')['__version__']
 assert __version__, 'Version string cannot be empty'
 
 description = (
-    'Package to compile and analyze quantum algorithms for '
-    'simulating fermionic systems.'
+    'Package to compile and analyze quantum algorithms for ' 'simulating fermionic systems.'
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -19,12 +19,15 @@
 __version__ = runpy.run_path('src/openfermion/_version.py')['__version__']
 assert __version__, 'Version string cannot be empty'
 
-# The readme file is used as the long_description:
-long_description = '===========\n' + 'OpenFermion\n' + '===========\n\n'
+description = (
+    'Package to compile and analyze quantum algorithms for ' 'simulating fermionic systems.'
```
arettig, line 23: nit: the string is split into two but both parts are on the same line.

arettig (approved):
looks good - just one insignificant nit

setup.py
mhucka: Oh, huh, it wasn't originally. Guess running the formatter moved it and I didn't notice. Thanks for catching that.

PR #1380, opened 2026-06-25
Fix factor of 2 in Rz operator in trotter_export_to_qgates_test.py

### LINKED ISSUE #721: Angle of rotation in trotterize_exp_qubop_to_qasm is off by a factor of 2

arettig:
Fixes #721 (supersedes #1343). A factor of 2 error in Rz gates is corrected when outputting a Trotter decomposition to QASM (for both the controlled and uncontrolled versions). The corresponding tests are also corrected.

### DIFF
--- a/src/openfermion/circuits/trotter_exp_to_qgates.py
+++ b/src/openfermion/circuits/trotter_exp_to_qgates.py
@@ -243,19 +243,21 @@ def pauli_exp_to_qasm(qubit_operator_list, evolution_time=1.0, qubit_list=None, 
                 if len(qids) > 0:
                     ret_list = ret_list + [
                         "C-Phase {} {} {}".format(
-                            -2 * term_coeff * evolution_time, ancilla, qids[-1]
+                            2 * term_coeff * evolution_time, ancilla, qids[-1]
                         )
                     ]
                     ret_list = ret_list + [
-                        "Rz {} {}".format(1 * term_coeff * evolution_time, ancilla)
+                        "Rz {} {}".format(-1 * term_coeff * evolution_time, ancilla)
                     ]
                 else:
                     ret_list = ret_list + [
-                        "Rz {} {}".format(1 * term_coeff * evolution_time, ancilla)
+                        "Rz {} {}".format(-1 * term_coeff * evolution_time, ancilla)
                     ]
             else:
                 if len(qids) > 0:
-                    ret_list = ret_list + ["Rz {} {}".format(term_coeff * evolution_time, qids[-1])]
+                    ret_list = ret_list + [
+                        "Rz {} {}".format(2 * term_coeff * evolution_time, qids[-1])
+                    ]
 
             # 4. Second set of CNOTs
             ret_list = ret_list + cnots2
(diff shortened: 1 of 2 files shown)


mhucka:
It's weird how Gemini Code Assist is all over some PRs, and other ones it just kinda wipes its hands and says "eh, not my concern."

src/openfermion/circuits/trotter_exp_to_qgates.py
```diff
@@ -243,19 +243,21 @@ def pauli_exp_to_qasm(qubit_operator_list, evolution_time=1.0, qubit_list=None,
...
                     ]
                     ret_list = ret_list + [
-                        "Rz {} {}".format(1 * term_coeff * evolution_time, ancilla)
+                        "Rz {} {}".format(-1 * term_coeff * evolution_time, ancilla)
                     ]
                 else:
                     ret_list = ret_list + [
-                        "Rz {} {}".format(1 * term_coeff * evolution_time, ancilla)
+                        "Rz {} {}".format(-1 * term_coeff * evolution_time, ancilla)
                     ]
```
mhucka: Can this code duplication be avoided somehow?

mhucka: Also, I'm not sure the `-1` part is needed. It should work to write the more idiomatic form

```python
-term_coeff * evolution_time
```

unless there's a reason for human readability or something else?

mhucka: One more thing: instead of `somelist + [something]`, it's more efficient to do `somelist.append(something)`. The `+` operator creates a new copy of the list, which `append` avoids.

However, if the rest of the file has many more instances of the same style of construct (`ret_list` = `rel_list + […]`), then I would say it's better to do what you did here and keep the same pattern. A separate PR can fix the inefficient constructs (and do it for other files, because if it happens here it probably happens elsewhere in the code base).

mhucka (approved):
Thanks for doing this. I have some comments that are not critical, so I approved the PR and will leave it to you to address them (or not) before merging.

src/openfermion/circuits/trotter_exp_to_qgates.py
arettig: These are all excellent points, thanks!

Fixed!

PR #1381, opened 2026-06-25
Feature/issue 1333 pyproject modernize

### LINKED ISSUE #1333: Move setup.py metadata and as many tool configurations into `pyproject.toml` as possible

Following modern best practices, the project metadata that is currently in `setup.py` should be moved into `pyproject.toml`. In addition, many tool configurations can be moved out of individual rc files and into `pyproject.toml` too. I did this for qsim and it has worked out well.

Start by looking at the various dot files in the repo and seeing how the settings can be written equivalently in the `pyproject.toml` file. 

None

rosspeili:
Moved project metadata from `setup.py` into `pyproject.toml` (PEP 621), modeled on qsim: `[project]` metadata, dynamic version/deps from `dev_tools/requirements/deps/*.txt`, `resources` optional extra, and a PEP 735 `dev` dependency group for `uv sync --group dev`. Removed `setup.py`, and updated `MANIFEST.in` and `_version.py` accordingly.

Tested with `python -m build` (sdist + wheel), `pip install -e .`, and full pytest (2196 passed; 6 Windows-only parallel-test failures also on `main`), also updated `CONTRIBUTING.md` to point at `pyproject.toml` as the canonical config and document the `uv sync --group dev` workflow.

**Intentionally deferred:**
- Migrating `.pylintrc` / other dotfile tool configs into `pyproject.toml`
- Adding a committed `uv.lock` (if we want that workflow?)
- Changing CI to use `uv sync` instead of locked `dev_tools/requirements/envs/*.txt` (those files and `create-env-files.sh` are unchanged)

Closes #1333


rosspeili:
@mhucka keeping `license = "Apache-2.0"` (SPDX, PEP 639) per qsim. Adding the License trove classifier back causes setuptools to fail the build (`License classifiers have been superseded by license expressions`). Verified with `python -m build`. Should I ignore Gemini assist's recommendations?

mhucka:
> @mhucka keeping `license = "Apache-2.0"` (SPDX, PEP 639) per qsim. Adding the License trove classifier back causes setuptools to fail the build (`License classifiers have been superseded by license expressions`). Verified with `python -m build`. Should I ignore Gemini assist's recommendations?

Thank you for that work!

Re GCA's comments about the `license` field: it's wrong. PEP 621 has been superceded by PEP 639 and (as you noted) that one says [string values are preferred](https://peps.python.org/pep-0639/#add-string-value-to-license-key). Re GCA's comments about adding the classifier: let's ignore that one too, if setuptools doesn't work with it.

mhucka:
Regarding the timing of merging this PR: we're trying to get a release out this week, so we're going to leave this until after the release.

rosspeili:
Thanks @mhucka, glad this looks on track. No stress from my side, happy to wait until after the release. Ping me if anything else should change before merge. <3

pyproject.toml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
```
mhucka: Per company guidance, we need to keep the copyright years unchanged when modifying existing files.

CONTRIBUTING.md
```diff
@@ -222,12 +222,22 @@ After that, the hooks will run automatically when triggered by the corresponding
...
 
+    Or, with [uv](https://docs.astral.sh/uv/) (locked files are still generated from
+    `dev_tools/requirements/deps/*.txt`; see _Regenerating requirements files_ below):
+
+    ```shell
+    uv sync --group dev
+    ```
+
+    Package metadata and runtime dependencies are defined in `pyproject.toml` (there is no
+    `setup.py`).
```
mhucka: Probably better to leave off the parenthetical remark; the explanation stands on its own and adding the remark may actually confuse some readers. ("Wait, why are they making a point of telling me there's no setup.py?")

```suggestion
    Package metadata and runtime dependencies are defined in `pyproject.toml`.
```

mhucka (commented):
Thank you for taking this on, and sorry for the long delay in getting back to this PR.

This looks good. I have just a couple of fussy change requests.

Before merging this, I'd like to merge https://github.com/quantumlib/OpenFermion/pull/1445 first, to update the version of Pylint; once the other PR is merged, if you could update the version of pylint in this pyproject.toml, the result should help ensure everything works together.

rosspeili:
Nw at all @mhucka, and thanks for the review. Copyright restored and won't touch again, explainer referencing non existing setup.py dropped, and `pylint~=4.0` is already in `pyproject.toml` so this should line up once #1445 lands. Lmk how this looks <3

mhucka:
@rosspeili Other PR merged. Once you update the branch on this one, I think it'll be ready to merge.

rosspeili:
@mhucka done.

mhucka (approved):
Thanks again!

PR #1384, opened 2026-06-26
Make minor updates and tweaks to pre-commit setup

mhucka:
This makes some minor updates and tweaks to `.pre-commit-config.yaml`:

* Define the hook stages that should be installed, which simplifies the installation instructions to just `pre-commit install` without having to list 3 separate stages

* Use hadolint-py for the Docker hook instead of the previous hook; this one lets pre-commit automatically install a Python-wrapped installation of the hadolint program and saves the user from having to install hadolint manually.

* Add `config-file-validator` hook.

* Remove the Markdown link checker hook because it took too long to run.

* Add more common patterns to `exclude`.

* Fix up some hooks

* Update version hashes to latest versions as needed.

* Change the style of the descriptions printed by pre-commit, to (hopefully) make them more natural.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,45 +66,45 @@ repos:
       # Use separate yamllint hook later in this file.
 
       - id: no-commit-to-branch
-        name: 'Check that the commit is not made to the main branch'
+        name: 'The commit is not being made to the main branch'
         args: [--branch, main, --branch, master]
 
       - id: check-merge-conflict
-        name: "Check that files are free of merge git conflict strings"
+        name: "Files are free of git merge conflict strings"
 
       - id: check-illegal-windows-names
-        name: "Check that file names can be used on Windows"
+        name: "File names are valid for use on Windows"
 
       - id: check-case-conflict
-        name: "Check that file names won't conflict on case-insensitive systems"
+        name: "File names won't conflict on case-insensitive systems"
 
       - id: name-tests-test
-        name: 'Verify test files are named correctly'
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -56,87 +66,96 @@ repos:
...
+      - id: shellcheck
+        name: 'Shell scripts are lint-free'
+        files: '(\.sh$|^check/[^.]+$)'
+
+  - repo: https://github.com/Boeing/config-file-validator
+    rev: fc3dea5b2d0bfd47a785ca47375fe6080151f97b  # frozen: v2.3.0
+    hooks:
+      - id: config-file-validator
+        name: 'Config files are free of syntax errors'
+        types_or: [ini, toml, yaml]
```
arettig: nit: There are pre-commit hooks for .toml (check-toml) and .yaml (yamllint) files already. Does this hook check something different for these file types? Otherwise we might be double checking the handful of .yaml and .toml files leading to a (very tiny) inefficiency.

arettig (approved):
Looks good. Tested locally and I can confirm the new pre-commit setup installs and runs correctly. One minor nit/question below.

.pre-commit-config.yaml
mhucka: Whoops, you're right; the `check-toml` one earlier in the file can be removed, and this entry for `config-file-validator` doesn't need one for toml because the only toml file is already handled by the validator for pyproject.toml. Thanks for catching that. I'll push an update in a moment.

PR #1385, opened 2026-06-27
Fix #1383: replace some noninclusive terms in various files

### LINKED ISSUE #1383: Fix noninclusive language
### Describe the issue

I came across some instances of the word "dummy" in a couple of the files. We should fix that and sweep everything for other common noninclusive language.

1.7.1 in git

_No response_

mhucka:
This replaces some noninclusive terms used in different files with more neutral alternatives. There are no functional changes.


docs/tutorials/binary_code_transforms.ipynb
```diff
@@ -134,7 +134,7 @@
     "print(BinaryPolynomial('w0') * BinaryPolynomial('w1 + 1') + BinaryPolynomial('1'))\n",
     "print(BinaryPolynomial([(1, 0), (1, ), ('one', )]))\n",
     "\n",
-    "print('The native data type structure can be seen here:')\n",
+    "print('The built-in data type structure can be seen here:')\n",
```
pavoljuhas: In this context `built-in` may have a different meaning than `native`.  Please check with the original author if the new wording is still accurate or consider leaving this and other instances of *native* as is.

With some possibility of having a blind spot, I don't perceive the use of the `native` adjective here or in phrases such as "someone's native language" as non-inclusive or potentially offensive.

src/openfermion/resource_estimates/pbc/thc/factorizations/isdf_test.py
```diff
@@ -407,7 +407,7 @@ def test_kpoint_isdf_symmetries():
     assert np.allclose(kpts[ik_prime] - kpts[ik_prime_minus_q] - kpts[iq], delta_Gs[iq][iGsr])
     # (pk qk-Q | rk'-Q sk') = (q k-Q p k | sk' rk'-Q)*
     ik_prime_minus_q = momentum_map[iq, ik_prime]
-    # uncomment to check normal eris
+    # uncomment to check standard eris
```
pavoljuhas: This may be overzealous.  It is perhaps safer to leave `normal` as is.

pavoljuhas (commented):
The replacement of `native` with `built-in` might be misleading.  Otherwise LGTM.

docs/tutorials/binary_code_transforms.ipynb
mhucka: It does sound a bit odd to use "built-in" in this context, although 3 different AI systems report it's the preferred alternative. However, reading the sentences again, I think they can be rewritten to be simpler and avoid the word altogether.

src/openfermion/resource_estimates/pbc/thc/factorizations/isdf_test.py
mhucka: A problem with "normal" is that it is imprecise and in mathematics means something other than what I think was intended here anyway. The simplest solution is to remove the adjective because elsewhere in the file, it's not qualified, and I suspect the same thing was meant here.

pavoljuhas (approved):

PR #1387, opened 2026-06-29
Fix remaining mypy errors for issue #1282 (PR1)

rosspeili:
Clear strict mypy with full follow-imports across 20 files: hamiltonians, transforms, resource estimates, and circuit generators. Generator functions now return Iterator[cirq.OP_TREE], add ase.* to mypy optional-dep overrides.

### DIFF
--- a/src/openfermion/hamiltonians/jellium.py
+++ b/src/openfermion/hamiltonians/jellium.py
@@ -154,7 +154,6 @@ def plane_wave_potential(
     spins = [None] if spinless else [0, 1]
     if non_periodic and period_cutoff is None:
         period_cutoff = grid.volume_scale() ** (1.0 / grid.dimensions)
-    period_cutoff_value = period_cutoff
 
     # Pre-Computations.
     shifted_omega_indices_dict: dict[tuple[int, ...], list[int]] = {}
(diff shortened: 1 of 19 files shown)


src/openfermion/hamiltonians/jellium.py
```diff
@@ -206,7 +207,8 @@ def plane_wave_potential(
             momenta_squared, grid.dimensions, grid.volume_scale()
         )
         if non_periodic:
-            coefficient *= 1.0 - math.cos(period_cutoff * math.sqrt(momenta_squared))
+            assert period_cutoff_value is not None
```
mhucka: This looks like it's meant to do a run-time sanity check. It's a good idea to do the check, but `assert` statements should be avoided in production code because assertions can be turned off via a run-time options. The litmus test is "Could the assert statement be completely removed without breaking the correct operation or security of the program under normal execution?"

src/openfermion/hamiltonians/jellium.py
```diff
@@ -310,7 +313,8 @@ def dual_basis_jellium_model(
                     momenta_squared, grid.dimensions, grid.volume_scale()
                 )
                 if non_periodic:
-                    coef *= 1.0 - math.cos(period_cutoff * math.sqrt(momenta_squared))
+                    assert period_cutoff_value is not None
```
mhucka: Assert statements should be avoided in production (non-test) code because assertions can be turned off via a run-time flag.

mhucka (changes requested):
This looks great. My one comment is about the use of `assert` statements for what appears to be run-time value checking.

rosspeili:
Hey @mhucka, sorry about that, that was my bad, asserts were only for mypy, not real runtime checks. I first swapped for raise ValueError(...), but that wasn’t great either >.< those branches were unreachable because we already default period_cutoff earlier, and ci flagged that.

Latest commit pulls default into a small _non_periodic_period_cutoff() helper used in both spots. Same behavior, no assert, no dead raise paths, and mypy is happy.

Thanks for the review, and direction. Let me know if I am missing anything. <3

mhucka (approved):
Looks good. Thanks!

PR #1397, opened 2026-07-02
Support Cirq 1.7 change in exponents that affects rendering of circuit diagrams

mhucka:
In versions of Cirq below 1.7, `cirq.EigenGate._diagram_exponent()` automatically canonicalized exponents into the interval (-period/2, period/2]. Because `DoubleExcitationGate` has a diagram period of 2, an exponent of `2.3` used to be canonicalized to `0.3` for diagram purposes (because 2.3 mod 2 ≡ 0.3).

In the just-released Cirq 1.7.0, the canonicalization logic was removed from `EigenGate._diagram_exponent()` – I think it was in PR https://github.com/quantumlib/Cirq/pull/7954. As a result, `DoubleExcitation(d, b, a, c) ** 2.3` is now rendered in diagrams with `^2.3` instead of `^0.3`. Some tests in `four_qubit_gates_test.py` compared circuit diagrams, and those tests fail in Cirq 1.7.0

Since we need to support both 1.7.0 and lower versions of Cirq, this PR adds a step to put the exponent in the canonical form that was used before. This seems like the best approach to preserve backward compatibility and still work with the latest version of Cirq.

### DIFF
--- a/src/openfermion/circuits/gates/four_qubit_gates.py
+++ b/src/openfermion/circuits/gates/four_qubit_gates.py
@@ -124,9 +124,17 @@ class DoubleExcitationGate(cirq.EigenGate):
             # Split up this string to avoid SyntaxError in Python 3.12.
             down_up = r'\/ /' + '\\'
             wire_symbols = (up_down, up_down, down_up, down_up)
-        return cirq.CircuitDiagramInfo(
-            wire_symbols=wire_symbols, exponent=self._diagram_exponent(args)
-        )
+
+        exponent = self._diagram_exponent(args)
+        if isinstance(exponent, (int, float)):
+            # Canonicalize rounded exponent into (-1, 1] range (DoubleExcitation has period 2)
+            h = 1.0
+            if not (-h < exponent <= h):
+                exponent = h - exponent
+                exponent %= 2.0
+                exponent = h - exponent
+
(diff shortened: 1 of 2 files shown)


src/openfermion/circuits/gates/four_qubit_gates.py
```diff
@@ -124,9 +124,13 @@ def _circuit_diagram_info_(self, args: cirq.CircuitDiagramInfoArgs) -> cirq.Circ
             # Split up this string to avoid SyntaxError in Python 3.12.
             down_up = r'\/ /' + '\\'
             wire_symbols = (up_down, up_down, down_up, down_up)
-        return cirq.CircuitDiagramInfo(
-            wire_symbols=wire_symbols, exponent=self._diagram_exponent(args)
-        )
+
+        exponent = self._diagram_exponent(args)
+        if isinstance(exponent, (int, float, np.integer, np.floating)):
```
pavoljuhas: optional - consider using `numbers.Real` instead.

pavoljuhas (approved):
LGTM with an optional suggestion.

src/openfermion/circuits/gates/four_qubit_gates.py
mhucka: Very good point.

PR #1401, opened 2026-07-08
Mark `test_kpoint_isdf_symmetries()` as flaky

mhucka:
This function in file `src/openfermion/resource_estimates/pbc/thc/factorizations/isdf_test.py` occasionally fails to converge. This PR marks it as flaky and tells pytest to retry it once, so that people don't waste time chasing ghosts.


arettig (approved):
LGTM

PR #1402, opened 2026-07-09
Add pytest-instafail package

mhucka:
This adds the package `pytest-instafail` to the requirements for the development environment, so that we can inform people how to use the `--instafail` to learn about failures immediately. The built-in pytest option `-x` and `--instafail` work in complementary fashion: `pytest -x` makes it stop on first failure (and subsequent tests are not executed); `pytest --instafail` makes it print failures immediately, but continue execution.

Along with that, this PR adds more helpful test info to `CONTRIBUTING.md`.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -270,7 +270,7 @@ for formatting code, and provide scripts to run them.
     check/pylint-changed-files
     ```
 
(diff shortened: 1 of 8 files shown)


mhucka:
Closing as not planned for now.

PR #1404, opened 2026-07-09
Make multiprocessing calls aware of parallelism used during testing

mhucka:
Pytest on parallel systems was taking far more time to run than it should have. The problem turned out to be due to the fact that some of the OpenFermion code used multiprocessing without taking account of the available CPU resources on the system. When running parallel pytests (with pytest-xdist), this resulted in the host system being vastly overcommitted: system load would skyrocked, and on a 64-core Linux system, it would take 45 minutes to run `pytest -n auto -m "not slow"`.

This PR changes the way that available CPUs are counted to include a check for whether parallely pytest workers are being used. This is done in two ways:

* Change the OpenFermion classes that use multiprocessing directly (e.g., `src/openfermion/linalg/linear_qubit_operator.py`)

* Use the threadpoolctl package to limit how many threads native libraries such as BLAS use, so that packages like SciPy and NumPy don't end up using threads without control.

Additional changes:

* Add the threadpoolctl package to the set of pytest dependencies in `dev_tools/requirements/deps/pytest.txt`.

* Move `src/openfermion/conftest.py` to the top level, for more consistent pytest behavior when it's invoked in different ways.

* Change `check/pylint-changed-files` and `check/pytest-and-incremental-coverage` to not use parallism by default, and to pass command line flags to pylint and pytest so that users or calling scripts can control when flags like `-n auto` are used. This makes them more consistent with other scripts in `check/`.

* Change `check/all` to pass flags for parallel execution to the other scripts, and to accept a new option `--no-parallel` to turn off that behavior. In addition, make `check/all` accept short-form argument names, shorten `--apply-format-changes` to `--apply-changes`, and rewrite some non-Bash constructs to be consistently Bash throughout.

* Update `CONTRIBUTING.md` to mention the use of `-n auto` with pytest, and the use of `check/all` instead of a long list of individual final checks.

Some of this work was done with the help of Gemini CLI.


src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -185,6 +186,9 @@ def _matvec(self, x):
         if not self.linear_operators:
             return numpy.zeros(x.shape)
 
+        if self.options.processes <= 1:
+            return functools.reduce(numpy.add, (operator * x for operator in self.linear_operators))
```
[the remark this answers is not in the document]
mhucka: Moved to a new issue

src/openfermion/linalg/linear_qubit_operator_test.py
```diff
@@ -262,6 +262,15 @@ def test_closed_workers_not_reused(self):
         parallel_qubit_op.dot(state)
         self.assertIsNone(parallel_qubit_op.options.pool)
 
+    def test_matvec_single_process(self):
+        """Tests that when processes is 1, it computes correctly and doesn't crash."""
+        qubit_operator = QubitOperator('Z3') + QubitOperator('Y0') + QubitOperator('X1')
```
arettig: nit: could just use `self.qubit_operator`

conftest.py
```diff
@@ -0,0 +1,68 @@
...
+    """Set a fixed random seed when testing."""
+    random.seed(0)
+    np.random.seed(0)
+
+
+@pytest.fixture(autouse=True, scope="session")
+def set_threadpool_limits():
+    """Limit number of threads to prevent oversubscription with pytest-xdist.
+
+    This only has an effect if the Python threadpoolctl package is installed,
```
arettig, line 51: should we add `threadpoolctl` to our developer dependencies if it's required to prevent oversubscription?

src/openfermion/config_test.py
```diff
@@ -0,0 +1,174 @@
...
+        with patch.object(config, "os", mock_os):
+            self.assertEqual(get_available_cpu_count(), 8)
+
+        # Test Case 1b: Python 3.13+ returns 0/None -> fallback to 1.
+        mock_os = MockOS(process_cpu_count_val=0)
+        with patch.object(config, "os", mock_os):
+            self.assertEqual(get_available_cpu_count(), 1)
+
+        # Test Case 2: Linux/Unix (no process_cpu_count, has sched_getaffinity).
+        mock_os = MockOS(sched_getaffinity_val=[1, 2, 3, 4])
```
arettig:
```suggestion
        mock_os = MockOS(sched_getaffinity_val=[1, 2, 3])
```

Since os.cpu_count() will also give a value of 4 it might be better to use 2 or 3 cores here to ensure the `sched_getaffinity_val` length is returned;

src/openfermion/config.py
```diff
@@ -24,3 +24,47 @@
...
+    This prevents libraries like NumPy, SciPy, and JAX from oversubscribing
+    available CPU resources on high-core-count or multi-worker environments.
+    """
+    if "PYTEST_XDIST_WORKER_COUNT" in os.environ:
+        limit = str(max(get_available_cpu_count(), 1))
+    else:
+        limit = str(max(get_available_cpu_count() - 1, 1))
+    for var in ["MKL_NUM_THREADS", "OMP_NUM_THREADS", "OPENBLAS_NUM_THREADS"]:
+        if var not in os.environ:
+            os.environ[var] = limit
```
arettig: We could have `MKL_NUM_THREADS` and `OPENBLAS_NUM_THREADS` default to the value of `OMP_NUM_THREADS` if that is defined, as I think BLAS libraries will typically default to that value if the library specific environment variable is not set (you might even be able to get away with only setting `OMP_NUM_THREADS` here in this case).

arettig (approved):
I can confirm that `check/pytest -n auto` drops from 55 minutes to 3 minutes on my machine with this PR!!

Some very minor suggestions below.

conftest.py
mhucka: > should we add `threadpoolctl` to our developer dependencies if it's required to prevent oversubscription?

Yes. I was going to do it in a separate PR because adding it means also updating the generated requirements .txt files, but maybe it better be part of this PR.

mhucka: Done in the next push.

src/openfermion/config.py
mhucka: I realized tonight that the "if not in os.environ" condition doesn't make sense if running with pytest-xdist, because if the value of one of those variables is set, they will not be adjusted for the number of cores being used by pytest-xdist. So for now, I made it set the values unconditionally. I'm not sure if that's great either, but I can't think of what else to do.

mhucka: @arettig Given this change and other changes I made in the latest push, it might be good for you to review them.

check/pytest-and-incremental-coverage
```diff
@@ -68,7 +78,7 @@ fi
 
 # Run tests while producing coverage files.
 check/pytest . \
-    -n auto \
+    "${pytest_args[@]}" \
```
arettig, line 81: `check/pytest-and-incremental-coverage` now runs the pytests in serial by default. Is this intentional? I suppose it matches the `check/pytest` script now.

check/all
```diff
@@ -51,53 +63,63 @@ function error() {
...
 
 function run() {
-    echo "Running $* ..."
+    echo "Running $*"
     "$@" || errors+=( "$* failed" )
-    echo
 }
 
+if [[ -z "${no_parallel}" ]]; then
+    pylint_j=("-j0")
```
arettig: `pylint` defaults to parallel execution due to .pylintrc so it will still run in parallel even if `--no-parallel` is passed.

arettig (approved):
Looks good! I can confirm it is still very fast on my machine.

A couple small questions about defaulting to parallel vs. serial below. From what I can tell `pytest` and `mypy` run in serial by default whereas `pylint`, and `format-incremental` run in parallel by default. (Of course this largely doesn't matter for anything except the pytests).

check/pytest-and-incremental-coverage
mhucka: In fact, yes, as a way to make the default be uniform for all the scripts. 

Now, a case could be made that the default should be parallel by default. Some points against doing that are:

- Since plain pytest is not parallel by default, I assume (admittedly without evidence) that users who know about pytest will _expect_ not-parallel to be the default.

- This way, the instructions can tell people to add the `-n` flag, just as they would with plain pytest.

- If we flipped the default, we'd have to invent a flag to let the user turn it off.

So, that was the rationale.

check/all
mhucka: Oh, argh, thank you for catching that! I'll fix that in the next push.

check/pytest-and-incremental-coverage
arettig: Good points, this seems reasonable to me then!

mhucka:
I thought that `mypy` didn't have a parallelism option, but now reading the docs, [yes, it does have an option for parallel workers](https://mypy.readthedocs.io/en/latest/command_line.html#parallel-type-checking). I'll adjust the check/all script to make use of that.

mhucka:
Regarding format-incremental: it uses `black`, which defaults to parallel execution.

It's tempting to also change it, to make it consistent (i.e., make it run serially by default), but because it doesn't run any code (unlike pytest), and because I'm not sure what benefit there would be, I'm leaning towards leaving it as-is (parallel).

PR #1406, opened 2026-07-10
Use `retry_once_with_later_random_values()` in isdf_test.py

mhucka:
The test `test_kpoint_isdf_symmetries()` will occasionally fail, and when it does, retrying won't work unless a different random seed is used.


arettig (approved):
LGTM

PR #1408, opened 2026-07-13
Update the installation instructions in all tutorial notebooks

mhucka:
This replaces the previous instructions with a simpler and more direct pip installation command.


mhucka:
This may be more troublesome than it's worth. Closing until it can be thought through.

PR #1416, opened 2026-07-15
Fix #1342: revert to using Markdown format for issue templates

### LINKED ISSUE #1342: Revert to using Markdown files instead of YAML for issue templates

I feel like the use of GitHub YAML-based issue forms is a failure. On paper, it seems like a good idea, and the web interface is nice enough, but:

- On the command-line, `gh issue create` does not handle them -- in fact, it doesn't even know they exist. Conversely, Markdown-based issue templates work fine, including when you open an external editor.

- There was a hope that maybe GitHub would add proper support for YAML-based issue templates in `gh` eventually, but that still has not happened.

- The YAML version of templates is harder and more error-prone to edit and maintain.

Mea culpa because the push for using the YAML templates in the first place was my idea, but this is not worth it. Let's go back to using Markdown templates.

We have a starting point in the (private) template directory in Quantumlib, so the switch will not take long.

None

mhucka:
GitHub issue forms _seemed_ like a good idea, and the web interface is nice. After using them for some time, though, they turn out to introduce too much friction:

- On the command-line, `gh issue create` does not handle them – in fact, _it doesn't even know they exist_! Conversely, Markdown-based issue templates work fine using `gh`, including when you open an external editor.

- There was a hope that maybe GitHub would add proper support for YAML-based issue templates in `gh` eventually, but that still has not happened.

- The YAML version of templates is harder and more error-prone to edit and maintain.

This converts the forms back to GitHub's Markdown format for issue templates.


.github/ISSUE_TEMPLATE/1-bug-report.md
```diff
@@ -0,0 +1,37 @@
...
+### What version of OpenFermion are you using?
+<!--
+What version does this happen in? Write the version number below.
+-->
+
+### What computing environment are you using?
+<!--
+Please tell us what kind of computing system you are using (e.g.,
+Linux, Mac, Windows, Google Colab, Google Cloud, etc.) as well as the
+the version, and if relevant, any special hardware features it has,
```
arettig: nit
```suggestion
Linux, Mac, Windows, Google Colab, Google Cloud, etc.) as well as
the version, and if relevant, any special hardware features it has,
```

arettig (approved):
one typo, otherwise LGTM

Issue #1407, opened 2026-07-13
Operator +, - and scalar * spend nearly all their time in an avoidable deepcopy


blazingphoenix7:
Found this while profiling operator arithmetic on large QubitOperators.

`SymbolicOperator.__add__`, `__sub__`, `__mul__` and `accumulate` all copy the
operand with `copy.deepcopy` and then mutate the copy in place. But a term key
is a tuple of `(index, action)` tuples and a coefficient is an int, float,
complex or sympy expression, all immutable, so the recursive copy buys nothing:
the in-place operators only rebind `self.terms[term]` and add or remove keys,
they never mutate a key or coefficient object. A plain `dict(self.terms)` gives
the same out-of-place guarantee.

The gap is large. Random QubitOperators over 30 qubits:

    terms    op         main      patched
     5000    A + B      85.8 ms   2.5 ms
    20000    A + B       348 ms   11.8 ms
    20000    2.0 * A     339 ms   4.7 ms

Medians over 25 repetitions, two alternated main/patch rounds with fixed seeds,
on an Intel Core Ultra 5 225 laptop.

I have a patch ready: a small `_clone` helper (empty instance plus a copy of
the terms dict) used in the four call sites. The existing out-of-place tests
pass, black and mypy are clean, and every changed line is covered by the
current suite.

Opening a pull request currently appears to be limited to collaborators, so I'm
filing this instead; happy to send the patch over if wanted. It came out of a
wider look at the operator hot paths, and I have a couple more findings queued.

mhucka:
@blazingphoenix7 Thank you for this issue report, diagnosis, and offer to work on a PR. Very sorry about the limitation on pull requests; I had set it that way not long ago to stop some LLM-produced slop at the time, and then (somewhat predictably) forgot to unset the limitation after a time.

Pull requests should be enabled for everyone now, so you should be able to open a PR if you're still interested.

blazingphoenix7:
Thanks for flipping it back. Opened #1415.

blazingphoenix7:
Fixed by #1415, which landed on the 17th. `+`, `-` and scalar `*` no longer deepcopy the operand. Thanks again for re-opening PRs to outside contributors. Closing.

Issue #1405, opened 2026-07-09
Deadlock possibility in linear_qubit_operator.py


mhucka:
Gemini Code Assist reports a problem with the multiprocessing code in linear_qubit_operator.py:

**Deadlock in Multi-Process Path (Lines 192-198):**

Currently, `pool.join()` is called *before* consuming the `vecs` iterator returned by `pool.imap_unordered`. Since `imap_unordered` yields results lazily, the worker processes write their results to an internal OS pipe. If the state vector `x` is large (which is very common in quantum simulations), the pipe buffer (typically 64KB on Linux) will fill up, causing the worker processes to block on `write()`. Meanwhile, the main process is blocked on `pool.join()`, resulting in a permanent deadlock.
   
**Fix:** 

You must consume the iterator completely (e.g., by performing the reduction) *before* calling `pool.join()`:
   
   ```python
   pool = self.options.get_pool(len(self.linear_operators))
   vecs = pool.imap_unordered(
       apply_operator, [(operator, x) for operator in self.linear_operators]
   )
   pool.close()
   # Consume the iterator BEFORE joining the pool to avoid deadlocks
   result = functools.reduce(numpy.add, vecs)
   pool.join()
   return result
   ```

_Originally posted by @gemini-code-assist[bot] in https://github.com/quantumlib/OpenFermion/pull/1404#discussion_r3553904905_

rosspeili:
@mhucka happy to pick this up.

mhucka:
@rosspeili Thanks!

rosspeili:
@mhucka open a small pr, lmk how this looks.

Issue #499, opened 2019-01-07
Improve LinearQubitOperator by porting functionality from Cirq


kevinsung:
Currently, we have a function `generate_linear_qubit_operator` in `utils/_linear_qubit_operator.py` which takes a QubitOperator and outputs a LinearQubitOperator or ParallelLinearQubitOperator (subclasses of scipy.sparse.linalg.LinearOperator) which can be used, for instance, in scipy.sparse eigenvalue routines. This class can be made a lot more useful by improving the implementation of `LinearQubitOperator._matvec`.

The code below uses [Cirq](https://github.com/quantumlib/Cirq) and constitutes a functional implementation of `LinearQubitOperator._matvec`:
```
def apply_qubit_operator_to_state(qubit_op, state):
    num_qubits = state.shape[0].bit_length() - 1
    result = np.zeros(state.shape, dtype=state.dtype)
    for term, coeff in qubit_op.terms.items():
        mutated_state = np.reshape(np.copy(state), (2,) * num_qubits)
        for index, action in term:
            if action == 'X':
                pauli = cirq.X
            elif action == 'Y':
                pauli = cirq.Y
            else:
                pauli = cirq.Z
            buffer = np.empty(mutated_state.shape, dtype=state.dtype)
            args = cirq.ApplyUnitaryArgs(target_tensor=mutated_state,
                                         available_buffer=buffer,
                                         axes=(index,))
            mutated_state = cirq.apply_unitary(pauli, args)
        result += coeff*np.reshape(mutated_state, state.shape)
    return result
```
I found that this implementation gives an 80-100x speedup at large (at least 16 qubits) system sizes with smaller but still significant gains at smaller ones.

The task for this issue is basically to port this code over into an implementation of `LinearQubitOperator._matvec`. The class ParallelLinearQubitOperator can then be deleted since it would be obsolete.

PR #1417, opened 2026-07-15
Speed up LinearQubitOperator matvec

blazingphoenix7:
Second of the operator hot-path findings from #1407; the first is #1415.

`LinearQubitOperator._matvec` applies a `QubitOperator` to a state vector, and it
is the inner loop behind `generate_linear_qubit_operator` and the eigensolvers
built on top of it. For each term it took the input vector, split it into `2**k`
pieces to reach the first non-identity qubit, split each of those in half, applied
the 2x2 Pauli to the pairs, and concatenated everything back, once per Pauli in
the term. That is a lot of Python-level list building and array copying for what
is, per term, a single reindexing of the amplitudes.

A Pauli string acts on the state vector as a signed permutation, so it can be
written directly on the flat index. Number the amplitudes `0 .. 2**n - 1` and put
qubit `q` at bit `n - 1 - q` (the ordering the split version already used, and the
one `qubit_operator_sparse` uses). Then for one term:

- the X and Y qubits flip their bit, so amplitude `c` moves to `c ^ x_mask`;
- the Y and Z qubits give a factor of -1 whenever the matching bit of `c` is set,
  that is `(-1)` raised to the parity of `popcount(c & z_mask)`;
- each Y gives an extra factor of `1j`.

So the term becomes `out[idx ^ x_mask] += coeff * 1j**(y_count % 4) * sign * x`,
where `sign` is a vector of +-1 from the bit parity and `idx ^ x_mask` is a
permutation of the indices, which is what makes the fancy-indexed `+=` safe. The
masks are cheap integer folds over the term and the per-amplitude work is a few
vectorized numpy operations with no intermediate lists.

This is an implementation change only. Public behavior is unchanged: complex
output, inputs of shape `(dim,)` or `(dim, 1)`, the identity term, operators
declared on more qubits than they act on, and the empty operator all behave as
before. Only the serial `_matvec` is touched; `ParallelLinearQubitOperator` is
left alone.

The parity of `popcount` uses a small xor-fold helper on uint64 rather than
`numpy.bitwise_count`, because the project still supports numpy 1.26 (the
max_compat CI job pins `numpy==1.26.4`) and `bitwise_count` needs numpy 2.0.

### Correctness

`linear_qubit_operator_test.py` passes unchanged. On top of that I compared the
new matvec against `qubit_operator_sparse(op) @ x` over random operators for
`n = 2..12` (mixed X/Y/Z, term counts including the empty and identity-only
operators), real and complex `x`, shapes `(dim,)` and `(dim, 1)`, and `n_qubits`
oversized by 2, plus a `scipy.sparse.linalg.eigsh` run checked against dense
eigenvalues. Worst absolute differences:

    random real x, shape (dim,)          5.4e-15
    random complex x, shape (dim,)       9.1e-15
    random complex x, shape (dim, 1)     9.1e-15
    oversized n_qubits by 2              2.4e-15
    identity-only and empty operator     0
    eigsh vs dense (3 smallest)          8.9e-15

black and mypy are clean on the file.

### Motivation

This is what #499 asked for. That issue called for a better `_matvec`, measured
that one was worth 80-100x at system sizes of 16 qubits and up, and noted that
`ParallelLinearQubitOperator` could then be deleted as obsolete once the serial
path was fast. The sketch there routed each Pauli through Cirq's
`apply_unitary`; no port ever landed and the issue was closed for inactivity in
2025. This PR gets the same win with plain numpy on the flat index, so it adds
no dependency, and the 120x at 16 qubits below bears out the issue's estimate.

The parallel path is also where the deadlock possibility in #1405 lives, and
with the serial matvec this fast there is rarely a reason to reach for it.
Removing it, as #499 suggested, would be a separate change; this PR fixes
neither issue.

### Benchmarks

Random QubitOperators with mixed X/Y/Z terms applied to a complex state vector:

    n_qubits   terms    before     after    speedup
       16       100     20.5 s     0.17 s      120x
       18       100     87.0 s     1.5 s        60x
       20        50      157 s     3.0 s        52x

Measured on my laptop (Intel Core Ultra 5 225, Windows, Python 3.12, numpy 2.5).
I ran the base/patch comparison twice in alternation with fixed seeds, taking
medians over three runs at n = 16 and single runs at n = 18 and 20 since the old
code takes minutes there; the per-round speedups agreed within a few percent, so
the table pools both rounds.

### DIFF
--- a/src/openfermion/linalg/linear_qubit_operator.py
+++ b/src/openfermion/linalg/linear_qubit_operator.py
@@ -115,7 +115,7 @@ class LinearQubitOperator(scipy.sparse.linalg.LinearOperator):
         """
         arr = numpy.asarray(x)
         vec = arr.reshape(-1)
-        indices = numpy.arange(vec.size, dtype=numpy.uint64)
+        indices = None
         retvec = numpy.zeros(vec.size, dtype=complex)
 
         for qubit_term, coefficient in self.qubit_operator.terms.items():
@@ -133,9 +133,12 @@ class LinearQubitOperator(scipy.sparse.linalg.LinearOperator):
                 else:
                     z_mask ^= bit
 
-            amplitudes = coefficient * [1, 1j, -1, -1j][y_count % 4] * vec
+            amplitudes = coefficient * (1, 1j, -1, -1j)[y_count % 4] * vec
+            if z_mask or x_mask:
+                if indices is None:
+                    indices = numpy.arange(vec.size, dtype=numpy.uint64)
             if z_mask:
-                signs = 1 - 2 * _bit_parity(indices & numpy.uint64(z_mask)).astype(numpy.int64)
+                signs = 1 - 2 * _bit_parity(indices & numpy.uint64(z_mask)).astype(numpy.int8)
                 amplitudes = signs * amplitudes
             if x_mask:
                 retvec[indices ^ numpy.uint64(x_mask)] += amplitudes
--- a/src/openfermion/linalg/linear_qubit_operator.py
+++ b/src/openfermion/linalg/linear_qubit_operator.py
@@ -24,6 +24,17 @@ from openfermion.utils.operator_utils import count_qubits
 from openfermion.config import get_available_cpu_count
 
 
+def _bit_parity(values):
+    """Returns the parity of the population count of each uint64 value."""
+    values = values ^ (values >> numpy.uint64(32))
+    values = values ^ (values >> numpy.uint64(16))
+    values = values ^ (values >> numpy.uint64(8))
+    values = values ^ (values >> numpy.uint64(4))
+    values = values ^ (values >> numpy.uint64(2))
+    values = values ^ (values >> numpy.uint64(1))
+    return values & numpy.uint64(1)
+
+
 class LinearQubitOperatorOptions:
     """Options for LinearQubitOperator."""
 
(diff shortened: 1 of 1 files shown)


src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -108,38 +113,36 @@ def _matvec(self, x):
...
+                if action == 'X':
+                    x_mask ^= bit
+                elif action == 'Y':
+                    x_mask ^= bit
+                    z_mask ^= bit
+                    y_count += 1
+                else:
+                    z_mask ^= bit
+
+            amplitudes = coefficient * (1j ** (y_count % 4)) * vec
```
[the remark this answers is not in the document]
blazingphoenix7: Thanks, took the table. One note for the thread: `1j ** (y_count % 4)` is actually already exact here since the exponent is a Python int, so CPython evaluates it by repeated multiplication rather than the polar form the example uses. The table makes that explicit though, and keeps it exact if the exponent ever comes through as a numpy scalar.

src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -108,38 +113,36 @@ def _matvec(self, x):
...
-                }
-                vecs = [v for vp in xyz[pauli_operator[1]](vec_pairs) for v in vp]
-                tensor_factor = pauli_operator[0] + 1
-
-            # No need to check tensor_factor, i.e. to deal with bits left.
-            retvec += coefficient * numpy.concatenate(vecs)
-        return retvec
+        arr = numpy.asarray(x)
+        vec = arr.reshape(-1)
+        indices = numpy.arange(vec.size, dtype=numpy.uint64)
```
mhucka: Could this array allocation be done in a lazy manner, for greater memory efficiency? Something like

```python
indices = None
for qubit_term, coefficient in self.qubit_operator.terms.items():
    # ...
    if z_mask or x_mask:
        if indices is None:
            indices = numpy.arange(vec.size, dtype=numpy.uint64)
```

src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -108,38 +113,36 @@ def _matvec(self, x):
...
+                elif action == 'Y':
+                    x_mask ^= bit
+                    z_mask ^= bit
+                    y_count += 1
+                else:
+                    z_mask ^= bit
+
+            amplitudes = coefficient * [1, 1j, -1, -1j][y_count % 4] * vec
+            if z_mask:
+                signs = 1 - 2 * _bit_parity(indices & numpy.uint64(z_mask)).astype(numpy.int64)
```
mhucka: Since the values of the parity array are strictly 0 or 1, is the use of a signed integer necessary? Reducing the memory footprint should help improve CPU cache locality for these arithmetic operations.

```suggestion
                signs = 1 - 2 * _bit_parity(indices & numpy.uint64(z_mask)).astype(numpy.int8)
```

src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -108,38 +113,36 @@ def _matvec(self, x):
...
+                if action == 'X':
+                    x_mask ^= bit
+                elif action == 'Y':
+                    x_mask ^= bit
+                    z_mask ^= bit
+                    y_count += 1
+                else:
+                    z_mask ^= bit
+
+            amplitudes = coefficient * [1, 1j, -1, -1j][y_count % 4] * vec
```
mhucka: Gemini Code Assist's suggestion used a list, which is constructed dynamically, but for something static like this, it could be a tuple instead, which will be static and a tiny bit faster:

```suggestion
            amplitudes = coefficient * (1, 1j, -1, -1j)[y_count % 4] * vec
```

mhucka (changes requested):
Thank you for this nice work! I had just a few suggestions.

src/openfermion/linalg/linear_qubit_operator.py
blazingphoenix7: Good point, switched to the tuple.

src/openfermion/linalg/linear_qubit_operator.py
blazingphoenix7: Done. The values are only +-1, so int8 is plenty and saves the memory.

src/openfermion/linalg/linear_qubit_operator.py
blazingphoenix7: Done. Made it lazy so an operator that is only the identity term skips the allocation.

mhucka (approved):
Thanks for the updates. Looks good.

PR #1418, opened 2026-07-15
Restore previous behavior to PolynomialTensor.rotate_basis() by default

arettig:
PolynomialTensor's rotate_basis method previously used the transpose of the rotation matrix. This is not correct, but changing it in #805 broke users' code. A `transpose` flag is introduced to specify whether or not to transpose the rotation matrix, with `True` being the default for now (to be changed to the preferred `False` in a later release). A deprecation warning is given if the transpose flag is not specified.

### DIFF
--- a/src/openfermion/ops/representations/interaction_rdm.py
+++ b/src/openfermion/ops/representations/interaction_rdm.py
@@ -61,7 +61,7 @@ class InteractionRDM(PolynomialTensor):
         """Set the value of the two-body tensor."""
         self.n_body_tensors[1, 1, 0, 0] = value
 
-    def rotate_basis(self, rotation_matrix):
+    def rotate_basis(self, rotation_matrix, transpose=None):
         """Rotate the orbital basis of the InteractionRDM.
 
         Note that rotating the basis of an RDM M via some rotation uses the
@@ -75,8 +75,11 @@ class InteractionRDM(PolynomialTensor):
         Args:
             rotation_matrix: A square numpy array or matrix having
                 dimensions of n_qubits by n_qubits. Assumed to be unitary.
+            transpose: If True, transposes the rotation matrix before applying it.
+                If None (default), behaves as True to maintain backwards
+                compatibility, but raises a DeprecationWarning.
         """
-        super().rotate_basis(numpy.conjugate(rotation_matrix))
+        super().rotate_basis(numpy.conjugate(rotation_matrix), transpose=transpose)
 
     def expectation(self, operator):
         """Return expectation value of an InteractionRDM with an operator.
--- a/src/openfermion/ops/representations/interaction_rdm.py
+++ b/src/openfermion/ops/representations/interaction_rdm.py
@@ -12,6 +12,7 @@
 """Class and functions to store reduced density matrices."""
 
 import copy
+import warnings
 import numpy
 
 from openfermion.ops.operators import FermionOperator, QubitOperator
@@ -61,7 +62,7 @@ class InteractionRDM(PolynomialTensor):
         """Set the value of the two-body tensor."""
         self.n_body_tensors[1, 1, 0, 0] = value
 
-    def rotate_basis(self, rotation_matrix, transpose=None):
+    def rotate_basis(self, rotation_matrix: numpy.ndarray, transpose: bool | None = None) -> None:
         """Rotate the orbital basis of the InteractionRDM.
 
         Note that rotating the basis of an RDM M via some rotation uses the
@@ -76,9 +77,20 @@ class InteractionRDM(PolynomialTensor):
             rotation_matrix: A square numpy array or matrix having
                 dimensions of n_qubits by n_qubits. Assumed to be unitary.
             transpose: If True, transposes the rotation matrix before applying it.
+                If False, the rotation matrix is not transposed.
                 If None (default), behaves as True to maintain backwards
-                compatibility, but raises a DeprecationWarning.
+                compatibility, but raises a FutureWarning.
         """
+        if transpose is None:
+            warnings.warn(
+                "InteractionRDM.rotate_basis was called without specifying the 'transpose' "
+                "parameter. The current default is transpose=True to maintain backwards "
+                "compatibility, but this will change to transpose=False in a future release. "
+                "Specify the value of 'transpose' explicitly to silence this warning.",
+                FutureWarning,
+                stacklevel=2,
+            )
+            transpose = True
         super().rotate_basis(numpy.conjugate(rotation_matrix), transpose=transpose)
 
     def expectation(self, operator):
--- a/src/openfermion/ops/representations/polynomial_tensor.py
+++ b/src/openfermion/ops/representations/polynomial_tensor.py
@@ -14,6 +14,7 @@ fermionic ladder operators."""
 
 import copy
 import itertools
+import warnings
 import operator
 
 import numpy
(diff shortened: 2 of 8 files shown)


src/openfermion/ops/representations/interaction_rdm.py
```diff
@@ -75,8 +75,11 @@ def rotate_basis(self, rotation_matrix):
         Args:
             rotation_matrix: A square numpy array or matrix having
                 dimensions of n_qubits by n_qubits. Assumed to be unitary.
+            transpose: If True, transposes the rotation matrix before applying it.
```
mhucka, line 79: Probably worth adding an "if False" clause to explain that the matrix is left as-is.

When I read this docstring the first time through, I instinctively expected to read about the False case, and actually misread the "If None" sentence as being "If False". Having all 3 cases explicit may help people to avoid misreading it.

src/openfermion/ops/representations/polynomial_tensor.py
```diff
@@ -360,15 +361,30 @@ def __str__(self):
...
                 dimensions of n_qubits by n_qubits. Assumed to be real and
                 invertible.
+            transpose: If True, transposes the rotation matrix before applying it.
+                If None (default), behaves as True to maintain backwards
+                compatibility, but raises a DeprecationWarning.
         """
+        if transpose is None:
+            warnings.warn(
+                "rotate_basis(R) is deprecated, please use rotate_basis(R, transpose). "
+                "tranpose=True will be used currently to maintain backwards compatiblity, "
```
mhucka:
```suggestion
                "transpose=True is the current default to maintain backwards compatibility, "
```

src/openfermion/ops/representations/polynomial_tensor.py
```diff
@@ -360,15 +361,30 @@ def __str__(self):
...
+                "but the default will change to transpose=False a future release",
```
mhucka:
```suggestion
                "but the default will change to transpose=False in a future release",
```

src/openfermion/ops/representations/polynomial_tensor.py
```diff
@@ -360,15 +361,30 @@ def __str__(self):
             strings.append('{} {}\n'.format(key, self[key]))
         return ''.join(strings) if strings else '0'
 
-    def rotate_basis(self, rotation_matrix):
+    def rotate_basis(self, rotation_matrix, transpose=None):
```
mhucka: Let's also add the missing type hints while we're at it:

```suggestion
    def rotate_basis(
        self, 
        rotation_matrix: numpy.ndarray, 
        transpose: bool | None = None
    ) -> None:
```

src/openfermion/ops/representations/interaction_rdm.py
```diff
@@ -61,7 +61,7 @@ def two_body_tensor(self, value):
         """Set the value of the two-body tensor."""
         self.n_body_tensors[1, 1, 0, 0] = value
 
-    def rotate_basis(self, rotation_matrix):
+    def rotate_basis(self, rotation_matrix, transpose=None):
```
mhucka: Let's also add missing type hints while we're at it:

```suggestion
    def rotate_basis(
        self, 
        rotation_matrix: numpy.ndarray, 
        transpose: bool | None = None
    ) -> None:
```

src/openfermion/ops/representations/polynomial_tensor.py
```diff
@@ -360,15 +361,30 @@ def __str__(self):
...
+            transpose: If True, transposes the rotation matrix before applying it.
+                If None (default), behaves as True to maintain backwards
+                compatibility, but raises a DeprecationWarning.
         """
+        if transpose is None:
+            warnings.warn(
+                "rotate_basis(R) is deprecated, please use rotate_basis(R, transpose). "
+                "tranpose=True will be used currently to maintain backwards compatiblity, "
+                "but the default will change to transpose=False a future release",
+                DeprecationWarning,
```
mhucka: This may need a `stacklevel` argument to point to the user's code where the call to rotate_basis is made, rather than pointing inside _this_ function.

Gemini tells me `stacklevel=2` is needed.

mhucka: And also, this is actually a case for FutureWarning rather than DeprecationWarning.

src/openfermion/ops/representations/polynomial_tensor.py
```diff
@@ -360,15 +361,30 @@ def __str__(self):
...
             rotation_matrix: A square numpy array or matrix having
                 dimensions of n_qubits by n_qubits. Assumed to be real and
                 invertible.
+            transpose: If True, transposes the rotation matrix before applying it.
+                If None (default), behaves as True to maintain backwards
+                compatibility, but raises a DeprecationWarning.
         """
+        if transpose is None:
+            warnings.warn(
+                "rotate_basis(R) is deprecated, please use rotate_basis(R, transpose). "
```
mhucka: Grammar fix for comma splice:

```suggestion
                "rotate_basis(R) is deprecated. Please use rotate_basis(R, transpose). "
```

src/openfermion/ops/representations/polynomial_tensor_test.py
```diff
@@ -537,6 +537,47 @@ def do_rotate_basis_quadratic_hamiltonian(self, real):
...
+        # Check that the orbital energies and constant are the same
+        new_orbital_energies, _, new_constant = quad_ham.diagonalizing_bogoliubov_transform()
+
+        self.assertTrue(numpy.allclose(orbital_energies, new_orbital_energies))
+        self.assertAlmostEqual(constant, new_constant)
+
+    def test_rotate_basis_warning(self):
+        n_qubits = 2
+        polynomial_tensor = PolynomialTensor({(1, 0): numpy.identity(n_qubits)})
+        with self.assertWarns(DeprecationWarning):
```
mhucka: If the other change is made, switch this test to FutureWarning too.

mhucka (changes requested):
Thanks for doing this so quickly. Here are some minor change requests. Overall looks good!

src/openfermion/ops/representations/interaction_rdm.py
arettig: Great point, thanks for the suggestion! Fixed.

src/openfermion/ops/representations/polynomial_tensor.py
arettig: Done.

src/openfermion/ops/representations/interaction_rdm.py
arettig: Done.

src/openfermion/ops/representations/polynomial_tensor.py
arettig: I added `stacklevel=2` and also raised the warning in `InteractionRDM.rotate_basis` so both cases will now point correctly to the calling code.

arettig: Good point, I changed the type to `FutureWarning` and adjusted the wording to reflect that.

src/openfermion/ops/representations/polynomial_tensor.py
arettig: done.

src/openfermion/ops/representations/polynomial_tensor_test.py
arettig: Done.

src/openfermion/ops/representations/polynomial_tensor.py
arettig: Done.

src/openfermion/ops/representations/polynomial_tensor.py
arettig: Done.

mhucka (approved):
LGTM!

mhucka (approved):
Looks great! Thanks for doing this.

PR #1419, opened 2026-07-16
Fix #1405: potential pool.join deadlock in ParallelLinearQubitOperator 

### LINKED ISSUE #1405: Deadlock possibility in linear_qubit_operator.py

rosspeili:
Consume imap_unordered results before joining the worker pool to avoid pipe backpressure deadlocks when state vectors are large.

Fixes #1405

### DIFF
--- a/src/openfermion/linalg/linear_qubit_operator.py
+++ b/src/openfermion/linalg/linear_qubit_operator.py
@@ -194,8 +194,11 @@ class ParallelLinearQubitOperator(scipy.sparse.linalg.LinearOperator):
             apply_operator, [(operator, x) for operator in self.linear_operators]
         )
         pool.close()
+        # Consume results before join(): imap_unordered uses a bounded pipe and
+        # workers block on write if the main process has not read them yet.
+        result = functools.reduce(numpy.add, vecs)
         pool.join()
-        return functools.reduce(numpy.add, vecs)
+        return result
 
 
 def apply_operator(args):
--- a/src/openfermion/linalg/linear_qubit_operator.py
+++ b/src/openfermion/linalg/linear_qubit_operator.py
@@ -196,10 +196,10 @@ class ParallelLinearQubitOperator(scipy.sparse.linalg.LinearOperator):
         pool.close()
         # Consume results before join(): imap_unordered uses a bounded pipe and
         # workers block on write if the main process has not read them yet.
-    try:
-        result = functools.reduce(numpy.add, vecs)
-    finally:
-        pool.join()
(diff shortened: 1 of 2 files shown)


src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -194,8 +194,11 @@ def _matvec(self, x):
             apply_operator, [(operator, x) for operator in self.linear_operators]
         )
         pool.close()
+        # Consume results before join(): imap_unordered uses a bounded pipe and
+        # workers block on write if the main process has not read them yet.
+        result = functools.reduce(numpy.add, vecs)
         pool.join()
```
mhucka: `reduce` can throw an exception. Might be best to guard against this:

```suggestion
    try:
        result = functools.reduce(numpy.add, vecs)
    finally:
        pool.join()
```

mhucka (changes requested):
Nice! Thanks for this. One minor suggestion.

rosspeili:
@mhucka applied suggestion but ci failed on black cause gh apply dropped the try/finally indentation and moved return result into the finally block. Pushed a fix that keeps your guard (pool.join() always runs) with correct indentation and return after the try/finally. Thanks for the hint <3

mhucka (approved):
Awesome. Thank you!

PR #1421, opened 2026-07-16
Support using flag `--skipslow` instead of `-m "not slow"` for pytest

mhucka:
The flag `-m "not slow"` was not only annoying to type: it invited misspelling and waste of time. 

This PR changes the flag to be `--skipslow`, following the same thing done in the ReCirq project.


conftest.py
```diff
@@ -66,3 +66,12 @@ def set_threadpool_limits():
...
         yield
+
+
+def pytest_addoption(parser: pytest.Parser) -> None:
+    parser.addoption("--skipslow", action="store_true", help="skips slow tests")
+
+
+def pytest_runtest_setup(item: pytest.Item) -> None:
+    if "slow" in item.keywords and item.config.getoption("skipslow"):
+        pytest.skip("skipped because of --skipslow option")
```
pavoljuhas, line 77: This reports `conftest.py` as the source file for the skipped tests:
```shell
$ pytest -rs --skipslow src/openfermion/ops/representations/doci_hamiltonian_test.py::IntegralTransformsTest::test_fermionic_hamiltonian_from_integrals
...
================================================== short test summary info ===================================================
SKIPPED [1] conftest.py:77: skipped because of --skipslow option
===================================================== 1 skipped in 0.06s =====================================================
```



I think it is better to use the `pytest_collection_modifyitems` hook as in [cirq](https://github.com/quantumlib/Cirq/blob/7a66e618c09de70313bd968c2f8e60604ccfeae2/conftest.py#L24) which reports the actual test skipped:

```shell
$ pytest -rs dev_tools/notebooks/notebook_test.py
...
================================================== short test summary info ===================================================
SKIPPED [59] dev_tools/notebooks/notebook_test.py:101: need --enable-slow-tests option to run
=============================================== 2 passed, 59 skipped in 0.17s ================================================

```

src/openfermion/config_test.py
```diff
@@ -170,3 +170,50 @@ def test_set_threading_limits(self):
...
+
+    def test_pytest_addoption(self):
+        import conftest
+        from unittest.mock import MagicMock
+
+        parser = MagicMock()
+        conftest.pytest_addoption(parser)
+        parser.addoption.assert_called_once_with(
+            "--skipslow", action="store_true", help="skips slow tests"
+        )
```
pavoljuhas, line 185: There seems to be too much mockery and AI generated crud for this test to be meaningful.

It verifies that conftest hook functions do what they do on their arguments, but that tells nothing of if the hooks are used in a pytest session and if they have  desired effects.

I suggest to delete this; it is a second order test-of-a-test-code anyway.

pavoljuhas (changes requested):
Please remove the LLM code-salad ConftestTest.

The second comment to use the `pytest_collection_modifyitems` is optional, feel free to leave it as is with a bit less accurate skip report.

Otherwise LGTM.

Issue #587, opened 2020-04-18
Parallelization of Jordan-Wigner transformation


Emieeel:
For systems of >10 orbitals, I noticed that a jordan wigner transformation on the fermionic hamiltonian is quite expensive, probably because it is pure python code so it runs on just 1 core. Is there any way of improving the speed of the get_jordan_wigner() function and/or parallelize it?

babbush:
Probably. It would be nice to have faster code for this. The runtime should be roughly linear in the number of terms in the Hamiltonian. Of course, for large chemistry Hamiltonians with O(N^4) terms, that's not very good. But for the case of molecules there is a special routine that maps the InteractionOperators directly to qubits without going through the fermion Hamiltonian as an intermediate, and that is faster. Feel free to suggest a better implementation for the jordan wigner transform applied to the FermionOperator data structure.

kevinsung:
I think at some point we might want to implement the basic transforms in Cython.

Issue #121, opened 2017-12-10
jordan_wigner_interaction_op is surprisingly slow


kevinsung:
In the code for `jordan_wigner`, InteractionOperators are handled with the helper function `jordan_wigner_interaction_op`, rather than simply mapping to a FermionOperator first and then transforming with the existing code. This is supposedly for the sake of speed. However, I found that doing it this way on a Hubbard Hamiltonian is apparently around 10 times slower. Does anyone have an idea why? Was this actually tested for speed? Should we just ditch the helper function?
```
In [1]: from openfermion.hamiltonians import fermi_hubbard

In [2]: from openfermion.transforms import jordan_wigner, get_fermion_operator, get_interaction_operator

In [3]: hubbard_model = fermi_hubbard(4, 4, 1., 4., 2., 2.)

In [4]: iop = get_interaction_operator(hubbard_model)

In [5]: def jordan_wigner_iop(iop):
   ...:     ferm_op = get_fermion_operator(iop)
   ...:     qubit_op = jordan_wigner(ferm_op)
   ...:     return qubit_op
   ...: 

In [6]: %timeit -n 3 -r 7 qubit_op_1 = jordan_wigner_iop(iop)
250 ms ± 5.32 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)

In [7]: %timeit -n 3 -r 7 qubit_op_2 = jordan_wigner(iop)
2.33 s ± 96.4 ms per loop (mean ± std. dev. of 7 runs, 3 loops each)
```

kevinsung:
Actually, it seems that the currently implementation is faster for random InteractionOperators and those derived from molecular Hamiltonians, probably closer to what it would actually be used for. So probably this isn't a real issue.

PR #1427, opened 2026-07-17
Avoid building per-term QubitOperators in jordan_wigner interaction transform

blazingphoenix7:
Third of the operator hot-path findings from #1407; the earlier two are #1415 and #1417.

`jordan_wigner` on an `InteractionOperator` maps the one- and two-body integrals
straight to Pauli terms instead of going through a `FermionOperator` first, and
#587 points at exactly this routine as the fast path to build on ("there is a
special routine that maps the InteractionOperators directly to qubits ... and
that is faster"). #121 is the other side of it: it flagged this same routine as
"surprisingly slow" back in 2017 and got closed without a fix.

The transform walks every one-body pair and every pair of pairs, so the two-body
part is O(N^4). For each Pauli string it built a fresh single-term
`QubitOperator` and summed that into the running total. Each of those
constructions parses the term, validates every `(index, action)` factor, runs the
Pauli `_simplify`, and in the three-unique-index case multiplies by a `Z`
operator, which deepcopies. None of that is needed to add a term to a sum: a
`QubitOperator`'s `terms` is just a dict from a sorted tuple of `(index, action)`
factors to a coefficient, and the strings this routine emits are already sorted
with their factors already combined, so there is nothing for the constructor to
normalize.

This accumulates the `pauli_string -> coefficient` pairs into a plain dict with a
small `_add_term` helper that does what `__iadd__` / `__isub__` do to a single
term (add onto the running value, drop it once it falls under `EQ_TOLERANCE`), and
builds one `QubitOperator` at the end by assigning the finished dict to `.terms`.
Accumulating into a dict instead of building and summing single-term operators
avoids that per-term construction and validation, and it measures about 2x
faster, with identical output.

### Public functions

`jordan_wigner_one_body` and `jordan_wigner_two_body` are public and re-exported,
and `bksf_test` and `jordan_wigner_test` call them directly, so they keep their
signatures and their output. I pulled the term-building logic into two private
helpers (`_one_body_terms`, `_two_body_terms`) that return the dict, and the
public functions and the interaction-op driver both go through them, so there is
one copy of the logic and the public results are unchanged. The `Z` times
hopping-term product in the three-unique case is a single-qubit relabeling (the
shared qubit carries at most a parity `Z`, and `Z` squared is the identity), so it
is a direct toggle of that index in the string now rather than an operator
multiply and deepcopy.

### Correctness

Before editing the source I captured the current `jordan_wigner` output on 104
random real `InteractionOperator`s and saved it. The mix is physical
(symmetric, molecular-style) operators, dense generic real tensors, sparse and
very sparse tensors, one-body-only, constant-only, and the exact-zero operator,
at N = 1, 2, 3, 4, 6, 8, 10, 12. The new code reproduces that captured output
exactly on every case: identical term-key sets, max absolute coefficient
difference 0, and `(new - old).induced_norm(1)` equal to 0. It also matches an
independent dict-based reference I wrote separately, to the same 0. Seeds are
fixed, so the check is deterministic.

`jordan_wigner_test.py` and `bksf_test.py` pass (the two public helpers are
checked there against the `FermionOperator` path over a full index grid with
complex coefficients), and `transforms/` passes as a whole. black and mypy are
clean on the file. Every changed line is hit by the existing suite except the
pre-existing `n_qubits < count_qubits(iop)` guard, which the old code did not
cover either and which I left untouched.

The routine is still documented as real-only and this does not change that. It
computes the same coefficients from the same integrals; only the container they
land in changed.

### Benchmarks

Random real InteractionOperators, cumulative `jordan_wigner` time. Medians over 5
repeats within each of 6 alternated upstream/branch rounds, fixed seeds, on an
Intel Core Ultra 5 225 laptop (Windows, Python 3.12, numpy 2.5).

    N     upstream    ours       speedup
    8     16 ms       7.6 ms     2.1x
    12    101 ms      48 ms      2.1x
    16    359 ms      177 ms     2.0x
    20    963 ms      476 ms     2.0x
    24    2222 ms     1093 ms    2.0x


src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -129,6 +129,38 @@ def _jordan_wigner_diagonal_coulomb_hamiltonian(operator):
...
+    if QubitOperator._issmall(value):
+        terms.pop(key, None)
+    else:
+        terms[key] = value
+
+
+def _merge_terms(terms, contribution):
+    """Add every term of a contribution dict into terms."""
+    for key, value in contribution.items():
+        _add_term(terms, key, value)
```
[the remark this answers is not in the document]
blazingphoenix7: Thanks, but I'd keep `_merge_terms` and the two-step here. The aim is output that stays bit-for-bit identical to the current transform, and the two levels of dropping are what make that hold. Upstream builds each contribution as its own operator via `jordan_wigner_one_body`/`two_body` (whose `+=` drops sub-tolerance terms on its own, against zero), then folds that into the running total with a second `+=` that drops again. `_one_body_terms`/`_two_body_terms` plus `_merge_terms` mirror exactly that. Accumulating straight into the shared dict removes the first drop, so a sub-`EQ_TOLERANCE` contribution gets tested against the live total instead of against zero. Two places it shows up: an all-zero operator comes out as `{}` instead of `{(): 0.0}`, and a sub-tolerance term folded onto an existing value shifts it by up to `EQ_TOLERANCE`. Both stay within `isclose`, but they'd break the exact match, so I'll keep the helper.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -191,8 +225,15 @@ def jordan_wigner_one_body(p, q, coefficient=1.0):
...
     """
-    # Handle off-diagonal terms.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _one_body_terms(p, q, coefficient)
+    return qubit_operator
+
+
+def _one_body_terms(p, q, coefficient):
+    r"""Terms of a^\dagger_p a_q + h.c. under Jordan-Wigner, as a dict."""
+    terms = {}
```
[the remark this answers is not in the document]
blazingphoenix7: Keeping the dict-return form here, same reason as the `_merge_terms` thread: the local drop has to happen against zero, not against the running total.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -221,12 +262,18 @@ def jordan_wigner_two_body(p, q, r, s, coefficient=1.0):
...
     """
-    # Initialize qubit operator.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _two_body_terms(p, q, r, s, coefficient)
+    return qubit_operator
+
+
+def _two_body_terms(p, q, r, s, coefficient):
+    r"""Terms of a^\dagger_p a^\dagger_q a_r a_s + h.c. under JW, as a dict."""
+    terms = {}
```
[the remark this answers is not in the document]
blazingphoenix7: Same here, keeping the dict return so that first drop stays local.

src/openfermion/transforms/opconversions/jordan_wigner.py
arettig, line 273: You could check for a small `coefficient` at the start of `_one_body_terms` and return immediately if so, since all the transformed coefficients are guaranteed to be smaller than `coefficient`. This could have a large impact for sparse interaction operators.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -221,12 +262,18 @@ def jordan_wigner_two_body(p, q, r, s, coefficient=1.0):
...
 
     # Return zero terms.
     if (p == q) or (r == s):
```
arettig, line 330: Same as above: consider returning immediately if `coefficient` is small.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -129,6 +129,38 @@ def _jordan_wigner_diagonal_coulomb_hamiltonian(operator):
...
+    """Multiply a sorted Pauli string by a Z on the given qubit.
+
+    In the only caller the qubit carries at most a parity Z, so Z squared
+    cancels to the identity and no phase is introduced.
+    """
+    new_operators = [factor for factor in operators if factor[0] != index]
+    if len(new_operators) == len(operators):
+        new_operators.append((index, 'Z'))
+    new_operators.sort(key=lambda factor: factor[0])
+    return tuple(new_operators)
```
arettig: This method is a bit inefficient - the one caller of this method uses a list of operators that is contiguous so you should be able to just check the first and last indices to see if you need to insert or remove Z(index). Insertion would only happen on the beginning or end so there should be no need to sort. Deletion would only happen in the middle of the list.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -129,6 +129,38 @@ def _jordan_wigner_diagonal_coulomb_hamiltonian(operator):
     return qubit_operator
 
 
+def _add_term(terms, key, coefficient):
+    """Add a single Pauli term into a dict, dropping it if it becomes negligible.
+
+    This mirrors QubitOperator.__iadd__/__isub__ for one term so that
+    accumulating strings into a dict matches building and summing QubitOperators.
+    """
```
arettig, line 144: Please change this to a properly formatted Google docstring. I know the rest of this class doesn't follow this, but we're trying to make sure new code does.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -129,6 +129,38 @@ def _jordan_wigner_diagonal_coulomb_hamiltonian(operator):
...
+    value = terms.get(key, 0) + coefficient
+    if QubitOperator._issmall(value):
+        terms.pop(key, None)
+    else:
+        terms[key] = value
+
+
+def _merge_terms(terms, contribution):
+    """Add every term of a contribution dict into terms."""
```
arettig: Please change this to a properly formatted Google docstring.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -129,6 +129,38 @@ def _jordan_wigner_diagonal_coulomb_hamiltonian(operator):
...
+    for key, value in contribution.items():
+        _add_term(terms, key, value)
+
+
+def _toggle_z(operators, index):
+    """Multiply a sorted Pauli string by a Z on the given qubit.
+
+    In the only caller the qubit carries at most a parity Z, so Z squared
+    cancels to the identity and no phase is introduced.
+    """
```
arettig, line 178: Please change this to a properly formatted Google docstring.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -191,8 +225,15 @@ def jordan_wigner_one_body(p, q, coefficient=1.0):
...
     because they are equal to their own Hermitian conjugate.
     """
-    # Handle off-diagonal terms.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _one_body_terms(p, q, coefficient)
+    return qubit_operator
+
+
+def _one_body_terms(p, q, coefficient):
+    r"""Terms of a^\dagger_p a_q + h.c. under Jordan-Wigner, as a dict."""
```
arettig: Please change this to a properly formatted Google docstring.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -221,12 +262,18 @@ def jordan_wigner_two_body(p, q, r, s, coefficient=1.0):
...
     because they are equal to their own Hermitian conjugate.
     """
-    # Initialize qubit operator.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _two_body_terms(p, q, r, s, coefficient)
+    return qubit_operator
+
+
+def _two_body_terms(p, q, r, s, coefficient):
+    r"""Terms of a^\dagger_p a^\dagger_q a_r a_s + h.c. under JW, as a dict."""
```
arettig: Please change this to a properly formatted Google docstring.

arettig (changes requested):
Great PR, thanks so much for doing this! I left some comments mostly just asking for properly formatted docstrings on the new methods, but I have a couple suggestions to address some minor inefficiencies.

src/openfermion/transforms/opconversions/jordan_wigner.py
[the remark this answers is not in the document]
arettig: I agree with you here. Let's leave the double thresholding for now to ensure no numerical differences for this PR. The local copying is no worse than it was before.

That being said, we should definitely change this in the future in a separate PR. Accumulating in-place could be significantly faster, and threshing once at the end is numerically better practice. I wouldn't expect any physically meaningful changes to output, but it could break some tight tests so it would have to be done carefully.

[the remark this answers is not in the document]
blazingphoenix7: Thanks, that makes sense. Happy to take the in-place version as a follow-up once this is in, since the tolerance change deserves looking at on its own.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done, all five helpers now have Args and Returns sections.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Good catch, that's much better. The caller always passes a string covering a contiguous run of qubits in increasing order, so I check the ends and otherwise index straight to the offset: below the run prepends, above appends, and inside drops the parity Z sitting there. No scan and no sort now. I checked the new version against the old one over every string the three-unique branch actually produces and they agree.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done. No string either helper emits takes more than half the coefficient, and no string gets more than one contribution per call, so a negligible coefficient could only produce values that `_add_term` drops again. I used the same `_issmall` test so the boundary matches. I checked the result against the current output over a grid of coefficients around the tolerance, including complex ones, and the terms come out identical.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Same here.

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -191,8 +248,30 @@ def jordan_wigner_one_body(p, q, coefficient=1.0):
...
     because they are equal to their own Hermitian conjugate.
     """
-    # Handle off-diagonal terms.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _one_body_terms(p, q, coefficient)
+    return qubit_operator
+
+
+def _one_body_terms(p, q, coefficient):
+    r"""Terms of a^\dagger_p a_q + h.c. under Jordan-Wigner, as a dict.
```
arettig: nit: To make sure the math gets formatted:

```suggestion
    r"""Terms of $a^\dagger_p a_q + \text{h.c.}$ under Jordan-Wigner, as a dict.
```

src/openfermion/transforms/opconversions/jordan_wigner.py
```diff
@@ -221,12 +300,35 @@ def jordan_wigner_two_body(p, q, r, s, coefficient=1.0):
...
     because they are equal to their own Hermitian conjugate.
     """
-    # Initialize qubit operator.
     qubit_operator = QubitOperator()
+    qubit_operator.terms = _two_body_terms(p, q, r, s, coefficient)
+    return qubit_operator
+
+
+def _two_body_terms(p, q, r, s, coefficient):
+    r"""Terms of a^\dagger_p a^\dagger_q a_r a_s + h.c. under JW, as a dict.
```
arettig: nit:
```suggestion
    r"""Terms of $a^\dagger_p a^\dagger_q a_r a_s + \text{h.c.}$ under JW, as a dict.
```

arettig (changes requested):
Looks great, thanks for making the changes! Please add the latex formatting to your docstrings and hit the update branch button, then I'll merge it in!

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

src/openfermion/transforms/opconversions/jordan_wigner.py
blazingphoenix7: Done.

arettig (approved):
Looks great, thank you!

PR #1439, opened 2026-08-12
Use local RNG in MultiTensor tests

fzlzjerry:
## Summary

- give every randomized MultiTensor test a fresh, deterministically seeded `numpy.random.Generator`
- pass that generator into the tests explicitly and replace all 29 legacy global RNG calls in the module
- keep the suite-wide seed in place for modules that have not yet been migrated

## Rationale

This is an incremental part of #1060. It makes the ten randomized tests in `_multitensor_test.py` isolated and reproducible without changing random behavior elsewhere in the suite.

## Validation

- `check/all --only-changed-files origin/main` (`2287 passed`; formatting, Pylint, mypy, incremental coverage, ShellCheck, and notebook formatting all passed)
- representability suite (`44 passed`)
- MultiTensor tests with `pytest-randomly` seeds `1`, `42`, and `4294967295` (`10 passed` for each seed)
- MultiTensor tests with legacy `np.random.random` patched to raise on use (`10 passed`)


src/openfermion/contrib/representability/_multitensor_test.py
```diff
@@ -6,10 +6,15 @@
 from openfermion.contrib.representability._dualbasis import DualBasis, DualBasisElement
 
 
-def test_tmap():
-    a = np.random.random((5, 5))
-    b = np.random.random((4, 4))
-    c = np.random.random((3, 3))
+@pytest.fixture(name='rng')
+def rng_fixture() -> np.random.Generator:
```
mhucka: A more conventional way of defining this fixture would be to simply rename the fixture to `rng` and avoid the need for `name = 'rng'`. Was there a particular reason why that wasn't done here?

```suggestion
@pytest.fixture
def rng() -> np.random.Generator:
```

src/openfermion/contrib/representability/_multitensor_test.py
```diff
@@ -6,10 +6,15 @@
...
+    return np.random.default_rng(0)
+
+
+def test_tmap(rng):
```
mhucka: Please also add type annotations to the function signatures, here and in the other functions. (It's true that the original code didn't have that, but may as well add it when the signatures are being changed anyway.)

```suggestion
def test_tmap(rng: np.random.Generator) -> None:
```

mhucka (changes requested):
Thank you for this contribution. I have just a couple of small review comments.

src/openfermion/contrib/representability/_multitensor_test.py
fzlzjerry: Done in 4818c70. The fixture is now defined directly as `rng`, without the explicit `name` argument.

src/openfermion/contrib/representability/_multitensor_test.py
fzlzjerry: Done in 4818c70. Every test that accepts the fixture now annotates it as `np.random.Generator` and returns `None`; I also annotated the remaining nested helper. The file passes mypy, pylint, formatting, and the representability test suite (44 passed).

mhucka (approved):
Thanks for this work!

PR #1441, opened 2026-08-12
Add default template for repositories related to basis downfolding

### LINKED ISSUE #244: Basis downfolding
[arXiv:1502.04698](https://arxiv.org/abs/1502.04698) describes a method for "folding" correlations from Hamiltonians in larger basis sets into Hamiltonians in smaller basis sets. Thus, this is a method for obtaining minimal basis sets with accuracy much higher than something like STO-6G. The technique essentially seeks an effective Hamiltonian which describes ground states with more accuracy than would be possible with the usual Galerkin formulation. This technique seems especially important for quantum computers since most experiments are limited to a very small numbers of qubits.

abhiavi:
Add default template for repositories related to basis downfolding

Signed-off-by: Abhishek Khaparde <khaparde.abhishek@gmail.com>

Closes #244

### DIFF
--- a/.github/workflows/nightly-pytest.yaml
+++ b/.github/workflows/nightly-pytest.yaml
@@ -95,7 +95,7 @@ jobs:
         run: echo '::add-matcher::.github/problem-matchers/pytest.json'
 
       - name: Run Pytest
-        run: check/pytest
+        run: check/pytest -n auto
 
       - name: Print debugging info upon job failure
         if: failure()


mhucka:
@abhiavi Can you explain what is happening with this PR? The one file change it contains doesn't seem to the stated topic of the PR.

mhucka:
Closing because this looks like some kind of mistake.

PR #1446, opened 2026-08-18
Make `create-env-files.sh` respect min python version supported

mhucka:
This makes `dev_tools/requirements/create-env-files.sh` get the minimum version of Python the project supports from pyproject.toml, and use that as a constraint passed to `uv`.

### DIFF
--- a/dev_tools/requirements/create-env-files.sh
+++ b/dev_tools/requirements/create-env-files.sh
@@ -16,36 +16,56 @@
 set -o errexit
 set -o nounset
 
-declare -r usage="Usage: ${0} [-h] [UV_OPTIONS]
+# Go to the top of the local TFQ git tree. Do it early in case this fails.
+script_dir=$(CDPATH="" cd -- "$(dirname -- "${0}")" && pwd -P)
+repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel 2>/dev/null)
+cd "${repo_dir}"
+
+default_min_python=$(python3 -c 'import re
+s = open("pyproject.toml").read()
+m = re.search(r"requires-python.*?>=?(\d+\.\d+)", s)
+if not m:
+    m = re.search(r"target-version.*?py(\d)(\d+)", s)
+print(m.group(1) if "requires-python" in m.re.pattern else m.expand(r"\1.\2"))
+' 2>/dev/null || echo "3.10")
+
+declare -r usage="Usage: ${0} [-h] [-p X.Y] [UV_OPTIONS]
 Generate environment files for OpenFermion development using uv's
 'universal' option, making the result compatible with multiple
 Python versions. The output is written to subdirectories under
 dev_tools/requirements/.
 
 Options:
-  -h   Show this help message and exit
+  -h       Show this help message and exit
+  -p X.Y   Minimum Python version to support (default: ${default_min_python})
 
 All other options on the command line will be passed directly to
 'uv pip compile'. Run 'uv pip compile --help' to learn about the
 options available."
 
+min_python=("--python-version" "${default_min_python}")
+
 while [[ $# -gt 0 ]]; do
     case "${1}" in
         -h|--help) echo "${usage}"; exit 0 ;;
+        -p|--min-python|--python-version)
+            if [[ $# -lt 2 || -z "${2:-}" ]]; then
+                echo "Error: option '${1}' requires an argument" >&2
+                echo "${usage}" >&2
+                exit 1
+            fi
+            min_python=("--python-version" "${2}")
+            shift 2
+            ;;
         *) break ;;
     esac
 done
-
-# Go to the top of the local TFQ git tree. Do it early in case this fails.
-script_dir=$(CDPATH="" cd -- "$(dirname -- "${0}")" && pwd -P)
-repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel 2>/dev/null)
-cd "${repo_dir}"
 
 mkdir -p dev_tools/requirements/envs dev_tools/requirements/max_compat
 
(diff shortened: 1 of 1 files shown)


dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
 set -o errexit
 set -o nounset
 
-declare -r usage="Usage: ${0} [-h] [UV_OPTIONS]
+# Go to the top of the local TFQ git tree. Do it early in case this fails.
+script_dir=$(CDPATH="" cd -- "$(dirname -- "${0}")" && pwd -P)
+repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel 2>/dev/null)
```
pavoljuhas, line 21: We should keep the stderr message so user gets a clue what is going on if the script exits here on git-rev-parse failure.

```suggestion
repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel)
```

dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
...
+cd "${repo_dir}"
+
+default_min_python=$(python3 -c 'import re
```
pavoljuhas, line 24: This is way too complicated and as is it extracts the `black` formatter setting which may be unrelated to the actual supported version.  A better way would be to grep setup.py for `python_requires`, but I feel even that would be more fragile and more laborious to maintain in a long term than to just set
```
default_min_python=3.10
```
and bump it up when the minimum required version goes up.

Add: not needed at all per https://github.com/quantumlib/OpenFermion/pull/1446#discussion_r3808024299

dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
...
+
+declare -r usage="Usage: ${0} [-h] [-p X.Y] [UV_OPTIONS]
 Generate environment files for OpenFermion development using uv's
 'universal' option, making the result compatible with multiple
 Python versions. The output is written to subdirectories under
 dev_tools/requirements/.
 
 Options:
-  -h   Show this help message and exit
+  -h       Show this help message and exit
```
pavoljuhas, line 39:
```suggestion
  -h, --help  Show this help message and exit
```

dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
...
 All other options on the command line will be passed directly to
 'uv pip compile'. Run 'uv pip compile --help' to learn about the
 options available."
 
+min_python=("--python-version" "${default_min_python}")
+
 while [[ $# -gt 0 ]]; do
     case "${1}" in
         -h|--help) echo "${usage}"; exit 0 ;;
+        -p|--min-python|--python-version)
```
pavoljuhas, line 51: This is shadowing the uv-pip `-p, --python` option which would be ignored if specified first, but processed if placed as a later argument.  In addition, user would not be able to pass `--python-version=X.Y.Z` because uv does not accept that option multiple times.

I recommend to add the `python-version` setting to pyproject.toml instead and drop the custom `-p, --min-python` option here altogether:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 72ba61d..b8de85d 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,4 +73,5 @@ follow_untyped_imports = true
 [tool.uv.pip]
 universal = true
 generate-hashes = true
 custom-compile-command = "dev_tools/requirements/create-env-files.sh"
+python-version = "3.10"

```

dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
...
-script_dir=$(CDPATH="" cd -- "$(dirname -- "${0}")" && pwd -P)
-repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel 2>/dev/null)
-cd "${repo_dir}"
-
 mkdir -p dev_tools/requirements/envs dev_tools/requirements/max_compat
 
 # ~~~~ Generate basic requirements files ~~~~
 
-uv pip compile "$@" \
+uv pip compile "${min_python[@]}" "$@" \
```
pavoljuhas, line 68: This and below can be reverted after setting python-version in pyproject.toml.

pavoljuhas (changes requested):
Please set `python-version` via pyproject.toml instead.  Most of the changes in this PR can be then reverted.

dev_tools/requirements/create-env-files.sh
```diff
@@ -16,36 +16,56 @@
...
+repo_dir=$(git -C "${script_dir}" rev-parse --show-toplevel 2>/dev/null)
+cd "${repo_dir}"
+
+default_min_python=$(python3 -c 'import re
+s = open("pyproject.toml").read()
+m = re.search(r"requires-python.*?>=?(\d+\.\d+)", s)
+if not m:
+    m = re.search(r"target-version.*?py(\d)(\d+)", s)
+print(m.group(1) if "requires-python" in m.re.pattern else m.expand(r"\1.\2"))
+' 2>/dev/null || echo "3.10")
```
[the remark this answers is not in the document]
pavoljuhas: Slop and Bloat Gemini, Yuck!

dev_tools/requirements/create-env-files.sh
mhucka: Setup.py is going away in another PR, which is why this doesn't look in setup.py

mhucka:
> Please set `python-version` via pyproject.toml instead. Most of the changes in this PR can be then reverted.

Good point. There's another PR that will move setup.py settings to pyproject.toml. I'll close this one and then check if any changes are even needed after that, because the python version set in pyproject.toml may be all that's needed.

PR #1450, opened 2026-09-03
Accumulate ParallelLinearQubitOperator matvecs in-place (#1410)

### LINKED ISSUE #1410: Possible micro-optimization in linear_qubit_operator.py

Gemini Code Assist reported the following for linear_qubit_operator.py. This looks simple enough to do, but it should be checked for correctness, and some kind of unit tests should be added to go along with this.

**Performance/Memory Overhead in Single-Process Path (Lines 189-190):**

Using `functools.reduce(numpy.add, ...)` creates $N-1$ intermediate numpy arrays, where $N$ is the number of operators. For large state vectors, this causes significant memory churn and garbage collection overhead.
   
**Fix:** 

Initialize a single zero array and accumulate the results in-place using `+=` as suggested below.

```python
        if self.options.processes <= 1:
            res = numpy.zeros(x.shape, dtype=complex)
            for operator in self.linear_operators:
                res += operator * x
            return res
```

_Originally posted by @gemini-code-assist[bot] in https://github.com/quantumlib/OpenFermion/pull/1404#discussion_r3553904905_

GCA's suggestion should be implemented and tested.

P3 – not blocked; it's an idea

rosspeili:
Implements the suggestion from #1410: avoid `functools.reduce(numpy.add, …)` when combining partial matvecs, which was allocating a new full-sized array for every partial sum.

Used the approach from GCA and went a bit further:
- Shared in-place `+=` accumulation for both the single-process and multiprocess paths (same cost when combining worker results).
- Only set `forkserver` when `processes > 1`, so the single-process path works on Windows.

There is one nuance with `processes=1`, where operator grouping usually yields a single group so the biggest + is on the multiprocess reduce. Happy to trim or adjust if its better to keep the change scoped only to the single-process path.

Tests passed, and will wait for CI to see if I missed anything.

Fixes #1410

### DIFF
--- a/src/openfermion/linalg/linear_qubit_operator.py
+++ b/src/openfermion/linalg/linear_qubit_operator.py
@@ -11,7 +11,6 @@
 #   limitations under the License.
 """LinearQubitOperator is a linear operator from QubitOperator."""
 
-import functools
 import logging
 import multiprocessing
 
@@ -22,6 +21,20 @@ import scipy.sparse.linalg
 
 from openfermion.utils.operator_utils import count_qubits
 from openfermion.config import get_available_cpu_count
+
+
+def _accumulate_vectors(vectors, shape):
+    """Sum vectors into one array without reduce() intermediate allocations.
+
+    ``functools.reduce(numpy.add, ...)`` builds a new full-sized array for every
+    partial sum. For large state vectors that creates substantial temporary
+    memory pressure. Accumulating with in-place ``+=`` keeps a single result
+    buffer instead.
+    """
+    result = numpy.zeros(shape, dtype=complex)
+    for vector in vectors:
+        result += vector
+    return result
 
 
 def _bit_parity(values):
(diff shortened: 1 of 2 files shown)


src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -169,7 +182,9 @@ def __init__(self, qubit_operator, n_qubits=None, options=None):
         self.n_qubits = n_qubits
         self.options = options or LinearQubitOperatorOptions()
 
-        if not ParallelLinearQubitOperator._start_method_set:
+        # Only required when actually spawning workers; the single-process path
+        # must remain usable on platforms without forkserver (e.g. Windows).
+        if self.options.processes > 1 and not ParallelLinearQubitOperator._start_method_set:
```
mhucka: If the caller passes in an `options` object that has no `processes` field or `processes = None`, this will result in an error. Can you add a defensive check or wrap it in a `try-except` clause?

Maybe something like
```
    processes = getattr(self.options, 'processes', None) or 1
```
and then use `processes` instead of `self.options.processes`.

src/openfermion/linalg/linear_qubit_operator.py
```diff
@@ -24,6 +23,20 @@
 from openfermion.config import get_available_cpu_count
 
 
+def _accumulate_vectors(vectors, shape):
+    """Sum vectors into one array without reduce() intermediate allocations.
+
+    ``functools.reduce(numpy.add, ...)`` builds a new full-sized array for every
+    partial sum. For large state vectors that creates substantial temporary
+    memory pressure. Accumulating with in-place ``+=`` keeps a single result
+    buffer instead.
```
mhucka: Please switch the use of double backquotes to single backquotes in the docstring.

mhucka (changes requested):
Thank you for this contribution. If you could address a couple of minor items, I think it'll be ready to go after that.

rosspeili:
Thanks @mhucka, both addressed, lmk how this looks.

mhucka (approved):
Thanks for this!

